PR #277, opened 2024-09-11
Update TVM - Added pass that fuses padding into conv2d and maxpool2d

LPanosTT:
I merged this small PR which fuses pad ops that are placed on the inputs to conv2d/maxpool by setting the padding attribute instead. https://github.com/tenstorrent/tt-tvm/pull/31


LPanosTT:
No longer necessary, as TVM was uplifted in https://github.com/tenstorrent/tt-forge-fe/pull/242

Issue #283, opened 2024-09-12
Remove concept of forge shape


pilkicTT:
These are the remains of the buda stack, and should no longer be needed.

Tensor in "forge shape" represents 4d tile padded tensor.

pilkicTT:
Done with #2207

Issue #172, opened 2024-08-26
[Ops] Decompose indexing op into regular matmul instead of SMM


nvukobratTT:
Baed on the details from this task:
- https://github.com/tenstorrent/tt-forge-fe/issues/121

mbezuljTT:
I believe it should resolve it. 

to_layout should add padding. I'm still confused on where padding value is set. Some times padding value is important, like you need -inf when calculating max, or zero value at tensor that goes to matmul, or non zero value if doing eltwise division etc.

nvukobratTT:
> I believe it should resolve it.
> 
> to_layout should add padding. I'm still confused on where padding value is set. Some times padding value is important, like you need -inf when calculating max, or zero value at tensor that goes to matmul, or non zero value if doing eltwise division etc.

Thanks! 

@dgolubovicTT we should get the following PR merged soon:
- https://github.com/tenstorrent/tt-mlir/pull/624

Once that's done, let's see if this issue is resolved.

dgolubovicTT:
It seems like it is a runtime issue. In runtime runProgram method, all ttnn ops are executed, and then at the end of the method we get memory corruption: free(): invalid pointer (trying to deallocate).

dgolubovicTT:
I talked with @pilkicTT and it turns out that we had padding of constant input (lhs of matmul), due to old logic from buda that padded all constants to tile size (since every tensor in buda had to be rounded to tile size). I tried the quick fix, and managed to get tests passing. I expect to send this on code review as soon as Pedja's changes land in main.

dgolubovicTT:
I had one more test case failing due to transpose changing rank of tensor (3D -> 4D). This was because transpose runtime uses unsqueeze op to set tensor to 4D, but never returns it back to 3D. This is the PR on mlir side: [PR](https://github.com/tenstorrent/tt-mlir/pull/730). As soon as this is merged I will create PR for Indexing decompose.

Side note, is that currently Indexing decompose consists of one matmul and 2 transposes (if needed). When stride is 1 indexing could be simplified to some kind of select. However, select is still unsupported in MLIR. As soon as select is supported, I will follow up with the update of Indexing. Tracking issue: https://github.com/tenstorrent/tt-forge-fe/issues/299

PR #284, opened 2024-09-12
[consteval] don't pad tensors to tile

pilkicTT:
In buda, we always padded constant tensors to tiles. Now, that should not be the case.

Note: I'm not removing the concept of forge shape (4-d padded tiled tensor) and related params, just in case we need to workaround any ttnn issues. Filed issue #283


dgolubovicTT (approved):

nvukobratTT (approved):

pilkicTT:
Merging now, since this is a blocker for:
- #172

Issue #281, opened 2024-09-12
Fix data formats propagation for TensorFlow modules


pilkicTT:
The data types are not properly propagated from TensorFlow module to the Forge module (they always get overridden as `float32`).

See test `test_conv2d` in `forge/test/test_ops_tf.py`.

LPanosTT:
So in the latest main, TF input activation data types are passed correctly, but not parameters. After a little digging I've found this section of code that is responsible:  line 709 in `third_party/tvm/python/tvm/relay/frontend/tensorflow.py`

```
unsuported_types = [dtypes.bfloat16, ]
np_array = tensor_util.MakeNdarray(value.tensor)
orig_dtype = str(dtypes.as_dtype(value.tensor.dtype))[9:-2] # "i.e. <dtype: 'int32'> -> int32"
if dtypes.as_dtype(value.tensor.dtype) in unsuported_types:
    np_array = np_array.astype('float32')
```

This is here because bfloat16 is not a numpy data type.

nvukobratTT:
> So in the latest main, TF input activation data types are passed correctly, but not parameters. After a little digging I've found this section of code that is responsible: line 709 in `third_party/tvm/python/tvm/relay/frontend/tensorflow.py`
> 
> ```
> unsuported_types = [dtypes.bfloat16, ]
> np_array = tensor_util.MakeNdarray(value.tensor)
> orig_dtype = str(dtypes.as_dtype(value.tensor.dtype))[9:-2] # "i.e. <dtype: 'int32'> -> int32"
> if dtypes.as_dtype(value.tensor.dtype) in unsuported_types:
>     np_array = np_array.astype('float32')
> ```
> 
> This is done whilst building the TVM relay module from the tensorflow GraphDef.
> 
> This is here because bfloat16 is not a numpy data type.

Good catch @LPanosTT! 

Can we model this as convert it properly to bfloat16 at some point? E.g., if we need to keep it in float32 during TVM compile, let's keep track of it one way or another, and do a proper conversion bit before or during codegen to ForgeModule. 

What do you think? :))

LPanosTT:
> Can we model this as convert it properly to bfloat16 at some point? E.g., if we need to keep it in float32 during TVM compile, let's keep track of it one way or another, and do a proper conversion bit before or during codegen to ForgeModule.

[...]

PR #286, opened 2024-09-12
Disable default CPU fallback and conversion of TVM params

### LINKED ISSUE #285: Disable param conversion in TVM by default

nvukobratTT:
Speed up compile time and reduce memory consumption ~2x on host

- Disabling CPU fallback by default:
  - Focuses more on the preferable way of running models (on the device)
  - Reduces compile time as well
- Disabling conversion to TVM params:
  - Disables TVM verification as well (assert for checking it already exists)
  - Disables TVM consteval (adding assert with this check)

Fixes #285

Related TVM PR:
- https://github.com/tenstorrent/tt-tvm/pull/32

### DIFF
--- a/forge/forge/config.py
+++ b/forge/forge/config.py
@@ -121,7 +121,7 @@ class CompilerConfig:
 
     compile_depth: int = field(default=CompileDepth.FULL, metadata=as_json(CompileDepth))  # Defines compilation depth. Used to limit scope of some unit tests
 
-    enable_tvm_cpu_fallback: bool = True    # Create cpu device for unsupported forge ops
+    enable_tvm_cpu_fallback: bool = False    # Create cpu device for unsupported forge ops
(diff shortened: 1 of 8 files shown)


LPanosTT:
Does this solve the issue of tensorflow params having there data type forced to fp32? https://github.com/tenstorrent/tt-forge-fe/issues/281

nvukobratTT:
=

> > Does this solve the issue of tensorflow params having there data type forced to fp32? #281
> 
> Might be the case as this will disable setting params to TVM relay. Can you double-check it?

If it works, let's just add proper asserts as verification for TF will fail for these reasons as bug still exists. 

Therefore, based on your check, let's ether:
1. Fix the TF issue if this doesn't work
2. Push TF issue as lower prio of ResNet 50 milestone, and we can tackle it once ResNet runs e2e

dgolubovicTT (approved):

LPanosTT:
> 1. Fix the TF issue if this doesn't work

I've checked, it doesn't work. And the issue may actually be harder to solve than I previously thought. I've written more details in #281

Issue #181, opened 2024-08-26
Disable consteval in training mode


vladimirjovanovicTT:
Consteval on a parameter currently disrupts connection between forward and backward passes.
In this version, we will disable it when in training mode.

We will find a long-term solution (allowing consteval when it makes sense in training) in a next iteration.

PR #292, opened 2024-09-13
Disabled consteval in training mode.

vladimirjovanovicTT:
Solves #181 .
As described in the issue, we are simply disabling consteval in training mode for now, we will move to a fine-grained solution in the future.

### DIFF
--- a/forge/forge/compile.py
+++ b/forge/forge/compile.py
@@ -633,8 +633,11 @@ def run_post_initial_graph_pass(context: CompileContext) -> CompileDepth:
     graph_name = context.graph_name
     graph, intermediate_tensors = context.graph, context.intermediate_tensors
 
-    if compiler_cfg.enable_consteval:
+    # Issue #293 - check why we are running consteval here and also later as a separate pass.
+    # This should ideally be removed.
+    if should_run_consteval(context):
         run_consteval_graph_pass(graph)
+
     inserted_node_id_mapping, context.fracture_chip_id_assignments = run_post_initial_graph_passes(graph, compiler_cfg, compiler_cfg.fracture_groups)
 
     for inserted_node_id, original_node_id in inserted_node_id_mapping:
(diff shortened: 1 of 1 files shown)


forge/forge/compile.py
```diff
@@ -267,6 +267,9 @@ def forge_compile_from_context(context: CompileContext) -> CompiledModel:
         CompileDepth.FINISH_COMPILE: finish_compile,
     }
 
+    if context.loss_module is not None:
+        context.compiler_cfg.enable_consteval = False
```
nvukobratTT, line 271: What do you think about doing assert here, instead of implicit disable? 

I'm a bit scared for us to go into a domain where the compiler changes its own configurations. E.g. use will print config, see that cont eval is enabled, then during compile it'll report that we're skipping it.

forge/forge/compile.py
```diff
@@ -267,6 +267,9 @@ def forge_compile_from_context(context: CompileContext) -> CompiledModel:
         CompileDepth.FINISH_COMPILE: finish_compile,
     }
 
+    if context.loss_module is not None:
```
nvukobratTT, line 270: How does this check work if the loss module is on the CPU? I also see that there is a specific `training` flag. Can we use it, or something similar?

pilkicTT: We require and expect loss and optimizer modules to be added as `forge.compile()` arguments, so that's not the issue. Anyhow, it's better to check the `training` flag, so I've modified the code.

forge/forge/compile.py
pilkicTT: I agree that it can be confusing...

We can print a warning if you'd like that better...

pilkicTT: Also, we can drop this altogether and disable consteval manually when needed for now... But I would like to go this way, since I think it's better to automatically handle the failure cases.

nvukobratTT (approved):

PR #304, opened 2024-09-17
Optional producer and/or consumer op on Maxpool2d tests

LPanosTT:
Add test options to place conv before and/or after maxpool in maxpool test


forge/test/mlir/test_ops_tf.py
```diff
@@ -149,12 +153,18 @@ def test_maxpool2d(
     class MaxPool(tf.keras.Model):
         def __init__(self):
             super().__init__() 
+            if pre_conv:
```
nvukobratTT, line 156: Looks good! Let's add cases with pre/post eltwise ops as well, just to include some ops that don't have special input handling, and that are not bound to common use cases like conv + pool is.

LPanosTT: Actually, that isn't working so smoothly. Issue here: https://github.com/tenstorrent/tt-mlir/issues/738

nvukobratTT:
Closing this one for now in favor of newer conv2d PRs

PR #310, opened 2024-09-18
Adding decompose for Indexing op using Transpose and Matmul. 

dgolubovicTT:
- Update tt-mlir to consume bugfix for Transpose.
- Adding decompose for Indexing op using Transpose and Matmul. Based on indexing parameters create left hand side picker matrix and multiplying it with input matrix. Using transposes to manipulate dimension order.


forge/forge/op/eval/forge/tm.py
```diff
@@ -1413,6 +1378,35 @@ def decompose(type, attr, dc, inputs):
                 rank -= 1
             dc.fuse(result)
             return
+def create_row_picker_matrix(row_indices,  lhs_num_cols, lhs_num_channels=None, lhs_batch_size=None):
+    """
+    Create a sparse matrix that picks rows from a matrix.
```
nvukobratTT, line 1383: Function docs seem out-of-date. Can you update it to batter match the intended purpose? 

Also, a minor example will be beneficial for easier future understanding :))

forge/test/mlir/test_ops.py
```diff
@@ -558,3 +558,32 @@ def forward(self, x):
...
+    if len(shape) == 2 and dim == -3:
+        pytest.skip("Skipping since indexing on dim=-3, 2D tensor doesn't make sense")
+    if stop > shape[dim]:
+        pytest.skip("Skipping since stop > shape[dim]")
+    class ForgeIndexing(forge.ForgeModule):
+        def __init__(self, dim, start, stop, stride):
+            super().__init__("ForgeTest")
+
+        def forward(self, x):
+            return forge.op.Index("indexing_op_1", x, dim, start, stop, stride)
```
nvukobratTT, line 577: Can we use PyTorch op instead of native Index op on our end?

nvukobratTT: Did you try to run pure indexing over it? E.g. `input_tensor[:,:,1]`? To what is it lowered through TVM?

dgolubovicTT: I tried now, and I get some error in generate graph: `AttributeError: 'Tensor' object has no attribute 'src_op'`. I can take a look into this but it seems like a problem with generating initial graph...

dgolubovicTT: Both pure indexing and index_select yield the same error. I have to look into this, but I wouldn't delay this mr because of that. What do you think?

nvukobratTT: I agree. Feel free to create a new issue to tackle with this problem. I don't think this one is a blocker for Llama 3B.

nvukobratTT (approved):
LGTM! 💪

nvukobratTT:
cc @chandrasekaranpradeep

LPanosTT:
This looks good to me, but is there not an index op in ttnn already?

dgolubovicTT:
Not that I am aware of. I found similar behaviour for embedding op, but I can't find explicit indexing...

PR #329, opened 2024-09-23
Forge tests for Softmax operator

vobojevicTT:
Add tests for Forge Softmax operator.

### DIFF
--- a/forge/test/operators/pytorch/nn/test_softmax.py
+++ b/forge/test/operators/pytorch/nn/test_softmax.py
@@ -1,0 +1,436 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+
+# Tests for testing of nn operators
+#
+# In this test we use pytorch tensors and operators to verify forge operators
+
+
+# GENERAL OP SUPPORT TEST PLAN:
+# 1. Operand type - any supported type
+# 2. Operand source(s):
+# (+)  2.1 From another op
+#       - Operator -> input
+# (+)  2.2 From DRAM queue
+#       - input_queue flag = false
+#       - Special case of From host? May it be triggered if the operator is not the first node of the network?
+#       - Can this be triggered from pybuda.Parameter?
+#       - Can this be triggered from big pybuda.Constant?
+# (+)  2.3 Const Inputs (const eval pass)
+#       - Operator where all inputs are constants. Does it make difference if tensor is big > L1
+#       - Verification via netlists that scenario is triggered???
+# (+)  2.4 From host
+#       - Input tensor as input of network -> Operator is first node in network and input_queue flag = true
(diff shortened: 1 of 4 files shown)


forge/test/operators/pytorch/nn/test_softmax.py
```diff
@@ -0,0 +1,436 @@
...
+class ModelFromAnotherOp(torch.nn.Module):
+
+    model_name = "model_op_src_from_another_op"
+
+    def __init__(self, shape, dim):
+        super(ModelFromAnotherOp, self).__init__()
+        self.testname = "Softmax_operator_test_op_src_from_another_op"
+        self.shape = shape
+        self.dim = dim
+        self.softmax = torch.nn.Softmax(dim=self.dim)
```
vbrkicTT, line 80: @nvukobratTT 
Here we use class variant of softmax https://pytorch.org/docs/stable/generated/torch.nn.Softmax.html. There is also functional variant https://pytorch.org/docs/stable/generated/torch.nn.functional.softmax.html.
Class variant is basically wrapper around functional so it probably doesn't make difference which we select.
Maybe package `nn` is too wide here. I suggest we should something more specific like use `activations`.

nvukobratTT: I don't have hard preferences on this one. Using either class definition or functional one would output the same results compiler wise :)) 

From a model perspective, it's common to see both of these, so let's choose one of the approaches for our testing, and if needed we can extend it for other definitions as well (but this is unlikely)

dgolubovicTT (approved):
Legit!

forge/test/operators/pytorch/nn/test_softmax.py
```diff
@@ -0,0 +1,436 @@
...
+    model: Type[torch.nn.Module],
+    dim: int,
+    input_shape: TensorShape,
+    number_of_operands: int,
+    input_params: List[Dict] = [],
+    input_source_flag: InputSourceFlags = None,
+    dev_data_format: forge.DataFormat = None,
+    math_fidelity: forge.MathFidelity = None,
+):
+    """Common verification function for all tests"""
```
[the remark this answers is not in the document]
dgolubovicTT: @vobojevicTT any thoughts?

forge/test/operators/pytorch/nn/test_softmax.py
vobojevicTT: I agree that nn package is too wide to organize our tests in that way. We should consider to reorganize tests in another PR.

forge/test/operators/pytorch/nn/test_softmax.py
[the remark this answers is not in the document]
vobojevicTT: Not sure why, but I missed this comment before.
This 'code' comment was misleading. Common functionality is shared via `VerifyUtils.verify`. Here we have an opportunity to adjust tensors and verification parameters per operator.

Issue #343, opened 2024-09-26
Remove duplicating embedding tables in pre_lowering_passes


dgolubovicTT:
Remove duplicating embedding tables when one operand isn't embedding op, because this is the spatial feature that was solving blocking issues.

PR #344, opened 2024-09-26
Remove duplicating embedding tables

dgolubovicTT:
Solves #343. Remove duplicating embedding tables when one operand isn't embedding op, because this is the spatial feature that was solving blocking issues. In Forge, we don't need this feature, so deleting it...


nvukobratTT (approved):
LGTM!

PR #359, opened 2024-10-03
Creating test tensors according to data formats

vbrkicTT:
Creating test tensors depending on targeting dev data format and related value range.

Features:
  - Choosing torch datatype according to forge dev data type
  - Calculating intersection between forge and torch data types to avoid conversion losses
  - Specifying value range. Support small, large, positive, negative

Issue #239

### DIFF
--- a/forge/test/operators/utils/compat.py
+++ b/forge/test/operators/utils/compat.py
@@ -8,11 +8,13 @@ import forge
 import torch
 
 from loguru import logger
-from typing import Optional, List
+from typing import Optional, List, Union
 
 from forge import ForgeModule, Module, VerifyConfig
 from forge.op_repo import TensorShape
 from forge.op.eval.common import compare_with_golden
+
+from .datatypes import OperatorParameterTypes, ValueRanges, ValueRange
 
 
 # TODO - Remove this class once TestDevice is available in Forge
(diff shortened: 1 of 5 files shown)


forge/test/operators/utils/compat.py
```diff
@@ -22,6 +22,85 @@ class TestDevice:
     pass
 
 
+class TestTensorsUtils:
+
+    dev_data_format_to_dtype = {
+        # forge.DataFormat.Bfp2: torch.float16,
+        forge.DataFormat.Bfp2: torch.bool,  # mapped
+        forge.DataFormat.Bfp2_b: torch.bfloat16,
```
nvukobratTT, line 30: Isn't this bool as well?

forge/test/operators/utils/compat.py
```diff
@@ -22,6 +22,85 @@ class TestDevice:
...
+        forge.DataFormat.Bfp4: torch.float16,
+        forge.DataFormat.Bfp4_b: torch.bfloat16,
+        forge.DataFormat.Bfp8: torch.float16,
```
nvukobratTT, line 33: I know that torch has 8bit FP as well: `torch.float8_e4m3fn`, however, seems like there is limited support for it. As that is the case, `torch.float16` is probably the smallest general dtype we can use (next to the bool and int variants). 

In sum, current implementation is valid.

forge/test/operators/utils/compat.py
```diff
@@ -22,6 +22,85 @@ class TestDevice:
...
+        if dev_data_format is None:
+            dtype = None
+        else:
+            # dtype = torch.float32
+            if dev_data_format in cls.dev_data_format_to_dtype:
+                dtype = cls.dev_data_format_to_dtype[dev_data_format]
+            else:
+                raise ValueError(f"Unsupported dtype for dev data format: {dev_data_format}")
+
+            if dtype in (torch.uint8,):
```
nvukobratTT, line 64: Why we're doing this casting?

forge/test/operators/utils/compat.py
```diff
@@ -22,6 +22,85 @@ class TestDevice:
     pass
 
 
+class TestTensorsUtils:
+
+    dev_data_format_to_dtype = {
+        # forge.DataFormat.Bfp2: torch.float16,
+        forge.DataFormat.Bfp2: torch.bool,  # mapped
```
dgolubovicTT, line 29: Bfp data formats represent floating point numbers, not boolean values. Therefore I think we should have here some float...

forge/test/operators/utils/compat.py
dgolubovicTT: I would say contrary... As stated above, it seems illogical to cast any float to boolean...

forge/test/operators/utils/compat.py
```diff
@@ -22,6 +22,85 @@ class TestDevice:
...
+        forge.DataFormat.Bfp8: torch.float16,
+        forge.DataFormat.Bfp8_b: torch.bfloat16,
+
+        forge.DataFormat.Float16: torch.float16,  # mapped
+        forge.DataFormat.Float16_b: torch.bfloat16,  # mapped
+        forge.DataFormat.Float32: torch.float32,  # mapped
+
+        forge.DataFormat.Int8: torch.int8,  # mapped
+        forge.DataFormat.Lf8: torch.float16,
+        forge.DataFormat.RawUInt16: torch.uint8,
```
dgolubovicTT, line 42: Casting RawUInt16 and RawUInt32 to uint8 can lead to significant loss in precision. What is the logic behind this?

forge/test/operators/utils/compat.py
```diff
@@ -22,6 +22,85 @@ class TestDevice:
...
+    def get_random_torch_inputs(cls, dtype: torch.dtype, input_shapes: List[TensorShape], generator: torch.Generator) -> List[torch.Tensor]:
+
+        if dtype is None:
+            inputs = [torch.rand(input_shape, generator=generator) for input_shape in input_shapes]
+        elif dtype in (torch.float16, torch.bfloat16, torch.float32):
+            # TODO Testing big and small numbers should be configurable depending on operator
+            # Some operators support mix small and big numbers
+            # TODO Testing small numbers close to 0
+            inputs = [torch.rand(input_shape, dtype=dtype, generator=generator) for input_shape in input_shapes]
+            min, max = 100, 10000
```
dgolubovicTT, line 88: Why this range? It is probably most important to test inputs around 0. Are we doing this range (close to zero somewhere else?)

forge/test/operators/utils/compat.py
vbrkicTT: Here was by mistake selected Bfp2 instead of Bfp2_b.

In general mapping of datatypes here is done according to forge.tensor.pytorch_dtype_to_forge_dataformat. E.x. https://github.com/tenstorrent/tt-forge-fe/blob/589e2ed6251228016ab3fa6fcd093f9a7b5f3687/forge/forge/tensor.py#L604

forge/test/operators/utils/compat.py
vbrkicTT: it's redundant and should be removed

forge/test/operators/utils/compat.py
vbrkicTT: We plan to test both small and big numbers. Depending on operator it should be configurable.

forge/test/operators/utils/compat.py
dgolubovicTT: Any thoughts on this @vbrkicTT

vbrkicTT: I think we clarified this today with @nvukobratTT.
You are right that it's wrong to map floating point to boolean. I followed mapping from pytorch_dtype_to_forge_dataformat which was probably completely misleading to do here. The types of input tensors should be as close as possible between forge and torch types.

forge/test/operators/utils/compat.py
vbrkicTT: Torch data formats are now restricted to float32 and int32. It looks like they are working fine and others like bfloat16, float16, int16, uint8, int8 are causing many issues.

forge/test/operators/utils/compat.py
vbrkicTT: No booleans for now

forge/test/operators/utils/compat.py
vbrkicTT: removed

forge/test/operators/utils/compat.py
vbrkicTT: no booleans anymore, only floats and integer for now

forge/test/operators/utils/compat.py
vbrkicTT: Idea was to use unsigned number and only uint8 isn't marked as limited support.
To avoid loss of data mismatch on inputs tensors after conversion torch -> forge there is now logic of calculating intersection between torch and forge datatypes. So now tensors in torch will be created only with values that could be converted to forge without loss for integer and without significant loss for floats.

However since specifying torch types to types other that int32 cause issues, for the initial implementation I choose int32 for all integer types

forge/test/operators/utils/compat.py
vbrkicTT: Included initial implementation of value range limiting via ValueRanges.
It's now supported to specify different value range for input tensors depending on operator and/or dev data type.

E.x. for sqrt accepts only positive numbers, for some operators we can specify small (<1) or large (unlimited) value ranges

forge/test/operators/utils/compat.py
```diff
@@ -22,12 +24,198 @@ class TestDevice:
...
+    def get_random_torch_input(
+        cls,
+        dev_data_format: forge.DataFormat,
+        dtype: torch.dtype,
+        input_shape: TensorShape,
+        generator: torch.Generator,
+        value_range: Optional[OperatorParameterTypes.RangeValue] = None,
+    ) -> List[torch.Tensor]:
+
+        if dtype is None:
```
dgolubovicTT, line 174: this if elseif will tend to explode with new data formats. Can we make some helper functions for generating random inputs for each category?

forge/test/operators/utils/compat.py
```diff
@@ -22,12 +24,198 @@ class TestDevice:
...
+        # forge.DataFormat.Float16: torch.float16,  # fatal python error
+        forge.DataFormat.Float16: torch.float32,  # mapped
+        forge.DataFormat.Float16_b: torch.float32,  # mapped
+        forge.DataFormat.Float32: torch.float32,  # mapped
+        # forge.DataFormat.Lf8: torch.float16,  # fatal python error
+        # forge.DataFormat.Lf8: torch.bfloat16,  # pcc failed
+        forge.DataFormat.Lf8: torch.float32,
+        # forge.DataFormat.RawUInt8: torch.int8,  # unsupported
+        # forge.DataFormat.RawUInt8: torch.uint8,  # Unsupported torch dtype
+        forge.DataFormat.RawUInt8: torch.int32,
```
dgolubovicTT, line 48: I would add some comment here to explain why we have this mapping, and that we won't have precission loss because we tailor our domen of generated values...

forge/test/operators/utils/compat.py
vbrkicTT: Dtypes are grouped as floats, integers and bools so it's not expected to have more ifs with new dataformats. However since too many ifs in single function is already unnecessary hard to read there is new method `get_signed_value_range`

vbrkicTT: Actually this whole method `get_signed_value_range` looks not needed. I will remove it

vbrkicTT: Removed

dgolubovicTT (approved):

PR #367, opened 2024-10-04
[training] add unit tests for split graph pass

### LINKED ISSUE #363: [testing] add unit tests for split graph pass

pilkicTT:
Unit tests are added to test some of the invariants that need to hold when extracting the forward and backward pass as separate graphs.

Added variants for training with recompute, however that part still doesn't work (#366). However, the testing infra is in place, so once the issue is fixed this test should hopefully work.

Changes to the `forge_graph_module.hpp` are only whitespace (formatting).

Closes #363

### DIFF
--- a/forge/csrc/autograd/autograd.hpp
+++ b/forge/csrc/autograd/autograd.hpp
@@ -19,7 +19,7 @@ namespace tt {
 
 namespace autograd2 {
 
-struct __attribute__ ((visibility ("hidden"))) autograd_config {
+struct autograd_config {
     bool recompute = false; // Add recompute
     py::object optimizer = py::none();
 };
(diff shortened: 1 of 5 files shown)


dgolubovicTT (approved):

nvukobratTT (approved):
LGTM! 

@vmilosevic @pilkicTT let's think about the structure of running tests when we start with Build/Test split. There are different kind of tests we want to cover:
- Unit tests
- Component tests
- E2E tests (e.g. ops)
- Model tests

For utilizing our infra resource as best as possible, running these in stages will be good. E.g. if some unit/component tests fail, there is no need to run full e2e or model tests.

PR #372, opened 2024-10-07
Adding test inference of Llama3B with prefil on device and decode on CPU

dgolubovicTT:
Adding test inference of Llama3B with prefil on device and decode on CPU without KV chache.

### DIFF
--- a/forge/test/mlir/llama/tests/test_llama_prefil.py
+++ b/forge/test/mlir/llama/tests/test_llama_prefil.py
@@ -1,0 +1,100 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+import pytest
+from transformers import LlamaTokenizer
+
+import forge
+from test.mlir.llama.utils.utils import load_model
+from forge.op.eval.common import compare_with_golden
+
+
+def prefil_on_cpu(model, input_ids):
+    with torch.no_grad():
+        transformer_outputs = model.model(
+            input_ids=input_ids,  # Pass the entire updated sequence
+        )
+        hidden_states = transformer_outputs.last_hidden_state
+    return hidden_states
+
+
+def decode_on_cpu(model, tokenizer, input_ids, hidden_states, max_new_tokens):
+    output_logits = []
+    for i in range(max_new_tokens):
+        # Use only the hidden state of the last token in the sequence
+        next_token_logits = model.lm_head(hidden_states[:, -1, :])
+        output_logits.append(next_token_logits)
+
+        # Get the next token ID
+        next_token_id = torch.argmax(next_token_logits, dim=-1)
+
+        # Stop if the EOS token is generated
+        if next_token_id == tokenizer.eos_token_id:
+            break
+
+        # Update input_ids with the new token
+        input_ids = torch.cat([input_ids, next_token_id.unsqueeze(0)], dim=-1)
+
+        # Run the model again to get the new hidden state
+        # Here we do effectively prefil again because we don't use KV cache.
+        # If we used KV cache, we could just run the model with for the new token.
+        with torch.no_grad():
+            transformer_outputs = model.model(
+                input_ids=input_ids,  # Pass the entire updated sequence
+            )
+        hidden_states = transformer_outputs.last_hidden_state
+
+    return input_ids, output_logits
+
+
+@pytest.mark.xfail()
+def test_llama_prefil_on_device_decode_on_cpu():
+    """
+    This function tests the inference of the Llama 3B model split into two parts:
+    - The first part is the prefilling of the model on the device.
+    - The second part is the decoding of the model on the CPU without KV cache.
+    """
+    # Load Llama 3B model and tokenizer
+    model_path = "openlm-research/open_llama_3b"
+    model, tokenizer = load_model(model_path, use_cache=False, return_dict=True)
+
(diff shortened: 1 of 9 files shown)


forge/test/mlir/llama/tests/test_llama_prefil.py
```diff
@@ -0,0 +1,78 @@
...
+    model_decoder.config.use_cache = False
+    model_decoder.config.output_attentions = False
+    model_decoder.config.output_hidden_states = False
+    model_decoder.config.return_dict = True
+    compiled_decoder = forge.compile(model_decoder, sample_inputs=input_ids)
+
+    # Prefill Phase - Process the initial prompt
+    transformer_outputs = compiled_decoder(input_ids)
+    # Get hidden states for all tokens from the last "transformer layer"
+    hidden_states = transformer_outputs.last_hidden_state
```
dgolubovicTT, line 69: @nvukobratTT Do we want to compare last hidden states for compiled model and torch golden? How do we want to do verification in text generation?

nvukobratTT: Do you see some benefits for comparison of hidden states? Can't we use logits (output of LM head) or just argmax out of it and compare tokens (logits are probably more flexible)?

That way, we can compile whole model with LM head as well. Let me know if I'm missing some details

forge/test/mlir/llama/tests/test_llama_prefil.py
```diff
@@ -0,0 +1,78 @@
...
+    model = LlamaForCausalLM.from_pretrained(model_path)
+    tokenizer = LlamaTokenizer.from_pretrained(model_path)
+
+    # Prepare input sentence
+    prompt = "Q: What is the largest animal?\nA:"
+    input_ids = tokenizer(prompt, return_tensors="pt").input_ids
+    print("length of input_ids before text generation: ", len(input_ids[0]))
+    # This is the part of the model needed for prefil; model without last Linear layer (lm_head)
+    model_decoder = model.get_decoder()
+    model_decoder.config.use_cache = False
```
nvukobratTT, line 60: Also, you can update util function for loading model with kwargs, and send these as well if needed :))

forge/test/mlir/llama/tests/test_llama_prefil.py
```diff
@@ -0,0 +1,78 @@
...
+
+        # Stop if the EOS token is generated
+        if next_token_id == tokenizer.eos_token_id:
+            break
+
+        # Update input_ids with the new token
+        input_ids = torch.cat([input_ids, next_token_id.unsqueeze(0)], dim=-1)
+
+        # Run the model again to get the new hidden state
+        hidden_states = prefil_on_cpu(model, input_ids)
```
nvukobratTT, line 36: Let's update a naming a bit, having prefil mentioned here is a bit confusing :))

forge/test/mlir/llama/tests/test_llama_prefil.py
```diff
@@ -0,0 +1,78 @@
...
+    
+    return input_ids, hidden_states
```
nvukobratTT, line 38: Do we need to return hidden states after decode?

forge/test/mlir/llama/tests/test_llama_prefil.py
dgolubovicTT: OFC that LM head is part of the model, but in this test it is ran on CPU so that was my line of resoning when I decided to compare last hidden states. But since this test will evolve so that LM is also on Device, I can switch to comparing output...

forge/test/mlir/llama/tests/test_llama_prefil.py
```diff
@@ -0,0 +1,78 @@
...
+
+@pytest.mark.xfail()
+def test_llama_prefil_on_device_decode_on_cpu():
+    """
+    This function tests the inference of the Llama 3B model split into two parts:
+    - The first part is the prefilling of the model on the device.
+    - The second part is the decoding of the model on the CPU.
+    """
+    # Load Llama 3B model and tokenizer
+    model_path = "openlm-research/open_llama_3b"
```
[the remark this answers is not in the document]
dgolubovicTT: No, I am using this model_path to load a tokenizer since load_model doesn't return a tokenizer... Should this be inside load_model?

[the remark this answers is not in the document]
dgolubovicTT: I added loading tokenizer inside load_model method...

forge/test/mlir/llama/tests/test_llama_prefil.py
```diff
@@ -0,0 +1,78 @@
...
+    model = load_model(model_path, use_cache=False, return_dict=True)
+    tokenizer = LlamaTokenizer.from_pretrained(model_path)
+
+    # Prepare input sentence
+    prompt = "Q: What is the largest animal?\nA:"
+    input_ids = tokenizer(prompt, return_tensors="pt").input_ids
+    
+    # This is the part of the model needed for prefill; model without the last Linear layer (lm_head)
```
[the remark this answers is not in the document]
dgolubovicTT: As we discussed offline, prefill stage is defined as the stage where we prepare hidden states and KV cache to be able to do efficient decode. In our case, putting lm head inside prefill stage would mean we have created one token output before we started the decode. Therefore, leaving it like this because is more inline with intention of prefill and decode...

nvukobratTT (approved):

PR #384, opened 2024-10-08
Draft: Add llama 3b decode tests with prefill on cpu

chandrasekaranpradeep:
This is the Draft PR for Llama 3b model to run decode on TT devices and Prefill to run on CPU
There are two test cases based upon the cache/no cache(i.e past key values)

- `test_llama_prefill_on_cpu_decode_on_tt_no_cache `- Prefill will be executed on CPU and return only logits. Decode stage will be executed on TT devices. In this decode stage, the last token predicted is appended to  the sequences of input token and passed to the model for predicting the next token untill the a specified maximum number of new tokens is reached or an end-of-sequence token is encountered
- `test_llama_prefill_on_cpu_decode_on_tt_cache_padding ` - Prefill will be executed on CPU and returns logits and past key values.
                                                           Decode stage will be executed on TT devices. In this decode stage, the predicted last token (i.e single token) and past key values are zero padded in the key_value_seq_dim (dim=-2) and then passed to the model for predicting the next token (i.e which outputs single token) untill the a specified maximum number of new tokens is reached or an end-of-sequence token is encountered

### DIFF
--- a/forge/test/mlir/llama/test_llama_inference.py
+++ b/forge/test/mlir/llama/test_llama_inference.py
@@ -7,7 +7,8 @@ import pytest
 from transformers import LlamaConfig, LlamaForCausalLM, LlamaTokenizer
 
 import forge
-from test.mlir.llama.utils.utils import load_model
+from forge.op.eval.common import compare_with_golden_pcc
+from test.mlir.llama.utils.utils import load_model_and_tokenizer
 
 
 @pytest.mark.xfail()
@@ -120,3 +121,239 @@ def test_llama_inference_cache_cpu():
     # Generated text
     generated_text = tokenizer.decode(generated_tokens[0], skip_special_tokens=True)
     print(generated_text)
+
+
+@pytest.mark.xfail()
+def test_llama_prefill_on_cpu_decode_on_tt_no_cache():
+
+    # Load Llama 3B model and tokenizer
+    model_path = "openlm-research/open_llama_3b"
+    framework_model, tokenizer = load_model_and_tokenizer(
+        model_path=model_path, use_cache=False
+    )
+    tokenizer.pad_token_id = framework_model.config.pad_token_id
+
+    # Prepare input sentence
+    max_sequence_length = 53
+    prompt = "Q: What is the largest animal?\nA:"
+    inputs = tokenizer(
+        prompt,
+        return_tensors="pt",
+        max_length=max_sequence_length,
+        pad_to_max_length=True,
+        truncation=True,
+    )
+    input_ids = inputs.input_ids
+    attention_mask = inputs.attention_mask
+
+    non_padding_seq_len = int(torch.sum(attention_mask))
+    padding_seq_len = attention_mask.shape[1] - non_padding_seq_len
+
+    # Run Prefill on CPU with no cache to get the initial logits
+    logits = framework_model(input_ids=input_ids, attention_mask=attention_mask)
+
+    # Take the last non padding token index in logits for the next token
+    next_token_logits = logits[0][:, non_padding_seq_len - 1, :]
+    next_token = torch.argmax(next_token_logits, dim=-1)
+
+    # Update the input_ids with predicted token from prefill in the first padding token index
+    input_ids[:, non_padding_seq_len] = next_token
+    attention_mask[:, non_padding_seq_len] = 1
+
+    non_padding_seq_len = int(torch.sum(attention_mask))
+    padding_seq_len = attention_mask.shape[1] - non_padding_seq_len
+
+    # Compile the model on TT
+    compiled_model = forge.compile(
+        framework_model, sample_inputs=[input_ids, attention_mask]
+    )
+
+    # Run decode stage on TT device and generate tokens by appending predicted token into sequence of input tokens
+    # untill the a specified maximum number of new tokens is reached or an end-of-sequence token is encountered.
+    max_new_tokens = padding_seq_len
+    for idx in range(max_new_tokens):
+
+        model_inputs = [input_ids, attention_mask]
+
+        # CPU Inference
+        framework_output = framework_model(
+            input_ids=input_ids, attention_mask=attention_mask
+        )
+
+        # Run on TT device
+        tt_output = compiled_model(*model_inputs)
+        tt_output = [tt_out.to("cpu") for tt_out in tt_output]
+
+        # Validate TT result with Framework
+        assert all(
+            [
+                compare_with_golden_pcc(golden=fw_out, calculated=tt_out, pcc=0.99)
+                for fw_out, tt_out in zip(framework_output, tt_output)
+            ]
+        )
+
+        next_token_index = non_padding_seq_len + idx
+        next_token_logits = tt_output[0][:, next_token_index - 1, :]
+        next_token = torch.argmax(next_token_logits, dim=-1)
+
+        if next_token == tokenizer.eos_token_id:
+            break
+
+        input_ids[:, next_token_index] = next_token
+        attention_mask[:, next_token_index] = 1
+
+    # Generated text
+    generated_text = tokenizer.decode(input_ids[0], skip_special_tokens=True)
+    print("generated_text=", generated_text)
+
+
+# The test is just workaround for checking with/without padding.
+@pytest.mark.xfail()
+def test_llama_prefill_on_cpu_decode_on_tt_cache_padding():
+
+    # Load Llama 3B model and tokenizer
+    model_path = "openlm-research/open_llama_3b"
+    framework_model, tokenizer = load_model_and_tokenizer(
+        model_path=model_path, use_cache=True, return_dict=True
+    )
+
+    class LlamaModelWrapper(torch.nn.Module):
+        """
+        In LlamaModelWrapper class, forward function takes single input token (i.e last predicted token)
+        and past key values from the previous iteration and return logits and past key values
+        Args:
+            input_id (`torch.Tensor`) - shape of (batch_size, 1)
+            past_key_values (`List[List[torch.Tensor, torch.Tensor]]`) - key/values shape - (batch_size, num_of_key_values_heads, key_value_seq_len, head_dim)
+        Returns:
+            outputs - Logits of shape (batch_size, 1, vocab_size) and past_key_values (`List[List[torch.Tensor, torch.Tensor]]`)
+            past key/values tensor shape - (batch_size, num_of_key_values_heads, key_value_seq_len + 1, head_dim)
+        """
+
+        def __init__(self, model):
+            super().__init__()
+            self.model = model
+
+        def forward(self, input_id, past_key_values):
+            model_outputs = self.model(
+                input_ids=input_id, past_key_values=past_key_values
+            )
+            return model_outputs.logits, model_outputs.past_key_values
+
+    llama_model = LlamaModelWrapper(framework_model)
+
+    # Prepare input sentence
+    prompt = "Q: What is the largest animal?\nA:"
+    inputs = tokenizer(prompt, return_tensors="pt")
+
+    # Run Prefill on CPU with cache to get the initial logits and past key-values
+    prefill_output = framework_model(
+        input_ids=inputs.input_ids, attention_mask=inputs.attention_mask
+    )
+    next_token_logits = prefill_output.logits[:, -1, :]
+    next_token = torch.argmax(next_token_logits, dim=-1)
+
+    generated_tokens = inputs.input_ids
+    generated_tokens = torch.cat([generated_tokens, next_token.unsqueeze(0)], dim=-1)
+
+    past_key_values_list = [[k, v] for k, v in prefill_output.past_key_values]
+
+    model_inputs = [next_token.unsqueeze(0), past_key_values_list]
+
+    # Padding on past key values can be enabled by setting apply_padding_on_past_key_values to True.
+    apply_padding_on_past_key_values = True
+    max_new_tokens = 32
+    if apply_padding_on_past_key_values:
+        padding_seq_len = max_new_tokens
+        non_padding_seq_len = past_key_values_list[0][0].shape[-2]
+
+        # Zero Pad past key values in key_value_seq_len(i.e -2) dimension
+        # Before padding past key values tensor shape -> (batch_size, num_of_key_values_heads, key_value_seq_len, head_dim)
+        # After Padding Past key value tensor shape -> (batch_size, num_of_key_values_heads, key_value_seq_len + padding_seq_len, head_dim)
+        for idx, (k, v) in enumerate(model_inputs[1]):
+            model_inputs[1][idx][0] = torch.cat(
+                [
+                    k,
+                    torch.zeros(
+                        k.shape[-4], k.shape[-3], max_new_tokens, k.shape[-1]
+                    ).to(k.dtype),
+                ],
+                dim=-2,
+            )
+            model_inputs[1][idx][1] = torch.cat(
+                [
+                    v,
+                    torch.zeros(
+                        v.shape[-4], v.shape[-3], max_new_tokens, v.shape[-1]
+                    ).to(k.dtype),
+                ],
+                dim=-2,
+            )
+
+    # # Compile the model
+    # compiled_model = forge.compile(llama_model, sample_inputs=model_inputs)
+
+    # Run decode stage on TT device and generate tokens by passing the last predicted token and the past key values.
+    # untill the a specified maximum number of new tokens is reached or an end-of-sequence token is encountered.
+    for max_new_tokens_idx in range(max_new_tokens):
+
+        # CPU Inference
+        model_outputs = llama_model(model_inputs[0], model_inputs[1])
+
+        # TT will return the logits and past key values as list of tensor, so flattening
+        # framework output past key values from List(List(Key1, Values1), ... , List(Key26, Values26)) to
+        # List(Key1, Values1, ... , Key26, Values26) for comparing the Framework and TT output in similar fashion.
+        framework_output = [model_outputs[0]]
+        for k, v in model_outputs[1]:
+            framework_output.append(k)
+            framework_output.append(v)
+
+        # # Run on TT device
+        # tt_output = compiled_model(*model_inputs)
+        # tt_output = [tt_out.to("cpu") for tt_out in tt_output]
+
+        # # Validate TT result with Framework
+        # assert all([compare_with_golden_pcc(golden=fw_out, calculated=tt_out, pcc=0.99) for fw_out, tt_out in zip(framework_output, tt_output)])
+
+        logits = framework_output[0]
+        past_key_values = framework_output[1:]
+
+        next_token_logits = logits[:, -1, :]
+        next_token = torch.argmax(next_token_logits, dim=-1)
+        print("next_token=", next_token)
+
+        if next_token == tokenizer.eos_token_id:
+            break
+
+        model_inputs = [next_token.unsqueeze(0)]
+
+        # Formate the past key values from List(Key1, Values1, ... , Key26, Values26) to
+        # List(List(Key1, Values1), ... , List(Key26, Values26))
+        model_inputs.append(
+            [
+                past_key_values[idx : idx + 2]
+                for idx in range(0, len(past_key_values), 2)
+            ]
+        )
+
+        if apply_padding_on_past_key_values:
+
+            for idx in range(len(model_inputs[1])):
+                key_state = torch.clone(model_inputs[1][idx][0][:, :, -1, :])
+                value_state = torch.clone(model_inputs[1][idx][1][:, :, -1, :])
+                model_inputs[1][idx][0][
+                    :, :, non_padding_seq_len + max_new_tokens_idx, :
+                ] = model_inputs[1][idx][0][:, :, -1, :]
+                model_inputs[1][idx][0] = model_inputs[1][idx][0][:, :, :-1, :]
+                model_inputs[1][idx][1][
+                    :, :, non_padding_seq_len + max_new_tokens_idx, :
+                ] = model_inputs[1][idx][1][:, :, -1, :]
+                model_inputs[1][idx][1] = model_inputs[1][idx][1][:, :, :-1, :]
+
+        generated_tokens = torch.cat(
+            [generated_tokens, next_token.unsqueeze(0)], dim=-1
+        )
+
+    print("generated_tokens[0]=", generated_tokens[0])
+    # Generated text
+    generated_text = tokenizer.decode(generated_tokens[0], skip_special_tokens=True)
+    print("generated_text=", generated_text)
--- a/forge/test/mlir/llama/utils/utils.py
+++ b/forge/test/mlir/llama/utils/utils.py
@@ -7,16 +7,26 @@ from transformers import LlamaConfig, LlamaForCausalLM, LlamaTokenizer
 import forge
 
 
-def load_model(model_path="openlm-research/open_llama_3b", use_cache=False):
-    # Load Llama 3B model
+def load_model_and_tokenizer(**kwargs):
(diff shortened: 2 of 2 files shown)


forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
...
+        [generated_tokens, next_token.unsqueeze(0)], dim=-1
+    )
+
+    model_inputs = [next_token.unsqueeze(0)]
+    for k, v in past_key_values:
+        model_inputs.append(k)
+        model_inputs.append(v)
+
+
+    # Run decode stage on TT device and generate tokens by passing the last predicted token and the past key values.
```
dgolubovicTT, line 235: Can we put this decode inside one method (decode_on_device) just to avoid duplicating code?

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
...
+                past_key_values.append(tuple([past_key_values_list[idx], past_key_values_list[idx + 1]]))
+            model_outputs = self.model(input_ids=inputs[0], past_key_values=tuple(past_key_values))
+            outputs = [model_outputs[0]]
+            for k, v in model_outputs[1]:
+                outputs.append(k)
+                outputs.append(v)
+            return outputs
+
+    llama_model= Llama_Model_Wrapper(framework_model)
+    llama_model.eval()
```
dgolubovicTT, line 213: I think we don't need this llama_model.eval() since load_model already sets mode to eval...

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
...
+    # Run decode stage on TT device and generate tokens by passing the last predicted token and the past key values.
+    # untill the a specified maximum number of new tokens is reached or an end-of-sequence token is encountered.
+    max_new_tokens = 32
+    for _ in range(max_new_tokens):
+
+        # CPU Inference
+        framework_output = llama_model(model_inputs)
+
+        # Compile the model
+        compiled_model = forge.compile(llama_model, sample_inputs=[model_inputs])
```
dgolubovicTT, line 244: Here we compile model in each iteration... @nvukobratTT this will be really slow? Should we fix some max input size ?

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
...
+        # Validate TT result with Framework
+        assert all([compare_with_golden_pcc(golden=fw_out, calculated=tt_out, pcc=0.99) for fw_out, tt_out in zip(framework_output, tt_output)])
+        next_token_logits = framework_output[0][:, -1, :]
+        next_token = torch.argmax(next_token_logits, dim=-1)
+
+        if next_token == tokenizer.eos_token_id:
+            break
+        else:
+            model_inputs = [next_token.unsqueeze(0)]
+            model_inputs.extend(framework_output[1:])
```
dgolubovicTT, line 259: Should we use here tt_output instead of framework_output?

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
     # Generated text
     generated_text = tokenizer.decode(generated_tokens[0], skip_special_tokens=True)
     print(generated_text)
+
+
+@pytest.mark.xfail()
+def test_llama_inference_no_cache_tt():
```
nvukobratTT, line 133: Let's outline in test name that it's decode on tt, prefill is on CPU

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
...
+
+    # Load Llama 3B model and tokenizer
+    model_path = "openlm-research/open_llama_3b"
```
nvukobratTT, line 136: Please check few comments on this PR as well:
- https://github.com/tenstorrent/tt-forge-fe/pull/372

More precisely, maybe we can unify this `load_model` function to be a bit more generic :)) 

@dgolubovicTT please provide your insights as well based on Prefil on TT comments :))

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
...
+    # untill the a specified maximum number of new tokens is reached or an end-of-sequence token is encountered.
+    max_new_tokens = 32
+    for _ in range(max_new_tokens):
+
+        # CPU Inference
+        framework_output = framework_model(input_ids)
+
+        # Compile the model
+        tt_inputs = [input_ids]
+        compiled_model = forge.compile(framework_model, sample_inputs=tt_inputs)
```
nvukobratTT, line 164: Do we need to compile the model in the loop? Instead, let's try to use fixed max seq len and compile model only once :))

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
...
+
+@pytest.mark.xfail()
+def test_llama_inference_cache_tt():
+
+    # Load Llama 3B model and tokenizer
+    model_path = "openlm-research/open_llama_3b"
+    framework_model = load_model(model_path, use_cache=True)
+    tokenizer = LlamaTokenizer.from_pretrained(model_path)
+
+    class Llama_Model_Wrapper(torch.nn.Module):
```
nvukobratTT, line 196: Let's document this wrapper and explain how PK influences need for it :)) 

Also, for naming, let's remove `_` to have a better fit with rest of the Python codebase

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
...
+        def __init__(self, model):
+            super().__init__()
+            self.model = model
+        def forward(self, inputs):
+            past_key_values_list = inputs[1:]
+            past_key_values = list()
+            for idx in range(0, len(past_key_values_list), 2):
```
nvukobratTT, line 203: Make sure to explain new PK formating in wrapper description :))

forge/test/mlir/llama/test_llama_inference.py
nvukobratTT: I agree, we should avoid compiling in the loop (also posted comment for example above).

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
...
+    logits, past_key_values = framework_model(input_ids=inputs.input_ids, attention_mask=inputs.attention_mask)
+    next_token_logits = logits[:, -1, :]
+    next_token = torch.argmax(next_token_logits, dim=-1)
+
+    generated_tokens = inputs.input_ids
+    generated_tokens = torch.cat(
+        [generated_tokens, next_token.unsqueeze(0)], dim=-1
+    )
+
+    model_inputs = [next_token.unsqueeze(0)]
```
nvukobratTT, line 229: One general question, do we need to bundle the KV cache with the next token in the same input? 

Can we have single input that will be next token, and another that will represent the past cache?

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,141 @@ def test_llama_inference_cache_cpu():
...
+        tt_output = [tt_out.to("cpu") for tt_out in tt_output]
+
+        # Validate TT result with Framework
+        assert all([compare_with_golden_pcc(golden=fw_out, calculated=tt_out, pcc=0.99) for fw_out, tt_out in zip(framework_output, tt_output)])
+        next_token_logits = framework_output[0][:, -1, :]
+        next_token = torch.argmax(next_token_logits, dim=-1)
+
+        if next_token == tokenizer.eos_token_id:
+            break
+        else:
```
nvukobratTT, line 257: Do we need to place in `else` block? We can just continue this logic without it as `if` condition contains only `break` :))

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: I will remove the llama_model.eval() since it present in load_model function as you mentioned

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: I Will change from framework_output to tt_output

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Sure

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Sure

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Will update it

forge/test/mlir/llama/test_llama_inference.py
dgolubovicTT: I would suggest to edit load_model method to return tokenizer too! so that. I will do that within https://github.com/tenstorrent/tt-forge-fe/pull/372.

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: The current push doesn't include, decode_on_device method. I will create a method and update the code.

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Removed llama_model.eval() as you mentioned load_model is already setting model to eval mode

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Changed framework_output to tt_output.

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Renamed 
`test_llama_inference_no_cache_tt` -> `test_llama_prefill_on_cpu_decode_on_tt_no_cache`
`test_llama_inference_cache_tt` -> `test_llama_prefill_on_cpu_decode_on_tt_cache`

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Replicated the change from PR https://github.com/tenstorrent/tt-forge-fe/pull/372 for load_model with minor change in the function name from `load_model` to `load_model_and_tokenizer`

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Instead of compile the model in loop, create a input_ids with max seq len by applying padding and passing input with max seq len and attention mask (i.e to avoid performing attention on padding token indices).

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Updated the naming and wrapper model doc.

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Removed the input formating because of unbundling input and past key values as two inputs to the model forward function and added the comment for output formating.

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Unbundled the input token and past key values to make it as two separate inputs to the wrapper forward function

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Removed the else block

chandrasekaranpradeep:
> @nvukobratTT @chandrasekaranpradeep since this PR is correlated to #372 can we merge them one by one just to avoid some duplicating code, and make everything well organized?

Sure @dgolubovicTT 
Can we change  the load_model  function name from `load_model` to `load_model_and_tokenizer` because  we are returning model and tokenizer

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: @nvukobratTT @dgolubovicTT 

1. **test_llama_prefill_on_cpu_decode_on_tt_no_cache** - Modified the test to compile outside the loop by making the input with max fixed size (i.e applying padding in the input_ids and additionally passing attention_mask to avoid attention on padding token indicies)

 2. **test_llama_prefill_on_cpu_decode_on_tt_cache** - In the decode, we pass the last predicted token of shape `(batch_size, 1)` **eg: (1, 1)** and past key values of `List[List[torch.Tensor, torch.Tensor]]` and past key values tensor shape will be `(batch_size, num_of_key_values_heads, key_value_seq_len, head_dim)` **eg: (1, 32, 12, 100)** to the LlamaModelWrapper model and the output will be logits of shape (batch_size, 1, vocab_size) **eg: (1, 1, 32000)** and past key values of `List[List[torch.Tensor, torch.Tensor]]` but the returned past key values tensor shape will be `(batch_size, num_of_key_values_heads, key_values_seq_len + 1, head_dim)` **eg: (1, 32, 13, 100)**  Here, you can see in the past key values tensor, key_value_seq_len is changed from 12 to 13 in the [LlamaAttention module](https://github.com/huggingface/transformers/blob/514de24abfd4416aeba6a6455ad5920f57f3567d/src/transformers/models/llama/modeling_llama.py#L374C9-L390C105) by concatenating the current key and value state into the past key values. 
 In this case, how we can compile the model  outside the loop, because in every iteration the key_value_seq_len will changed to key_value_seq_len + 1?

nvukobratTT: Will take a look into the changed PR a bit later, but regarding 2. point, can we also pad KV tensors to the max seq len we defined for inputs?

dgolubovicTT: I will look at this PR again later today...

chandrasekaranpradeep: I have tried adding padding to the past key values tensor in key_values_seq_len dimension but the generated tokens are different comparing to the past key values tensor without padding.

**Text generated in decode phase:** 

1. Without Past key values padded:
```
Q: What is the largest animal?
A: The blue whale.
Q: What is the largest animal?
A: The blue whale. It is the largest animal on Earth. It is also the largest
```

2. With Past key values padded:
```
Q: What is the largest animal?
A: The largest animal is the blue whale.
Q:
A: What is the largest animal?
A:
A:
A: The blue whale.
```

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -126,3 +127,293 @@ def test_llama_inference_cache_cpu():
...
+            self.model = model
+
+        def forward(self, input_id, past_key_values):
+            model_outputs = self.model(input_ids=input_id, past_key_values=past_key_values)
+
+            # TT will return the logits and past key values as list of tensor, so flattening
+            # the past key values from List(List(Key1, Values1), ... , List(Key26, Values26)) to
+            # List(Key1, Values1, ... , Key26, Values26) for comparing the Framework and TT output in similar fashion.
+            outputs = [model_outputs.logits]
+            for k, v in model_outputs.past_key_values:
```
nvukobratTT, line 229: Will this post-processing be included in the compiled model?

If that is the case, we should we do this post-processing outside of this wrapper?

forge/test/mlir/llama/test_llama_inference.py
dgolubovicTT: I am looking into how we can solve this. I have read the llama source code [LlamaAttention module](https://github.com/huggingface/transformers/blob/514de24abfd4416aeba6a6455ad5920f57f3567d/src/transformers/models/llama/modeling_llama.py#L374C9-L390C105) and I understand the problem. Will update here when I have some solution... @chandrasekaranpradeep can you provide some branch where you added padding so I can start from there...

dgolubovicTT: @chandrasekaranpradeep what values are you using for padding?

dgolubovicTT: So padding with 0 should work for calculating attention. However, we have to update kv cache differently than Llama is doing. They are concatenating old K and V states with new. Since out K and V are now constant length = max_seq_len we have to insert new k and v vectors at the end of sequence...

nvukobratTT: > @chandrasekaranpradeep can you provide some branch where you added padding so I can start from there...

Hey @dgolubovicTT, it's already there on PR. Check out this function:
```
test_llama_prefill_on_cpu_decode_on_tt_cache_padding
```

> So padding with 0 should work for calculating attention. However, we have to update kv cache differently than Llama is doing. They are concatenating old K and V states with new. Since out K and V are now constant length = max_seq_len we have to insert new k and v vectors at the end of sequence...

Yea, we'll probably have to manage KV pairs for every iteration in order to populate them in a proper way..

nvukobratTT: Let's see if we can figure out the solution with a current implementation in the next few days. If not, let's wait for other TTNN issues to get resolved, and then push Llama without past cache for a first run. After we finish with ResNet, we should look over PK once again :))

forge/test/mlir/llama/test_llama_inference.py
chandrasekaranpradeep: Sure Nikola. I will move the post processing outside the wrapper.

chandrasekaranpradeep: Moved the  post-processing outside the wrapper.

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -120,3 +121,346 @@ def test_llama_inference_cache_cpu():
...
+@pytest.mark.xfail()
+def test_llama_prefill_on_cpu_decode_on_tt_cache():
+
+    # Load Llama 3B model and tokenizer
+    model_path = "openlm-research/open_llama_3b"
+    framework_model, tokenizer = load_model_and_tokenizer(
+        model_path=model_path, use_cache=True, return_dict=True
+    )
+
+    class LlamaModelWrapper(torch.nn.Module):
```
dgolubovicTT, line 219: Can we move this class deffinition out of the test...

chandrasekaranpradeep: Sure @dgolubovicTT

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -120,3 +121,346 @@ def test_llama_inference_cache_cpu():
...
+
+    # Prepare input sentence
+    prompt = "Q: What is the largest animal?\nA:"
+    inputs = tokenizer(prompt, return_tensors="pt")
+
+    # Run Prefill on CPU with cache to get the initial logits and past key-values
+    prefill_output = framework_model(
+        input_ids=inputs.input_ids, attention_mask=inputs.attention_mask
+    )
+    next_token_logits = prefill_output.logits[:, -1, :]
```
dgolubovicTT, line 251: @chandrasekaranpradeep @nvukobratTT I think here is the problem. We take next token logits as last embeding of prefil_output. Actually this will be embeding of last padded token... We should take logit on index: seq_size - 1. Do you agree?

chandrasekaranpradeep: In prefill stage, we are only passing the input_ids and attention_mask which doesn't include any pad_token because we are not setting any max_seq_len in tokenizer.

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -120,3 +121,346 @@ def test_llama_inference_cache_cpu():
...
+    generated_tokens = torch.cat([generated_tokens, next_token.unsqueeze(0)], dim=-1)
+
+    past_key_values_list = [[k, v] for k, v in prefill_output.past_key_values]
+
+    model_inputs = [next_token.unsqueeze(0), past_key_values_list]
+
+    # Padding on past key values can be enabled by setting apply_padding_on_past_key_values to True.
+    apply_padding_on_past_key_values = True
+    max_new_tokens = 32
+    if apply_padding_on_past_key_values:
```
chandrasekaranpradeep, line 372: @dgolubovicTT  We are only applying zero padding in past key values in key_value_seq_length dimension before passing to the decoder stage input.

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -120,3 +121,346 @@ def test_llama_inference_cache_cpu():
...
+                key_state = torch.clone(model_inputs[1][idx][0][:, :, -1, :])
+                value_state = torch.clone(model_inputs[1][idx][1][:, :, -1, :])
+                model_inputs[1][idx][0][
+                    :, :, non_padding_seq_len + max_new_tokens_idx, :
+                ] = model_inputs[1][idx][0][:, :, -1, :]
+                model_inputs[1][idx][0] = model_inputs[1][idx][0][:, :, :-1, :]
+                model_inputs[1][idx][1][
+                    :, :, non_padding_seq_len + max_new_tokens_idx, :
+                ] = model_inputs[1][idx][1][:, :, -1, :]
+                model_inputs[1][idx][1] = model_inputs[1][idx][1][:, :, :-1, :]
```
chandrasekaranpradeep, line 457: Here, I am taking the last key_states and values_states that are concatenated at the end of the past key values in LlamaAttention to the first non padding index in past key values.

dgolubovicTT: Nice

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -120,3 +121,239 @@ def test_llama_inference_cache_cpu():
...
+        # tt_output = compiled_model(*model_inputs)
+        # tt_output = [tt_out.to("cpu") for tt_out in tt_output]
+
+        # # Validate TT result with Framework
+        # assert all([compare_with_golden_pcc(golden=fw_out, calculated=tt_out, pcc=0.99) for fw_out, tt_out in zip(framework_output, tt_output)])
+
+        logits = framework_output[0]
+        past_key_values = framework_output[1:]
+
+        next_token_logits = logits[:, -1, :]
```
dgolubovicTT, line 320: Next token should be calculated based on logits on index `non_padding_seq_len + max_new_tokens_idx`, not last -1. Because last logits correspond to padding tokens in this case. This would be ok if we didn't have padding. Then, last token would be on -1 but now it isn't.

dgolubovicTT:
Hey guys, I have looked into this and found few bugs. 
First one, thanks to @chandrasekaranpradeep's help. When we pad K matrix, the sequence length model sees is increased so rotary embedings are different for all tokens. The longer the sequence (no matter if we padded with zeros), the smaller are rotation angles...

Second one, during decode we have to provide modified attention mask that is padded. It should have -inf in places of padded tokens. If this is not used K*q will have 0 on places where K is padded, but when softmax is calculated this zeros will disturb the distribution for non padding tokens. 

The things get complicated because new k values are appended on K matrix inside attention so it is just not enough to move new k vector to its place after the forward pass. So padding is heavily influencing rotary embedings and attention mask...

@nvukobratTT lets talk offline today before making any further action...

PR #394, opened 2024-10-09
Collect workflow run data and test results

vmilosevic:
We want to start collecting workflow run data and test results for analytics. This is the initial version of the data collection pipeline, based on what tt-metal team is already running for their project.

The workflow starts by downloading GH data for workflow runs, jobs, and artifacts containing test reports. It then calls scripts from the infra folder to transform that data into the JSON format agreed upon with the data team and uploads it to the SFTP server.

GH has a few limitations and we had to add workarounds for those. One is that there is no way to connect artifacts with job that produced them, only to the parent workflow. Another is that current job id is not accessible from job context. The workaround is to get the job ud using GH API and to include it in a artifact filename.

One additional change is to include start_timestamp and end_timestamp in the test report metadata.


nvukobratTT (approved):
Let's test it out :)))

pilkicTT (approved):

Issue #395, opened 2024-10-09
[training] configurable recompute


pilkicTT:
After #366, we will have basic setup for training with recompute. However, the implemented recompute works on the whole graph only, and creates recompute nodes for each of the ops in the forward pass.

For this to be useful, we at least need a way to manually define ops (chain of ops) to be recomputed.

pilkicTT:
After this is implemented, the next step would be to have logic for automatically deciding what to recompute.

PR #396, opened 2024-10-09
[autograd] moving recompute to autograd

### LINKED ISSUE #366: [trainign

pilkicTT:
Moving `insert_recompute_ops()` under `autograd_engine`. Toggling recompute on/off is done via `autograd_config`.

Also, fixes a bug in split graph pass, uncovered by the split graph unit tests with recompute=on.

Closes #366


forge/csrc/autograd/autograd.cpp
```diff
@@ -484,6 +484,8 @@ Graph *autograd_engine::run()
     create_backward_graph(requires_grad_map);
     create_optimizer_graph();
 
+    insert_recompute_ops();
```
nvukobratTT, line 487: Does it make sense to have an `if` statement here to determine if we should insert recompute ops?

Just by briefly passing through the `Graph *autograd_engine::run()` function, on the first go it would seem like we're running a recompute pass each time. In this case, it isn't too big of an issue, but if the amount of passes increases for autograd, it'll be a bit trickier to determine what is run, and what is not.

forge/csrc/autograd/autograd.cpp
```diff
@@ -719,6 +721,152 @@ NodeContext autograd_engine::create_input(
...
+            fwd_to_bwd_edge.consumer_node_id,
+            fwd_to_bwd_edge.consumer_input_port_id,
+            fwd_to_bwd_edge.edge_type),
+        graph->get_edge_attributes(fwd_to_bwd_edge));
+    graph->remove_edge(fwd_to_bwd_edge);
+}
+
+void autograd_engine::insert_recompute_ops()
+{
+    if (!config.recompute)
```
nvukobratTT, line 795: Is this controlled using the `enable_recompute` flag?

forge/csrc/autograd/autograd.cpp
```diff
@@ -719,6 +721,152 @@ NodeContext autograd_engine::create_input(
...
+    }
+
+    std::unordered_map<Node *, Node *> forward_to_recompute;
+    std::vector<Node *> topo_order = graphlib::topological_sort(*graph);
+
+    // We loop through only the backward nodes and look for operand nodes that are marked
+    // as FWD. Each instance is an opportunity to perform a potential recompute on the original
+    // FWD op.
+    //
+    // TODO(#395): extend this to support configurable op-specific recompute; rather than an all-or-nothing sort-of
```
nvukobratTT, line 807: Like it! Maybe we can create one more issue to try to figure out "automagically" which part of the graph should be recomputed? E.g. based on potential DRAM vs Compute usage (explorational, maybe even for internship :))

forge/csrc/autograd/autograd.cpp
```diff
@@ -719,6 +721,152 @@ NodeContext autograd_engine::create_input(
...
+    // deal.
+    std::deque<std::string> recompute_node_names;
+
+    for (Node *node : topo_order)
+    {
+        if (node->is_forward() && node->node_type() == graphlib::NodeType::kPyOp)
```
nvukobratTT, line 813: Do you see the value of adding cleaner APIs for node type as well? 

Not a requirement for this change ofc. But if it make sense, let's open up issue to tackle those as well :))

forge/csrc/autograd/autograd.cpp
```diff
@@ -719,6 +721,152 @@ NodeContext autograd_engine::create_input(
     return NodeContext(node);
 }
 
+static Node *create_recompute_op(
```
nvukobratTT, line 724: Can we add a brief doc regarding how this function works? E.g., focusing on mapping fwd node(s) and replacing specific bwd node.

forge/csrc/autograd/autograd.cpp
```diff
@@ -719,6 +721,152 @@ NodeContext autograd_engine::create_input(
...
+        }
+    }
+
+    std::unordered_map<Node *, std::string> forward_to_recompute_name;
+    for (const auto [fwd_node, recompute_node] : forward_to_recompute)
+    {
+        forward_to_recompute_name[fwd_node] = recompute_node->name();
+    }
+
+    while (not recompute_node_names.empty())
```
nvukobratTT, line 840: Let's add an overview comment on this and the following code section.

forge/csrc/autograd/autograd.cpp
pilkicTT: We'll see, there are definitely changes that need to be made to update the `Graph` and `Node` structures to better reflect the current world.

forge/csrc/autograd/autograd.cpp
pilkicTT: Yup...

forge/csrc/autograd/autograd.cpp
pilkicTT: The `if` is inside the function.

Not sure if I understand, but `autograd_engine::run()` is called only once during compile on the main graph...

forge/csrc/autograd/autograd.cpp
pilkicTT: I didn't want to create additional issues, since at the moment it is not clear whether we want to pursue this path. This is just for note keeping/future work.

I have added a comment in the already created issue #395, to consider automatic recompute as a next step...

vladimirjovanovicTT: What @nvukobratTT is definitely a long-term plan which we are aware of. I would agree with @pilkicTT to not open an issue at this point to avoid confusion regarding issue prioritization.

forge/csrc/autograd/autograd.cpp
```diff
@@ -719,6 +721,161 @@ NodeContext autograd_engine::create_input(
     return NodeContext(node);
 }
 
+// From the node in the forward graph, create a recompute node for the backward graph.
+// This will also connect the newly created recompute node to its operands (based on the operands of the fwd node).
+// NOTE: it is expected that the create_recompute_op() is called in topological order.
+static Node *create_recompute_op(
+    graphlib::Graph *graph, Node *fwd_node, std::unordered_map<Node *, Node *> &fwd_to_recompute)
+{
+    graphlib::PyOpNode *recompute_node = graph->add_node(
```
dgolubovicTT, line 730: Just a thought here - maybe we can put these few lines of creating recompute node inside method create_recompute_node. I remember when we worked with buff queues we had few places in stack where we had similar code for creating buffering queues and it was chaotic. OFC recompute node is hopefully created only here so this shouldn't be the case...

dgolubovicTT (approved):

vladimirjovanovicTT (approved):
LGTM

forge/csrc/autograd/autograd.cpp
nvukobratTT: What do you think about doing this:
```cpp
    if (!config.recompute)
    {
        insert_recompute_ops();
    }
```

Looking at other passes it's easier to preview what is called when by first look (e.g. no need to check every function that is called).

nvukobratTT: Not a hard requirement, this is just a personal preference.

nvukobratTT (approved):

forge/csrc/autograd/autograd.cpp
pilkicTT: Done!

forge/csrc/autograd/autograd.cpp
pilkicTT: Yeah, I see your point with the buffering queues, but the recompute is not a special kind of node. And it is tightly coupled with autograd, so I don't think there should be any problems like with buffering queues.

However, I will modify this code to simply clone the forward node and change its epoch; instead of creating and adding the recompute node manually.

PR #398, opened 2024-10-10
Initial CI for Perf Benchmark Tests

vcanicTT:
Add the initial infrastructure for tracking the benchmark models on CI. This initial infrastructures includes separate benchmark test for MNIST Linear and bunch of pytest runs used to variate different parameters on MNIST.

The "old" benchmark folder, inherited from the previous project, is moved to 'benchmark_lagacy' folder. We still need it as a model for the new one.


forge/test/benchmark/benchmark/utils/utils.py
nobradovictt, line 1: Is this file empty?

nobradovictt (approved):

nvukobratTT (commented):
@vcanicTT what do you think about pushing the changes only related to MNIST in this PR? 

Don't we have a benchmarking reference in the old PyBuda repo? is there some benefit of having it copy-pased here as well?

nvukobratTT (changes requested):
@vcanicTT what do you think about pushing the changes only related to MNIST in this PR? 

Don't we have a benchmarking reference in the old PyBuda repo? is there some benefit of having it copy-pased here as well?

forge/test/benchmark/benchmark/utils/utils.py
vcanicTT: Yes, it is. I'll remove it.

vcanicTT:
> @vcanicTT what do you think about pushing the changes only related to MNIST in this PR?
> 
> Don't we have a benchmarking reference in the old PyBuda repo? is there some benefit of having it copy-pased here as well?

It is easier to make new one from scratch, because there is a big difference how we call the models and organize CI files.

nvukobratTT:
> > @vcanicTT what do you think about pushing the changes only related to MNIST in this PR?
> > Don't we have a benchmarking reference in the old PyBuda repo? is there some benefit of having it copy-pased here as well?
> 
> It is easier to make new one from scratch, because there is a big difference how we call the models and organize CI files.

If that is the case can we remove the `benchmarking_legacy` directory, and add parts of logic bit by bit to the newly defined `benchmarking` directory? :))

vcanicTT:
> > > @vcanicTT what do you think about pushing the changes only related to MNIST in this PR?
> > > Don't we have a benchmarking reference in the old PyBuda repo? is there some benefit of having it copy-pased here as well?
> > 
> > 
> > It is easier to make new one from scratch, because there is a big difference how we call the models and organize CI files.
> 
> If that is the case can we remove the `benchmarking_legacy` directory, and add parts of logic bit by bit to the newly defined `benchmarking` directory? :))

That's what we are doing, now, but it would be nice if we can keep legacy directory as illustration until we add all necessary models from it.

tt-mpantic:
> > > > @vcanicTT what do you think about pushing the changes only related to MNIST in this PR?
> > > > Don't we have a benchmarking reference in the old PyBuda repo? is there some benefit of having it copy-pased here as well?
> > > 
> > > 
> > > It is easier to make new one from scratch, because there is a big difference how we call the models and organize CI files.
> > 
> > 
> > If that is the case can we remove the `benchmarking_legacy` directory, and add parts of logic bit by bit to the newly defined `benchmarking` directory? :))
> 
> That's what we are doing, now, but it would be nice if we can keep legacy directory as illustration until we add all necessary models from it.

What Nikola is proposing is to keep that folder in gitlab where it currently sits, and just take parts when we need them...
No need to push entire folder to main at this point of time...

nvukobratTT:
> > > > @vcanicTT what do you think about pushing the changes only related to MNIST in this PR?
> > > > Don't we have a benchmarking reference in the old PyBuda repo? is there some benefit of having it copy-pased here as well?
> > > 
> > > 
> > > It is easier to make new one from scratch, because there is a big difference how we call the models and organize CI files.
> > 
> > 
> > If that is the case can we remove the `benchmarking_legacy` directory, and add parts of logic bit by bit to the newly defined `benchmarking` directory? :))
> 
> That's what we are doing, now, but it would be nice if we can keep legacy directory as illustration until we add all necessary models from it.

Is there a benefit of keeping legacy as an illustration of the Forge project, while we have it on Buda? 

We strive to clean all Buda legacy code/limitations/concepts, and this seems like a small anti-pattern :)) Let me know what you think, maybe I'm missing some key point..

vcanicTT:
> > > > > @vcanicTT what do you think about pushing the changes only related to MNIST in this PR?
> > > > > Don't we have a benchmarking reference in the old PyBuda repo? is there some benefit of having it copy-pased here as well?
> > > > 
> > > > 
> > > > It is easier to make new one from scratch, because there is a big difference how we call the models and organize CI files.
> > > 
> > > 
> > > If that is the case can we remove the `benchmarking_legacy` directory, and add parts of logic bit by bit to the newly defined `benchmarking` directory? :))
> > 
> > 
> > That's what we are doing, now, but it would be nice if we can keep legacy directory as illustration until we add all necessary models from it.
> 
> What Nikola is proposing is to keep that folder in gitlab where it currently sits, and just take parts when we need them... No need to push entire folder to main at this point of time...

The folder is already there, I just renamed it.

vcanicTT:
> > > > > > > @vcanicTT what do you think about pushing the changes only related to MNIST in this PR?
> > > > > > > Don't we have a benchmarking reference in the old PyBuda repo? is there some benefit of having it copy-pased here as well?
> > > > > > 
> > > > > > 
> > > > > > It is easier to make new one from scratch, because there is a big difference how we call the models and organize CI files.
> > > > > 
> > > > > 
> > > > > If that is the case can we remove the `benchmarking_legacy` directory, and add parts of logic bit by bit to the newly defined `benchmarking` directory? :))
> > > > 
> > > > 
> > > > That's what we are doing, now, but it would be nice if we can keep legacy directory as illustration until we add all necessary models from it.
> > > 
> > > 
> > > What Nikola is proposing is to keep that folder in gitlab where it currently sits, and just take parts when we need them... No need to push entire folder to main at this point of time...
> > 
> > 
> > The folder is already there, I just renamed it.
> 
> PR says it's added, not renamed..
> 
> If you check latest main, you'll see that benchmarking repo is removed:
> 
> * https://github.com/tenstorrent/tt-forge-fe/tree/main/forge/test

Now, I see, because I worked on old commit and the folder still was there. But, I don't understand why you removed the folder, because I don't anything about that, and because perf team worked on it.

forge/test/benchmark/benchmark/utils/utils.py
vcanicTT: Done.

vcanicTT:
> > > > > > > > @vcanicTT what do you think about pushing the changes only related to MNIST in this PR?
> > > > > > > > Don't we have a benchmarking reference in the old PyBuda repo? is there some benefit of having it copy-pased here as well?
> > > > > > > 
> > > > > > > 
> > > > > > > It is easier to make new one from scratch, because there is a big difference how we call the models and organize CI files.
> > > > > > 
> > > > > > 
> > > > > > If that is the case can we remove the `benchmarking_legacy` directory, and add parts of logic bit by bit to the newly defined `benchmarking` directory? :))
> > > > > 
> > > > > 
> > > > > That's what we are doing, now, but it would be nice if we can keep legacy directory as illustration until we add all necessary models from it.
> > > > 
> > > > 
> > > > What Nikola is proposing is to keep that folder in gitlab where it currently sits, and just take parts when we need them... No need to push entire folder to main at this point of time...
> > > 
> > > 
> > > The folder is already there, I just renamed it.
> > 
> > 
> > PR says it's added, not renamed..
> > If you check latest main, you'll see that benchmarking repo is removed:
> > 
> > * https://github.com/tenstorrent/tt-forge-fe/tree/main/forge/test
> 
> Now, I see, because I worked on old commit and the folder still was there. But, I don't understand why you removed the folder, because I don't know anything about that, and because perf team worked on it.

Folder removed.

vcanicTT:
@nvukobratTT I removed files, can you now take a look?

pytest.ini
```diff
@@ -24,5 +24,9 @@ testpaths =
     # Llama
     forge/test/mlir/llama/test_llama_inference.py::test_llama_inference
 
+    # Benchmark
+    # MNIST Linear
+    forge/test/benchmark/benchmark/models/mnist_linear.py::test_mnist_linear
```
nvukobratTT, line 29: Do we want to run benchmarking for every commit, or define nightly job to run it? 

Why do we need it here as it's already defined here: `.github/workflows/build-and-test.yml`?

forge/test/benchmark/scripts/run_benchmark_wh.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# SPDX-License-Identifier: Apache-2.0
+
+# ------------------------------------------------------------------------------------------------------------------------------------------------------------ #
+# Models we run on Wormhole B0
+# ------------------------------------------------------------------------------------------------------------------------------------------------------------ #
+
+
+# MNIST Linear
+python forge/test/benchmark/benchmark.py -m mnist_linear -bs 1 -o forge-benchmark-e2e-mnist.json
```
nvukobratTT, line 11: What is the purpose of this file? Local lab runs? 

As you already have test run defined here `.github/workflows/build-and-test.yml`, is it better to encapsulate this somehow? E.g. when you update how benchmarking model is run, you update single spot, instead of multiple :))

forge/test/benchmark/benchmark/models/mnist_linear.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+# @TODO - For now, we are skipping these parameters, because we are not supporting them
+# @pytest.mark.parametrize("math_fidelity", MATH_FIDELITY, ids=[f"math_fidelity={item}" for item in MATH_FIDELITY])
+# @pytest.mark.parametrize("dataformat", DATAFORMAT, ids=[f"dataformat={item}" for item in DATAFORMAT])
+# @pytest.mark.parametrize("arch", ARCH, ids=[f"arch={item}" for item in ARCH])
+@pytest.mark.parametrize("hidden_size", HIDDEN_SIZE, ids=[f"hidden_size={item}" for item in HIDDEN_SIZE])
+@pytest.mark.parametrize("input_size", INPUT_SIZE, ids=[f"input_size={item}" for item in INPUT_SIZE])
+@pytest.mark.parametrize("batch_size", BATCH_SIZE, ids=[f"batch_size={item}" for item in BATCH_SIZE])
+def test_mnist_linear(
```
nvukobratTT, line 79: Just a general note, for PoC this is okay. However, please think about how this can be integrated with regular model demos :))

.github/workflows/build-and-test.yml
```diff
@@ -87,6 +87,12 @@ jobs:
         name: test-log-${{ matrix.build.runs-on }}
         path: pytest.log
 
+    - name: Run Perf Benchmark
```
nvukobratTT, line 90: Should we run these on a nighly basis, rather than every commit?

pytest.ini
vcanicTT: For now, yes, it is small model, we're a still building infrastructure. This is not real-world model, maybe I remove it at all. This is initial commit, I don't know if we can even download files, preserve results, compare them, many things I didn't put into the test, like bigger batch, data format, many things don't work in MLIR, yet, this only starting point to see if we can run CI in a way we want. We will iterate. Vlada Milosevic is also working on many things necessary to make this to work end-to-end.

forge/test/benchmark/scripts/run_benchmark_wh.sh
vcanicTT: The idea is to be global, but I didn't figure out way if something hangs, so for now is only local. Also, it is nice to have for manual runs.

.github/workflows/build-and-test.yml
vcanicTT: Answered above. We're working on that. We still don't have any big model. When we add first bigger model like we will change that. Also, mnist is good to track all changes within the compiler, it's not big, and we can catch irregularities much faster.

forge/test/benchmark/benchmark/models/mnist_linear.py
vcanicTT: This will not be part of model demos, this is parametrized so we can track performance and how the optimizer acts. And probably it will exist only for MNIST, or some smaller models, because this amount of tests we can't run for any bigger model. Maybe, we can have it for some smaller parts of benchmarking models, but I will develop this idea further.

forge/test/benchmark/scripts/run_benchmark_wh.sh
vcanicTT: Also, these tests are not the same one is benchmark and the second one is pytest parameterized. In this case the first one is calling the other one, but that don't need to be the case for every model we added.

forge/test/benchmark/benchmark/models/mnist_linear.py
nvukobratTT: Sure, my point is once you start to scale this, it'll be good to also solve issues we had on the previous (Buda) stack :)) 

Approving this PR, thanks for the details @vcanicTT! :))

nvukobratTT (approved):

PR #399, opened 2024-10-10
Squash uplift changes

vmilosevic:
When auto-merging nightly uplift use "squash and merge"


pilkicTT (approved):
We should follow the next auto PR to see if it works, since I am not sure if this will prompt for a new commit message? Because, when you manually squash the PR, github prompts for the new message...

Issue #342, opened 2024-09-26
Implement missing TestDevice


vbrkicTT:
TestDevice class is atm completely missing but needed for specifying a target test device

skhatriTT:
@vbrkicTT What is our status on this one?

vbrkicTT:
> [@vbrkicTT](https://github.com/vbrkicTT) What is our status on this one?

Hi @skhatriTT sorry for slow response, I missed your comment.

It is still the same, class TestDevice is commented. https://github.com/tenstorrent/tt-forge-fe/blob/5710001647557336f09957922f44b8d3fcaa151c/forge/test/conftest.py#L304

However I don't know if it's needed. Maybe @nvukobratTT can confirm if it's still valid.

skhatriTT:
> > [@vbrkicTT](https://github.com/vbrkicTT) What is our status on this one?
> 
> Hi [@skhatriTT](https://github.com/skhatriTT) sorry for slow response, I missed your comment.
> 
> It is still the same, class TestDevice is commented.
> 
> [tt-forge-fe/forge/test/conftest.py](https://github.com/tenstorrent/tt-forge-fe/blob/5710001647557336f09957922f44b8d3fcaa151c/forge/test/conftest.py#L304)
> 
> Line 304 in [5710001](/tenstorrent/tt-forge-fe/commit/5710001647557336f09957922f44b8d3fcaa151c)
> 
>  # class TestDevice: 
> However I don't know if it's needed. Maybe [@nvukobratTT](https://github.com/nvukobratTT) can confirm if it's still valid.

No worries at all @dgolubovicTT, thanks for getting back to me! I'll wait to hear from @nvukobratTT to confirm whether TestDevice is still needed. Once confirmed, I’ll proceed accordingly.

nvukobratTT:
> > > [@vbrkicTT](https://github.com/vbrkicTT) What is our status on this one?
> > 
> > 
> > Hi [@skhatriTT](https://github.com/skhatriTT) sorry for slow response, I missed your comment.
> > It is still the same, class TestDevice is commented.
> > [tt-forge-fe/forge/test/conftest.py](https://github.com/tenstorrent/tt-forge-fe/blob/5710001647557336f09957922f44b8d3fcaa151c/forge/test/conftest.py#L304)
> > Line 304 in [5710001](/tenstorrent/tt-forge-fe/commit/5710001647557336f09957922f44b8d3fcaa151c)
[...]

PR #402, opened 2024-10-11
Warm reset tt device before test

vbrkicTT:
Support warm reset of tt device before test

Issue #239

### DIFF
--- a/forge/test/operators/utils/utils.py
+++ b/forge/test/operators/utils/utils.py
@@ -5,6 +5,7 @@
 # Operator test utilities
 
 import random
+import os
 import sys
 import forge
 import torch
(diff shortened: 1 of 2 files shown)


forge/test/operators/utils/utils.py
```diff
@@ -90,6 +91,15 @@ def set_math_fidelity(math_fidelity: MathFidelity):
         compiler_cfg.default_math_fidelity = math_fidelity
 
 
+class DeviceUtils:
+    '''Utility functions for Forge verification'''
+
+    @staticmethod
+    def warm_reset():
+        reset_command = "/home/software/syseng/wh/tt-smi -lr all wait -er"
```
nvukobratTT, line 99: I recommend to make it a bit more general, because around the corner is new chip type which will potentially have new warm reset command :))

vbrkicTT: If we need to distinguish devices we would have to resolve also #342

nvukobratTT (approved):

Issue #403, opened 2024-10-11
[CI] run pre-commit checks on CI


pilkicTT:
After #393 is completed, add support for running the pre-commit checks on CI.

PR #405, opened 2024-10-11
format everything

pilkicTT:
Running `black` and `clang-format` for all files. Need to do this before enabling pre-commit checks on CI (Issue #403).


nvukobratTT (approved):
Approving right away. No point in reviewing this amount of changes 😅 

If CI passes successfully, I think we should probably merge these changes asap, just to avoid potential merge conflicts.

PR #407, opened 2024-10-11
[ci] add pre-commit checks to CI

### LINKED ISSUE #403: [CI] run pre-commit checks on CI

pilkicTT:
Run pre-commit checks as part of CI.

Closes #403

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,22 @@
+name: Pre-commit
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+jobs:
+  pre-commit:
+
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+      - name: Run pre-commit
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,17 @@
...
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+jobs:
+  pre-commit:
+
+    timeout-minutes: 10
+    runs-on: macos-latest
```
nvukobratTT, line 11: Do we need to run on MacOS specifically? 

I assume we'll have to deprecate it at some point, so maybe it'll be better to use Linux OS flavor

pilkicTT:
Pre-commit workflow is failing for now; it should pass after 
- #405 

is merged...

.github/workflows/pre-commit.yml
pilkicTT: Good catch!

vmilosevic (approved):
Looks good

pilkicTT:
Merging now to start enforcing this ASAP.

Issue #354, opened 2024-10-03
reduce_sum lowering not working properly


pilkicTT:
When running MNIST training with batch > 1, hitting the following error when lowering to mlir:
```
2024-09-30 11:51:47.065 | INFO     | Always          - Emmiting mlir for function backward
loc("bw_in1_add_6_brcst_reduce_sum_0"("backward":4294967295:69)): error: 'ttir.sum' op requires attribute 'keep_dim'
loc("MNISTLinear":0:0): error: module verification failed.
```

Repro:
- checkout 6e2dd7a
- apply the following patch
```
diff --git a/forge/test/mlir/mnist/training/test_training.py b/forge/test/mlir/mnist/training/test_training.py
index 14d4520..1eccda7 100644
--- a/forge/test/mlir/mnist/training/test_training.py
+++ b/forge/test/mlir/mnist/training/test_training.py
@@ -16,7 +16,7 @@ def test_mnist_training():

     # Config
     num_epochs = 3
-    batch_size = 1
+    batch_size = 4
     learning_rate = 0.001

     # Limit number of batches to run - quicker test
```
- run: `pytest -svv forge/test/mlir/mnist/training/test_training.py`

Issue #355, opened 2024-10-03
fuse matmul bias not supported in ttir


pilkicTT:
In cases where we fuse the matmul bias, lowering to mlir fails with the following error:
```
2024-10-03 12:03:30.530 | INFO     | Always          - Emmiting mlir for function forward
python: /localdev/pilkic/tt-forge/third_party/tt-mlir/build/include/ttmlir/Dialect/TTIR/IR/TTIROps.cpp.inc:6556: static void mlir::tt::ttir::MatmulOp::build(::mlir::OpBuilder &, ::mlir::OperationState &, ::mlir::TypeRange, ::mlir::ValueRange, ::llvm::ArrayRef< ::mlir::NamedAttribute>): Assertion `operands.size() == 3u && "mismatched number of parameters"' failed.
Fatal Python error: Aborted
```

Disabling fusing of matmul bias (via env variable, FORGE_NO_FUSE_MATMUL_BIAS=1) fixes the issue.

Repro:
- checkout 6e2dd7a
- apply the following patch (sets batch size for mnist inference to 32)
```
diff --git a/forge/test/mlir/mnist/test_inference.py b/forge/test/mlir/mnist/test_inference.py
index 42fc726..7559370 100644
--- a/forge/test/mlir/mnist/test_inference.py
+++ b/forge/test/mlir/mnist/test_inference.py
@@ -8,7 +8,7 @@ import forge
 from forge.op.eval.common import compare_with_golden_pcc

 def test_mnist_inference():
-    inputs = [torch.rand(1, 784)]
+    inputs = [torch.rand(32, 784)]

     framework_model = MNISTLinear()
     fw_out = framework_model(*inputs)
```
- run: `pytest -svv forge/test/mlir/mnist/test_inference.py`

mstojkovicTT:
opened issue #419 and tenstorrent/tt-mlir/#925

PR #408, opened 2024-10-11
Fixed reduce_sum lowering not working properly in backward graph

ndrakulicTT:
reduce_sum when emited in creation of backward graph did not have named parametar keep_dim.
Added named parametars to the `AutogradContext.op()`. Passed `keep_dim`  for `reduce_sum` for broadcast and element wise backward operations.
Added test_batch_size for inference and training.
Fix for #354


forge/test/mlir/mnist/training/test_training.py
```diff
@@ -91,3 +93,39 @@ def test_mnist_training():
             break
 
     print(f"Test (total) loss: {test_loss}")
+
+@pytest.mark.parametrize("batch_size", [1, 2, 16])
```
nvukobratTT, line 97: Let's add a bit bigger batches that are commonly used as well (e.g. 64)

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -91,3 +93,39 @@ def test_mnist_training():
...
+@pytest.mark.parametrize("in_features", [784])
```
nvukobratTT, line 98: Also, can you move this test into `forge/test/mlir/test_features.py` file?

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -91,3 +93,39 @@ def test_mnist_training():
...
+@pytest.mark.parametrize("out_features", [10])
+def test_batch_size(batch_size, in_features, out_features):
```
nvukobratTT, line 100: Let's create 2 tests:
1. Targets inference with higher batches
2. Targetds training with higher batches

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -91,3 +93,39 @@ def test_mnist_training():
...
+
+    loss_fn = torch.nn.CrossEntropyLoss()
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.001)
+    tt_model = forge.compile(model, sample_inputs=[torch.rand(batch_size, in_features)], loss=loss_fn, optimizer=optimizer)
+
+    optimizer.zero_grad()
+    
+    pred = tt_model(in_data)[0]
+    golden_pred = model(in_data)
+    compare_with_golden(golden_pred, pred)
```
nvukobratTT, line 124: This will silently error out. You need to use `assert` here :))

Same comment for other samples

forge/csrc/autograd/autograd.cpp
```diff
@@ -371,7 +371,7 @@ void autograd_engine::create_backward_graph(const grad_map &requires_grad_map)
 
                     NodeContext src = last_out;
                     last_out = create_op(
-                        OpType("reduce_sum", {dim}),
+                        OpType("reduce_sum", {dim}, {}, {{"keep_dim", true}}),
```
nvukobratTT, line 374: Can we extract `keep_dim` from attributes? 

Getting this data from attributes will make our life easier when we go for support of keep_dim = False :))

Same comment for other occurrences :))

forge/test/mlir/mnist/training/test_training.py
nvukobratTT: Thanks! Can you comment this details on the #355 issue, and notify dev to update this test and check higher batch sizes as well? :))

pilkicTT (changes requested):
The PR title and the description should reflect the actual change. The `keep_dim` problem just emerged from running MNIST training with `batch_size>1`, but I would assume it could be reproduced in different ways.

I know this can sound nitpicky, but having proper context and history about changes in the codebase is of immense help to everybody (including future you). 😄

forge/test/mlir/mnist/training/test_training.py
nvukobratTT: > @nvukobratTT I think it would make more sense if we assert inside of this function...

I'm not sure if that is a better path at the moment..

More precisely, we need to define verification mechanics similar to the one we had on Buda (verification pipeline) where everything is encapsulated (data generation, asserts, etc.). 

In this case, this function is just one of the steps that we'll require there. Thus, having more flexibility around what we do with the output seems more meaningful at this point. 

Do you have some strong opinions on why we should do otherwise?

pilkicTT: No strong opinions, I just prefer this approach, since it reduces chances for silent failures.

It seems to me that we can do both, first encapsulate the assert directly into the `compare_with_golden()` until the verification pipeline is in-place, then do whatever is needed...

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -58,14 +60,14 @@ def test_mnist_training():
             # Forward pass (prediction) on device
             pred = tt_model(data)[0]
             golden_pred = framework_model(data)
-            compare_with_golden(golden_pred, pred)
+            assert compare_with_golden(golden_pred, pred, pcc=0.95)
```
vladimirjovanovicTT, line 63: It would make sense to add a comment regarding choice of pcc value.

forge/test/mlir/test_features.py
```diff
@@ -10,7 +10,7 @@
 from torch import nn
 
 import forge
-from forge.op.eval.common import compare_with_golden_pcc
+from forge.op.eval.common import compare_with_golden_pcc, compare_with_golden
```
vladimirjovanovicTT, line 13: Unrelated to current PR - we should establish a proper way to use compare_with_golden - seems a bit clunky to import both here.

vladimirjovanovicTT (approved):

PR #414, opened 2024-10-14
added tests for fuse_bias, changed default usage

mstojkovicTT:
Added tests for fuse_bias, changed default from using fuse_bias to global param FORGE_FUSE_MATMUL_BIAS (was prevous FORGE_NO_FUSE_MATMUL_BIAS)


forge/csrc/forge_passes.cpp
```diff
@@ -198,7 +198,9 @@ graphlib::Graph* run_pre_lowering_passes(
     recalculate_shapes(graph);
 
     // Fuse bias into matmuls
-    fuse_bias(graph);
+    if (env_as<bool>("FORGE_FUSE_MATMUL_BIAS")){
```
nvukobratTT, line 201: It's okay to keep it disabled by default at the moment. However, let's create an issue with the label `tech dept` and link it here. In the long run, we would like to always fuse bias into matmul for added performance :))

forge/csrc/passes/pre_lowering_passes.cpp
```diff
@@ -366,7 +363,7 @@ void fuse_bias(Graph *graph)
                 broadcast = true;
                 break;
             }
-
```
nvukobratTT, line 369: Make sure to run pre-commit checks as well :)) 

For more details, check this doc segment: https://docs.tenstorrent.com/tt-forge-fe/tools.html

forge/test/operators/matmul/test_matmul_fuse_bias.py
```diff
@@ -0,0 +1,34 @@
+import pytest
```
nvukobratTT, line 1: General comment on import organization (sorted by hierarchy):
1. Native python libraries (e.g. os, math, etc.)
2. Third-party libraries (e.g. torch, numpy, etc.)
3. Our libraries/files (e.g. forge)

Grouping them like this will make our python files a bit more easier to read and follow :))

forge/test/operators/matmul/test_matmul_fuse_bias.py
```diff
@@ -0,0 +1,34 @@
...
+
+import torch
+from torch import nn
+
+import forge
+from forge.op.eval.common import compare_with_golden_pcc
+
+import os
+
+@pytest.mark.parametrize("dims", [
```
nvukobratTT, line 12: These are more shapes as it seems :)) Also make sure to use formatted from pre-commit hooks :))

forge/test/operators/matmul/test_matmul_fuse_bias.py
```diff
@@ -0,0 +1,34 @@
...
+    (1, 784), (4, 784), (16, 784), (32, 784), (64, 784)
+])
+def test_matmul_bias_fusion(dims):
```
nvukobratTT, line 15: Can you also add a test where matmul and bias are fused? 

This will be good to check if your logic works. Also, what you can explore, try to use compile depth from config.py file :)) Using it you'll be able to reduce the compile depth of fused matmul + bias test, and get passing results even if TTIR doesn't supports it :))

P.S.
You can also have single test, but use parameters to toggle bias True/False :))

forge/test/operators/matmul/test_matmul_fuse_bias.py
```diff
@@ -0,0 +1,34 @@
...
+import torch
+from torch import nn
+
+import forge
+from forge.op.eval.common import compare_with_golden_pcc
+
+import os
+
+@pytest.mark.parametrize("dims", [
+    (1, 784), (4, 784), (16, 784), (32, 784), (64, 784)
```
nvukobratTT, line 13: One more advice, as 784 dim is appearing in every case, you can make it as separate pytest param, and only iterate over batch param :))

nvukobratTT:
@mstojkovicTT one general comment, please rebase your changes on top of the latest main, so you can get CI results :))

image

forge/test/operators/matmul/test_matmul_fuse_bias.py
mstojkovicTT: fixed

forge/test/operators/matmul/test_matmul_fuse_bias.py
mstojkovicTT: fixed

forge/test/operators/matmul/test_matmul_fuse_bias.py
mstojkovicTT: done

forge/test/operators/matmul/test_matmul_fuse_bias.py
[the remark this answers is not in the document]
mstojkovicTT: done

forge/csrc/forge_passes.cpp
mstojkovicTT: done!

pilkicTT (approved):

nvukobratTT (approved):

PR #417, opened 2024-10-15
Fix a bug where copy of data json also was uploaded to sftp

vmilosevic:
The Gihub artifacts name convention doesn't allow us to upload files with a colon in the name, so we create a renamed copy. This copy was also uploaded to SFTP by mistake, as we uploaded all pipeline_*.json files. Renaming file copy to pipelinecopy_ to avoid this duplicate upload.

### DIFF
--- a/.github/workflows/produce_data.yml
+++ b/.github/workflows/produce_data.yml
@@ -88,7 +88,7 @@ jobs:
         run: |
           python3 infra/src/generate_data.py --run_id ${{ steps.get-run-id-and-attempt.outputs.run-id }}
           # Workaround: Copy file to avoid GH upload filename limitations
-          cp pipeline_${{ steps.get-run-id-and-attempt.outputs.run-id }}*.json pipeline_${{ steps.get-run-id-and-attempt.outputs.run-id }}.json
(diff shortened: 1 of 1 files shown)


.github/workflows/produce_data.yml
```diff
@@ -88,7 +88,7 @@ jobs:
         run: |
           python3 infra/src/generate_data.py --run_id ${{ steps.get-run-id-and-attempt.outputs.run-id }}
           # Workaround: Copy file to avoid GH upload filename limitations
-          cp pipeline_${{ steps.get-run-id-and-attempt.outputs.run-id }}*.json pipeline_${{ steps.get-run-id-and-attempt.outputs.run-id }}.json
```
pilkicTT, line 91: should this be a `mv`, if we only want to rename the files?

pilkicTT (approved):

.github/workflows/produce_data.yml
vmilosevic: No, we still need to upload the file to SFTP with original name.   

The script will generate file named according to data team spec 
```pipeline_11345421170_2024-10-15T11:50:00.json```  and we are uploading that one to SFTP.
We also want to upload the same file to artifacts, but Github doesn't allow filenames with colon ":", so we create a copy of the same file named 
```pipelinecopy_11345421170.json``` and upload that one to artifacts.

PR #421, opened 2024-10-16
[training] add torch gradient hooks

### LINKED ISSUE #364: [training] add torch hooks for gradients
To avoid having user pass the needed gradients, when executing the loss on CPU, try registering torch hooks for needed gradients and process the grads automatically.

pilkicTT:
Currently, the loss and optimizer parts are ran on the CPU. In order to run backward pass (on the device) we need to know the gradients coming from the `loss.backward()`. Before this change, the gradients from the backward part of the loss function were manually passed to the backward pass by the user.

This change adds appropriate torch hooks which are triggered every time a gradient is computed. The hook will then tie the calculated gradient on the CPU to our runtime, so that it can be passed by the forge runtime to the backward program when needed.

To discern between graph outputs which are used/defined by the user and those which are used by us (to pass information between different graphs/programs), output type enum is introduced. In this space, there is still a lot of refactoring/redesigning to be done, once we get a firmer grasp of what we want/need.

Closes #364


forge/csrc/graph_lib/node_types.hpp
```diff
@@ -304,6 +304,16 @@ class ConstantInputNode : public InputNode
     bool equivalent(const ConstantInputNode *other) const;
 };
 
+enum class OutputType
+{
+    // Internal is used for outputs that are not exposed directly to the user, but used and handled internally for i/o
+    // between different graphs, e.g. passing intermediate values from forward graph to the backward graph.
+    Internal,
```
nvukobratTT, line 311: Do we have any other use case (we know of) next to the passing gradients for bwd pass? 

Having Internal/External output types is a bit fuzzily. For example, having specific output types like "Gradients" can make it cleaner. However, I'm not certain if there are other special use cases that better fit into internal/external usage.

forge/forge/compiled_graph_state.py
```diff
@@ -166,35 +170,54 @@ class ProgramId(IntEnum):
...
         self.intermediates = []
+        self.loss_grad = [None for _ in range(len(fwd_compiled_graph_state.ordered_external_output_names))]
+        self.outputs = {}
+
+    def tie_grad_fn(self, grad_id: int, grad: torch.Tensor):
+        """
+        Hook function to tie the gradients produced by torch as inputs to the backward pass which will be ran on the
+        TT device.
+
+        NOTE: Should be used only when loss is computed on CPU (outside of our runtime).
```
nvukobratTT, line 217: Is it possible to have an assertion if we detect loss on the TT device?

nvukobratTT (approved):
This is pretty much tied to PT, but I think it's okay to do it like this at this point. At least, until we get a better understanding of how we want to structure the training, and later on how (if) we want to include other frameworks.

pilkicTT:
> This is pretty much tied to PT, but I think it's okay to do it like this at this point. At least, until we get a better understanding of how we want to structure the training, and later on how (if) we want to include other frameworks.

Yeah, at this point we are pretty much coupled with torch. The extent of the coupling will be much less when we implement loss and optimizer on the device.

nvukobratTT:
> > This is pretty much tied to PT, but I think it's okay to do it like this at this point. At least, until we get a better understanding of how we want to structure the training, and later on how (if) we want to include other frameworks.
> 
> Yeah, at this point we are pretty much coupled with torch for training. The extent of the coupling will be much less when we implement loss and optimizer on the device.

Do we plan to keep an option for running loss and optimizer on the CPU as well in the long run? If yes, we'll still have framework counseling. However, this is the problem that we don't need to solve in next period for sure :))

forge/csrc/graph_lib/node_types.hpp
pilkicTT: Currently we are using it for passing intermediates from `fwd` to `bwd` programs. In future, for example, it will be also used for passing gradients from the `bwd` to the optimizer program. I'm not sure if differentiating between specific "types of things" we are passing between different programs adds any value.

Maybe I should've chosen a better name for this, instead of `OutputType`? Something like, `OutputScope`, `OutputVisibility`, or something else?

pilkicTT:
> > > This is pretty much tied to PT, but I think it's okay to do it like this at this point. At least, until we get a better understanding of how we want to structure the training, and later on how (if) we want to include other frameworks.
> > 
> > 
> > Yeah, at this point we are pretty much coupled with torch for training. The extent of the coupling will be much less when we implement loss and optimizer on the device.
> 
> Do we plan to keep an option for running loss and optimizer on the CPU as well in the long run? If yes, we'll still have framework counseling. However, this is the problem that we don't need to solve in next period for sure :))

Probably, but in my mind, it is an "edge case" use case, so I think we can be more relaxed (if that is the case) ... 😄 

Anyhow, if we want more robust support, we will probably need to implement it on a framework-by-framework basis...

forge/forge/compiled_graph_state.py
pilkicTT: Not possible, until we implement support for the loss and optimizer on the device. 😄

dgolubovicTT (approved):
Legit!

vladimirjovanovicTT:
> This is pretty much tied to PT, but I think it's okay to do it like this at this point. At least, until we get a better understanding of how we want to structure the training, and later on how (if) we want to include other frameworks.



> > This is pretty much tied to PT, but I think it's okay to do it like this at this point. At least, until we get a better understanding of how we want to structure the training, and later on how (if) we want to include other frameworks.
> 
> Yeah, at this point we are pretty much coupled with torch for training. The extent of the coupling will be much less when we implement loss and optimizer on the device.

Fair point - however I discussed this with Stan and we are ok to target torch for now and make a plan for other frameworks at a later point.

vladimirjovanovicTT:
> > > > This is pretty much tied to PT, but I think it's okay to do it like this at this point. At least, until we get a better understanding of how we want to structure the training, and later on how (if) we want to include other frameworks.
> > > 
> > > 
> > > Yeah, at this point we are pretty much coupled with torch for training. The extent of the coupling will be much less when we implement loss and optimizer on the device.
> > 
> > 
> > Do we plan to keep an option for running loss and optimizer on the CPU as well in the long run? If yes, we'll still have framework counseling. However, this is the problem that we don't need to solve in next period for sure :))
> 
> Probably, but in my mind, it is an "edge case" use case, so I think we can be more relaxed (if that is the case) ... 😄
> 
> Anyhow, if we want more robust support, we will probably need to implement it on a framework-by-framework basis...

I would say that loss-on-host is an edge case, but optimizer-on-host has its uses, e.g. after all-reducing gradients during data parallel. 
Even if not on host, optimizer can be on a different/specific node in a device cluster when we start scaling out training - something to keep in mind.

forge/forge/compiled_graph_state.py
```diff
@@ -166,35 +170,54 @@ class ProgramId(IntEnum):
...
     ):
         self.fwd_compiled_graph_state = fwd_compiled_graph_state
         self.bwd_compiled_graph_state = bwd_compiled_graph_state
         self.compiled_binary = compiled_binary
         self.inputs = []
         self.framework_module = framework_module
-        self.loss_module = loss_module
-        self.optimizer = optimizer
         self.intermediates = []
+        self.loss_grad = [None for _ in range(len(fwd_compiled_graph_state.ordered_external_output_names))]
```
vladimirjovanovicTT, line 209: very nit: maybe self.loss_grad = [None] * len(fwd...), or [None for _ in fwd....]?

pilkicTT: done

vladimirjovanovicTT (approved):
Looks great!

PR #521, opened 2024-10-22
Fixed reduce_sum not lowered properly into mlir

### LINKED ISSUE #420: reduce_sum not lowered properly into mlir
When running `test_mnist_training` in `forge/mlir/mnist/training/test_training.py`.
Gradient for the bias in last linear layer is calculated as shown in picture below.
Screenshot 2024-10-16 at 18 12 16
in reduce_sum is positional parametar -2 but in mlir when it is lowered there is none:
`%4 = "ttnn.sum"(%2, %3) <{keep_dim = true}> : (tensor<100x100xbf16, #tt.layout<(d0, d1) -> (d0, d......`
so reduce_sum reduces all dimensions instead of only one

ndrakulicTT:
reduce_sum in backward did not set the dim_arg when emited in backward.

Closes #420

### DIFF
--- a/forge/forge/op/eval/forge/eltwise_binary.py
+++ b/forge/forge/op/eval/forge/eltwise_binary.py
@@ -225,14 +225,19 @@ def backward(op_type, attr, ac, operand, inputs, output, grad):
             for i in range(len(shapes[operand])):
                 if shapes[operand][i] < grad_shape[i]:
                     # Negative indexing for reduce axis
-                    grad = ac.op("reduce_sum", (grad,), (i - grad_shape_len,), {"keep_dim": True})
+                    grad = ac.op(
+                        "reduce_sum",
+                        (grad,),
+                        (i - grad_shape_len,),
+                        {"keep_dim": True, "dim_arg": [i - grad_shape_len]},
+                    )
         return ac.op(Nop.create(), (grad,))  # pass gradient through
 
     elif op_type == "subtract":
         if inputs[operand].shape != grad.shape:
             for i in range(len(shapes[operand])):
                 if shapes[operand][i] < grad.shape[i]:
-                    grad = ac.op("reduce_sum", (grad,), (i,), {"keep_dim": True})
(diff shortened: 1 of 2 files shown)


forge/forge/op/eval/forge/eltwise_binary.py
```diff
@@ -225,14 +225,19 @@ def backward(op_type, attr, ac, operand, inputs, output, grad):
...
+                        {"keep_dim": True, "dim_arg": [i - grad_shape_len]},
+                    )
         return ac.op(Nop.create(), (grad,))  # pass gradient through
 
     elif op_type == "subtract":
         if inputs[operand].shape != grad.shape:
             for i in range(len(shapes[operand])):
                 if shapes[operand][i] < grad.shape[i]:
-                    grad = ac.op("reduce_sum", (grad,), (i,), {"keep_dim": True})
+                    grad = ac.op("reduce_sum", (grad,), (i,), {"keep_dim": True, "dim_arg": [i]})
```
vladimirjovanovicTT, line 240: Not really related to this PR - would it make sense to refactor code for "add", "subtract", "multiply"? The 'for' loops (when shapes differ) look the same for each op. @nvukobratTT @pilkicTT

pilkicTT: Personally, I don't see much value in it, but I would be fine with it if you wanted to... 😄

pilkicTT (approved):

vladimirjovanovicTT (approved):

forge/forge/op/eval/forge/eltwise_binary.py
nvukobratTT: Yeah, for now, let's keep it separated for flexibility. We can think about a better structure once we get at least a few models running with valid graphs. 

However I agree, we can create a more unified functions to encapsulate same (or similar) bwd definitions.

nvukobratTT (approved):

PR #527, opened 2024-10-23
Test suite

vbrkicTT:
Grouping of test plans in test suites.
  - Support combining of tests across operators
  - Support central place for triggering single op tests
  - Advanced querying and filtering support

### DIFF
--- a/forge/test/operators/pytorch/test_all.py
+++ b/forge/test/operators/pytorch/test_all.py
@@ -1,0 +1,414 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+
+# Tests for testing all pytorch operators
+
+
+# Examples
+# pytest -svv forge/test/operators/pytorch/test_all.py::test_unique --collect-only
+# TEST_ID='no_device-ge-FROM_HOST-None-(1, 2, 3, 4)-Float16_b-HiFi4' pytest -svv forge/test/operators/pytorch/test_all.py::test_single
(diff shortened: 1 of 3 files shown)


forge/test/operators/pytorch/test_all.py
```diff
@@ -0,0 +1,377 @@
...
+            dev_data_formats = [getattr(forge.DataFormat, dev_data_format) for dev_data_format in dev_data_formats]
+
+        math_fidelities = os.getenv("MATH_FIDELITIES", None)
+        if math_fidelities:
+            math_fidelities = math_fidelities.split(",")
+            math_fidelities = [getattr(forge.MathFidelity, math_fidelity) for math_fidelity in math_fidelities]
+
+        # TODO KWARGS
+
+        filter_collection = TestCollection(
```
dgolubovicTT, line 89: nit: filtered_collection sounds more accurate here.

forge/test/operators/pytorch/test_all.py
```diff
@@ -0,0 +1,377 @@
...
+    @classmethod
+    def verify(cls, test_vector: TestVector, test_device):
+        if cls.DRY_RUN:
+            # pytest.skip("Dry run")
+            return
+        test_vector.verify(test_device)
+
+
+class TestParamsData:
+
```
dgolubovicTT, line 52: Can we add just one or two sentences to explain each class, because our framework is growing. For example: This is a parameter manager that collects input from environment variables to determine the tests that should run, which operators are involved, and which devices and data formats to use. It can create both test collections and filtering lambdas...

Something like that just to give an overview?
Apart from that LGTM!

dgolubovicTT (approved):

forge/test/operators/pytorch/test_all.py
vbrkicTT: renamed

forge/test/operators/pytorch/test_all.py
vbrkicTT: Included code comments

PR #534, opened 2024-10-24
[build] fixup clean_all script

pilkicTT:
The way `tt-mlir` builds the environment and the toolchain has changed, so now if you delete the environment directories, everything will need to be rebuilt (including LLVM, Flatbuffers, etc.)

Since cleaning everything (including toolchains) is rarely needed, update the script to leave `env/build` dirs intact.

Also, tidying up the build docs...

### DIFF
--- a/clean_all.sh
+++ b/clean_all.sh
@@ -2,9 +2,11 @@
 
 # SPDX-License-Identifier: Apache-2.0
 
+# Cleans all build artifacts, except for the environment and toolchains.
+# If you want to clean the environment (should be rarely needed), remove the env/build directories.
+# E.g. rm -rf env/build third_party/tt-mlir/env/build
+
 rm -rf build
-rm -rf env/build
 rm -rf third_party/tt-mlir/build
-rm -rf third_party/tt-mlir/env/build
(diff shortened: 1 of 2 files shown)


clean_all.sh
```diff
@@ -2,9 +2,11 @@
 
 # SPDX-License-Identifier: Apache-2.0
 
+# Cleans all build artifacts, except for the environment and toolchains.
+# If you want to clean the environment (should be rarely needed), remove the env/build directories.
+# E.g. rm -rf env/build third_party/tt-mlir/env/build
+
 rm -rf build
```
nvukobratTT, line 9: What do you think about having two scripts:
- clean_build.sh - Cleans FFE, MLIR and Metal build
- clean_all.sh - Runs clean_build.sh + Removes env

nvukobratTT (approved):
LGTM! One minor suggestion :))

clean_all.sh
pilkicTT: I am ok with renaming `clean_all.sh` -> `clean_build.sh`... But I would avoid adding a script for cleaning up build + env, since the need for it should be extremely rare.

PR #541, opened 2024-10-25
Skip bypassing the node when binary ops involved in between

ashokkumarkannan1:
This fix is to avoid bypassing the nodes when there is a shape match between two nodes but binary ops involved in between.

Before removing Transpose Op:
image (4)

After removing transpose Op:
image (5)

### DIFF
--- a/forge/csrc/passes/erase_inverse_ops.cpp
+++ b/forge/csrc/passes/erase_inverse_ops.cpp
@@ -80,6 +80,23 @@ static std::vector<graphlib::Node *> find_path_to_inverse_op(
     if (not found_inverse)
         path.clear();
 
+    // Skip bypassing the node when binary ops involved in between
+    // Ex: transpose -> multiply -> add -> transpose
+    bool is_valid_path = true;
+    int path_len = path.size();
+    if (path_len > 2)
+    {
+        for (int idx = 1; idx < path_len - 1; idx++)
+        {
+            auto &node = path.at(idx);
+            graphlib::OpNode *op_node = dynamic_cast<graphlib::OpNode *>(node);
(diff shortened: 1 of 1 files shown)


forge/csrc/passes/erase_inverse_ops.cpp
```diff
@@ -80,6 +80,23 @@ static std::vector<graphlib::Node *> find_path_to_inverse_op(
     if (not found_inverse)
         path.clear();
 
+    // Skip bypassing the node when binary ops involved in between
```
nvukobratTT, line 83: I'm not into introducing this workaround. 

If this takes longer, let's add a feature flag  (`forge/forge/config.py`) and disable erase inverse ops for problematic cases. 

However, let's push as a priority to fix this case properly.

forge/csrc/passes/erase_inverse_ops.cpp
```diff
@@ -80,6 +80,23 @@ static std::vector<graphlib::Node *> find_path_to_inverse_op(
...
+    // Ex: transpose -> multiply -> add -> transpose
+    bool is_valid_path = true;
+    int path_len = path.size();
+    if (path_len > 2)
+    {
+        for (int idx = 1; idx < path_len - 1; idx++)
+        {
+            auto &node = path.at(idx);
+            graphlib::OpNode *op_node = dynamic_cast<graphlib::OpNode *>(node);
```
pilkicTT, line 92: As a general note, please use `node->as<graphlib::OpNode>()` instead of manually casting to a specific node type. The `as()` method also asserts.

Additionally, if we want to change something with regards to casting in the future, we can do it at one place...

ashokkumarkannan1:
Fixed in https://github.com/tenstorrent/tt-forge-fe/pull/566

PR #543, opened 2024-10-25
Deprecated decompose of Transpose op

### LINKED ISSUE #497: Unsupported operation for lowering from TTForge to TTIR: pad_tile
## Description

-  Unsupported operation for lowering from TTForge to TTIR: pad_tile on Forge inference 

## Reproduce

```
git checkout kkannan/forge_model_issues
export BACKEND_ARCH_NAME=wormhole_b0 ARCH_NAME=wormhole_b0 PYBUDA_FORCE_EMULATE_HARVESTED=1 
pytest forge/test/model_demos/high_prio/nlp/pytorch/test_llama3.py::test_llama3_sequence_classification[no_device-meta-llama/Meta-Llama-3-8B] -svv
```

## Logs

- [llama3_seq_issue.log](https://github.com/user-attachments/files/17460284/llama3_seq_issue.log)

mstojkovicTT:
Removed deprecated decompose of Transpose op, transpose op is now supported by ttnn
fixes #497

### DIFF
--- a/forge/forge/op/eval/forge/transpose.py
+++ b/forge/forge/op/eval/forge/transpose.py
@@ -63,67 +63,6 @@ class TransposeTM(PyTM):
                 )
             )
 
-    def decompose_post_optimize(self, dc, inputs):
-        orig_shape = inputs[0].shape
-        if (
-            len(orig_shape) > 2
-            and self.dim0 == -3
-            and self.dim1 == -1
-            and ((len(orig_shape) == 4 and orig_shape[-4] == 1) or len(orig_shape) < 4)
-        ):
-            # XZ transpose
-            result = inputs[0]
-            use_sparse_mm = True
-
-            result = inputs[0]
-            result = dc.op("pad_tile", [result], (-2, orig_shape[-2]))
-            result = dc.op("pad_tile", [result], (-1, orig_shape[-1]))
-            result = dc.op(TransposeTM.create(-2, -1), [result])
-
-            if result.shape[-3] > 1:
-                result = dc.op("vstack", [result], (orig_shape[-3],))
-            i_spm = sparse_utils.create_sparse_interleave_picker_matrix(
-                result.shape[-2], orig_shape[-1], orig_shape[-3]
-            )
-            result = picker_matmul(use_sparse_mm, dc, i_spm, result)
-
-            if orig_shape[-1] > 1:
-                result = dc.op("vslice", [result], (orig_shape[-1],))
-            result = dc.op(TransposeTM.create(-2, -1), [result])
-
-            result = dc.op("narrow", [result], (-2, 0, orig_shape[-2], result.shape[-2]))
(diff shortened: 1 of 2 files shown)


forge/forge/op/eval/forge/transpose.py
```diff
@@ -63,67 +63,6 @@ def decompose(self, dc, inputs):
                 )
             )
 
-    def decompose_post_optimize(self, dc, inputs):
```
nvukobratTT, line 66: Can we expand `forge/test/mlir/test_ops.py::test_transpose` op test to include -3 dims as well? 

It's okay to add lower compile depth for -3 (Z dim) cases until we uplift metal fix into FFE as well :))

mstojkovicTT: Added tests for -3 dims, as well as compile depth set to FINISH_COMPILE until metal uplift. Should I make an issue that reminds us to undo those changes?

nvukobratTT: Yup! And feel free to assign it to yourself :))

forge/test/mlir/test_ops.py
```diff
@@ -87,18 +90,23 @@ def __init__(self, dims):
         def forward(self, a):
             return torch.transpose(a, *self.dims)
 
+    # currently the lowering to TTNN is not supported for this case
+    compiler_cfg = forge.config._get_global_compiler_config()
+    compiler_cfg.compile_depth = forge.CompileDepth.FINISH_COMPILE
```
nvukobratTT, line 95: This way, you'll limit all Transpose tests to run until the specified compile depth. It'll be good to early stop only those which are problematic, and run e2e the ones that are functional. 

To do so, you can detect dim args from test param, and lower compile depth only in those cases :))

forge/forge/op/eval/forge/transpose.py
dgolubovicTT: @mstojkovicTT did you try to run transpose on -3 without reducing compile depth? I think this will work too because we are doing some runtime decompositions... These are in: `third_party/tt-mlir/runtime/lib/ttnn/operations/data_movement/transpose.cpp`

forge/test/mlir/test_ops.py
mstojkovicTT: I was thinking about that, will do that :)

forge/forge/op/eval/forge/transpose.py
mstojkovicTT: @dgolubovicTT that would work if i return from test before comparing with golden_pcc (before giving inputs to compiled model)

mstojkovicTT: update: i removed compile depth from the code, and just left return before giving inputs to the compiled_model

dgolubovicTT (approved):

PR #557, opened 2024-10-29
Update tvm reference

dgolubovicTT:
Solves #563. Updating TVM reference to enable bfloat16 through TVM. Previously, it was silently casted to float32. Since float32 isn't natively supported on our hardware it is more natural for bfloat16 to be default data format. This prepares forge for transitioning on bfloat16 by default.


nvukobratTT (approved):

nvukobratTT:
Looks good! Let's just map issue to it as well :))

PR #562, opened 2024-10-30
Add option for dumping the unique op config as xlsx file and script to create sheet for model unique op configs

chandrasekaranpradeep:


forge/csrc/passes/extract_unique_op_configuration.cpp
```diff
@@ -207,12 +206,79 @@ void export_unique_op_configuration_to_csv_file(
...
+    std::string export_unique_op_config_file = export_unique_op_config_path + stage + ".xlsx";
+
+    py::module my_module = py::module::import("forge.utils");
+
+    py::str py_export_unique_op_config_file_path = export_unique_op_config_file.c_str();
+    py::str py_graph_name = graph_name.c_str();
+    py::str py_stage = stage.c_str();
+    py::dict py_unique_op_shapes_attrs;
+
+    for (const auto &[op_name, shapes_attrs] : unique_op_shapes_attrs)
```
dgolubovicTT, line 231: Can we try not to use auto when we don't need to. It makes things easier when we have actual type here

scripts/export_models_ops_correlation.py
```diff
@@ -70,6 +70,7 @@ def run_and_export_models_unique_ops_configuration(compilation_depth, pytest_dir
                     FORGE_COMPILE_DEPTH=compilation_depth,
                     FORGE_EXTRACT_UNIQUE_OP_CONFIG_AT=compilation_depth.upper(),
                     FORGE_EXPORT_UNIQUE_OP_CONFIG_TO_CSV="1",
+                    FORGE_EXPORT_UNIQUE_OP_CONFIG_TO_XLSX="1",
```
dgolubovicTT, line 73: Is it better to have only one env variable `FORGE_EXPORT_UNIQUE_OP_CONFIG_TYPE` and its value being "csv" or "xslx"? This is just to try having the least amount of env variables...

scripts/export_models_ops_correlation.py
```diff
@@ -117,9 +213,9 @@ def create_sheet_from_exported_models_unique_op_configuration(
                 dumped_op_configuration_file_path = os.path.join(model_path, dumped_op_configuration_file)
                 op_config_df = pd.read_csv(
                     dumped_op_configuration_file_path,
-                    sep="-",
+                    sep=os.getenv("FORGE_EXPORT_UNIQUE_OP_CONFIG_CSV_DELIMITER", "/"),
```
dgolubovicTT, line 216: Why we are defining delimiter through env variable? Is it expected to change, and if yes, why?

nvukobratTT: I can give more context here.

Using a configurable delimiter provides flexibility in this logic for different use cases. For example:
- Using ',' will outline issues for how shapes are interpreted
- Using '-' will outline issues around the usage of negative dims

Having unknowns around op specifics, it's more flexible to use a configurable delimiter. When this logic matures, we can remove configurable part, and just use delimiter we see fit for all cases :))

forge/csrc/passes/extract_unique_op_configuration.cpp
chandrasekaranpradeep: Sure I will update it

scripts/export_models_ops_correlation.py
chandrasekaranpradeep: The script used for cross correlating models variants and ops and sheet creation for models unique op configuration is currently using the csv file as input and xlsx are dumped in addition if someone want to see each  model unique configurations separately. If we are using one env variable, for generating sheet they need to run script with csv value for env variable and for better visualization as xlsx file, they need to run script with xlsx value for env variable and then dump the unique op configuration.
I will try to check if I am able to use xlsx file as input for generating sheet that contains cross correlation between model varaints and unique op configs. If yes, I will try to use single env variable.

forge/csrc/passes/extract_unique_op_configuration.cpp
chandrasekaranpradeep: Replaced auto with actual type

scripts/export_models_ops_correlation.py
chandrasekaranpradeep: Modified the code for dumping either as csv or xlsx file by having one env variable. 
New env Flag:
`FORGE_EXPORT_UNIQUE_OP_CONFIG_FILE_TYPE = {csv/xlsx}`
The script also has the option to process either as csv or xlsx file.

dgolubovicTT (approved):

nvukobratTT (approved):

Issue #560, opened 2024-10-30
Attribute 'keep_dim' error in 'ttir.mean' op in adaptive_avg_pool_2d


kamalrajkannan78:
## Description
 
- adaptive_avg_pool_2d faced `loc("avg_pool2d_0.dc.reduce_avg.2"("forward":4294967295:13)): error: 'ttir.mean' op requires attribute 'keep_dim'`

## Reproduce

```
git checkout kkannan/adaptive_avg_pool
git submodule update --recursive
pytest forge/test/mlir/test_ops.py -k "pool" -svv 
```

## Observed Behaviour
> loc("avg_pool2d_0.dc.reduce_avg.2"("forward":4294967295:13)): error: 'ttir.mean' op requires attribute 'keep_dim'

## Logs

- [pool_issue.log](https://github.com/user-attachments/files/17570614/oct30_pool_issue.log)
- [pool_fix.log](https://github.com/user-attachments/files/17570613/oct30_pool_fix.log)

kamalrajkannan78:
FYI @dgolubovicTT

Issue #559, opened 2024-10-30
Attribute 'shape' error in 'ttir.reshape' op in adaptive_avg_pool_2d 


kamalrajkannan78:
## Description
 

- adaptive_avg_pool_2d  faced `loc("avg_pool2d_16.dc.reshape.0"("forward":4294967295:78)): error: 'ttir.reshape' op requires attribute 'shape'`

## Reproduce

```
git checkout kkannan/adaptive_avg_pool
git submodule update --recursive
pytest forge/test/mlir/test_ops.py -k "issue" -svv 
```

## Observed Behaviour
> `loc("avg_pool2d_16.dc.reshape.0"("forward":4294967295:78)): error: 'ttir.reshape' op requires attribute 'shape'`

## Logs

- [reshape_issue.log](https://github.com/user-attachments/files/17570360/oct30_reshape_issue.log)

- [reshape_fix.log](https://github.com/user-attachments/files/17570359/oct30_reshape_fix.log)

kamalrajkannan78:
FYI @dgolubovicTT

PR #567, opened 2024-10-30
Fix decompose of AvgPool2D to account for named attributes and skip runtime reinterpreting shapes

dgolubovicTT:
Sloves #560, solves #559. 

This PR includes:
    - AvgPool2D fix decompose to account for named attributes that are needed for mlir lowering.
    - skip lower_reinterpret_shape in optimization passes to avoid changing expected shapes on input.
    - Test  test_avg_pool2d to test this.


forge/csrc/forge_passes.cpp
```diff
@@ -157,7 +157,8 @@ void run_optimization_graph_passes(graphlib::Graph *graph)
 
     passes::hoist_transforms_to_inputs(graph);
     passes::erase_consecutive_reshape(graph, true);
-    passes::lower_reinterpret_shape(graph);
+    // skip lower_reinterpret_shape in optimization passes to avoid changing expected shapes on inputs
```
nvukobratTT, line 160: Do you see a value in fully removing this code section? If we need a reference in future, we can always check PyBuda repo.

forge/forge/op/eval/forge/pooling.py
```diff
@@ -478,15 +478,20 @@ def decompose(type, attr, dc, inputs):
         # If global average
         if y == kH and x == kW and ((stride[0] == kH and stride[1] == kW) or all(pad == 0 for pad in padding)):
             if channel_last:
-                result = dc.op("reshape", [activations], (w, 1, y * x, cin))
-                result = dc.op("reduce_avg", [result], (-2,))
-                result = dc.op("reshape", [result], (w, 1, 1, cin))
+                # for reshape use dc.op_with_named_attrs
```
nvukobratTT, line 481: Do we need this comment? In the future, we should transfer to using named attributes for each op :))

nvukobratTT (approved):

forge/forge/op/eval/forge/pooling.py
dgolubovicTT: Yeah it is unnecessary...

forge/csrc/forge_passes.cpp
pilkicTT: +1 just remove it 😄

i would also remove that pass entirely.

dgolubovicTT: Hmmm, I wanted to leave some clue why we are skipping this for now. Looking back to PyBuda doesn't sound amazing to me. Do we have some third option?

nvukobratTT: > Do we have some third option?

To remove it and never look back? 😂

PR #568, opened 2024-10-30
Fix for passing unnecessary inputs to the backward pass of the 

ndrakulicTT:
Filtered inputs from the forward pass that are passed to backward pass by name.

For test: test_forge_vs_torch_gradients added option to freeze layer
if the layer is freezed gradient should be None.

Modified `forge/test/mlir/mnist/utils.py::get_param_grads` to skip parameters that dont have grad.

### DIFF
--- a/forge/forge/compiled_graph_state.py
+++ b/forge/forge/compiled_graph_state.py
@@ -281,11 +281,17 @@ class CompiledModel:
         for grad in self.loss_grad:
             assert grad is not None, "Gradients not provided for backward pass."
 
+        inputs = [
+            self.inputs[i]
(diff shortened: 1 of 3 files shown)


forge/forge/compiled_graph_state.py
```diff
@@ -281,11 +281,17 @@ def backward(self) -> List[torch.Tensor]:
         for grad in self.loss_grad:
             assert grad is not None, "Gradients not provided for backward pass."
 
+        inputs = [
```
nvukobratTT, line 284: Let's add a block comment explaining inputs filtering :))

nvukobratTT (approved):

pilkicTT (approved):

PR #603, opened 2024-11-01
Script to run generated standalone Forge tests + report summary

### LINKED ISSUE #602: [Model Analysis] Create script to run standalone tests and generate output report
We should have a script that will run all standalone tests and generate a report based on test results. Make sure to store logs for each test run, and generate end summary files that will have key errors from each test.

nvukobratTT:
- Purpose of this script is to run standalone Forge op tests, and generate summary with passing/failing states
- During scirpt execution, we're outputing logs with time required for each test run
- Have in mind that each pytest is run as separated thread. Why? Because MLIR doesn't have good execption handling, and often fails by crushing pytest thread. Therefore, we're using main thread as monitor to record those failures as well
- Each test log is sotred in as separated files. Final report is generated based on those logs.
- Main summary file contains information about key errors, and is configurable to have cleaner, or more verbose error notes

Fix #602

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -1,0 +1,157 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-License-Identifier: Apache-2.0
+import subprocess
+import os
+import time
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


dgolubovicTT (approved):
LGTM!

Issue #506, opened 2024-10-21
 AssertionError: Setting a tensor value of incorrect shape: (1, 256, 16, 2, 16) vs torch.Size([1, 256, 16, 16, 2])


kamalrajkannan78:
## Description

-   AssertionError: Setting a tensor value of incorrect shape: (1, 256, 16, 2, 16) vs torch.Size([1, 256, 16, 16, 2])

## Reproduce

```
git checkout kkannan/forge_model_issues
git submodule update --recursive
export BACKEND_ARCH_NAME=wormhole_b0 ARCH_NAME=wormhole_b0 PYBUDA_FORCE_EMULATE_HARVESTED=1 
pytest forge/test/model_demos/high_prio/nlp/pytorch/test_codegen.py -svv
```

## Observed Behaviour
```
self = Forge Empty Tensor: (1, 256, 16, 2, 16)
value = tensor([[[[[-3.4713e+00,  6.3812e-01],
           [ 3.0785e+00,  1.1727e+00],
           [ 2.6389e+00,  1.5917e+00],
 ....3141e+00],
           [ 2.3826e+00,  4.5052e-01],
           [-8.0871e-01,  1.4025e+00]]]]], grad_fn=<StackBackward0>)

    def set_value(self, value: torch.Tensor):
        assert (
>           self.tensor_shape.get_pytorch_shape() == value.shape
        ), f"Setting a tensor value of incorrect shape: {self.tensor_shape.get_pytorch_shape()} vs {value.shape}"
E       AssertionError: Setting a tensor value of incorrect shape: (1, 256, 16, 2, 16) vs torch.Size([1, 256, 16, 16, 2])
```

## Logs
- [oct21_codegen.log](https://github.com/user-attachments/files/17464517/oct21_codegen.log)

PR #606, opened 2024-11-04
Fix stack op sanity failure

meenakshiramanathan1:
Fix for #506 
1. Axis has been changed in eltwise_nary.py to address #506  for codegen model
2. Stack op test case was failing with error `Trying to access element outside of dimensions: 4` for test cases with dim =-2 and dim =3 in optmised graph pass. It was failing because new_dim was not calculated correctly in can_commute_reshape_through_dim in commute_utils.cpp, hence it was going out of bounds, so a condition was added to skip the out of bounds test case.
Full log: [stack_sanity.log](https://github.com/user-attachments/files/17616011/stack_sanity.log)

### DIFF
--- a/forge/csrc/passes/commute_utils.cpp
+++ b/forge/csrc/passes/commute_utils.cpp
@@ -45,9 +45,12 @@ std::tuple<bool, int> can_commute_reshape_through_dim(
     {
         if (input_shape_vec[i] == output_shape_vec[dim])
         {
+            log_trace("output_shape_vec.size() = {}", output_shape_vec.size());
+            log_trace("i value = {}", i);
             // check whether volume above and below matching dim is the same
             if ((volume_above(input_shape_vec, i) == volume_above(output_shape_vec, dim)) and
-                (volume_below(input_shape_vec, i) == volume_below(output_shape_vec, dim)))
+                (volume_below(input_shape_vec, i) == volume_below(output_shape_vec, dim)) and
+                (i < output_shape_vec.size()))
             {
                 can_commute = true;
                 new_dim = i;
--- a/forge/forge/op/eval/forge/eltwise_nary.py
+++ b/forge/forge/op/eval/forge/eltwise_nary.py
@@ -142,9 +142,13 @@ def shape(type, attr, ops) -> Tuple[Tuple, List]:
 
     elif type == "stack":
         axis = attr[0]
-
         output_shape = list(ops[0])
-        output_shape.insert(axis, len(ops))
+        if axis == -1:
+            output_shape.append(len(ops))
+        elif axis < 0:
+            output_shape.insert(axis + 1, len(ops))
+        else:
+            output_shape.insert(axis, len(ops))
         return output_shape, []
 
     elif type == "interleave":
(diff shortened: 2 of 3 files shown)


forge/csrc/passes/commute_utils.cpp
```diff
@@ -45,9 +45,12 @@ std::tuple<bool, int> can_commute_reshape_through_dim(
     {
         if (input_shape_vec[i] == output_shape_vec[dim])
         {
+            log_info("output_shape_vec.size() = {}", output_shape_vec.size());
```
nvukobratTT, line 48: As we're adding these logs, it'll be good to add a few more for easier debugging of this pass. Also, these are better fit for trace logs

forge/forge/op/eval/forge/eltwise_nary.py
```diff
@@ -246,11 +246,12 @@ def decompose(type, attr, dc, inputs):
         new_inputs = []
         for inp in inputs:
             inp_shape = inp.shape.as_list()
-            inp_shape.insert(axis, 1)
-            new_inp = dc.op("reshape", [inp], (*inp_shape,))
+            insertion_index = axis if axis >= 0 else len(inp_shape) + axis + 1
```
nvukobratTT, line 249: Why do we need to convert to positive indexing?

forge/test/mlir/test_ops.py
```diff
@@ -49,6 +49,51 @@ def forward(self, x):
...
+        ([(1, 256, 24, 24), (1, 256, 24, 24)], 4),
+        ([(1, 256, 24, 24), (1, 256, 24, 24)], 2),
+        ([(5, 64, 128, 128), (5, 64, 128, 128)], 1),
+        ([(1, 30, 30, 16), (1, 30, 30, 16)], 0),
+    ],
+)
+def test_stack(params):
+    _, _, dim = params[0][0], params[0][1], params[1]
+    if dim == -3 or dim == 1:
+        pytest.xfail("Tensor rank is not 4")
```
nvukobratTT, line 69: Is this failing on the FFE side?

forge/forge/op/eval/forge/eltwise_nary.py
meenakshiramanathan1: With negative axis, insertion is not applied correctly
When axis = -4, output shape is (2, 1, 256, 24, 24) while expected shape from eval function is (1, 2, 256, 24, 24), this is being rectified when axis is being converted to negative

forge/test/mlir/test_ops.py
meenakshiramanathan1: It is failing in run_mlir_compiler stage
log: [stack_mlir.log](https://github.com/user-attachments/files/17616626/stack_mlir.log)

forge/csrc/passes/commute_utils.cpp
```diff
@@ -45,9 +45,12 @@ std::tuple<bool, int> can_commute_reshape_through_dim(
     {
         if (input_shape_vec[i] == output_shape_vec[dim])
         {
+            log_trace("output_shape_vec.size() = {}", output_shape_vec.size());
+            log_trace("i value = {}", i);
             // check whether volume above and below matching dim is the same
             if ((volume_above(input_shape_vec, i) == volume_above(output_shape_vec, dim)) and
-                (volume_below(input_shape_vec, i) == volume_below(output_shape_vec, dim)))
+                (volume_below(input_shape_vec, i) == volume_below(output_shape_vec, dim)) and
+                (i < output_shape_vec.size()))
```
dgolubovicTT, line 53: What does this check prevent from happening?

dgolubovicTT:
@meenakshiramanathan1 It is not clear to me what is the problem we are solving in second point of the PR, and how we are solving it. Is this a workaround to skip out of bounds test case, or is it general fix? Can you please reference/open the issue and explain what is happening?

meenakshiramanathan1:
> @meenakshiramanathan1 It is not clear to me what is the problem we are solving in second point of the PR, and how we are solving it. Is this a workaround to skip out of bounds test case, or is it general fix? Can you please reference/open the issue and explain what is happening?

Please refer to #615

nvukobratTT:
Let's discuss this one on today's sync. I still have few unknowns that I don't fully grasp at the moment

meenakshiramanathan1:
Opened a new PR #669  related to this

Issue #336, opened 2024-09-25
Support for lowering of ttir.cast op


nvukobratTT:
More details for HLO sample can be found here:
- https://github.com/tenstorrent/tt-mlir/pull/811

nvukobratTT:
Done few days ago

Issue #128, opened 2024-08-16
[Llama 3B] Support for Decode (prefill on host)


nvukobratTT:

nvukobratTT:
PoC supported with KV cache. Waiting on TTNN issues to be fully resolved for full model e2e run

PR #607, opened 2024-11-04
Add support for Llama 3B decode with and without cache tests and cast op

### LINKED ISSUE #422: Problems with KV cache padding to max_sequence length
Since our compiler deals with operations that have static shapes (inputs and outputs) in order to enable KV cache we have to pad K and V tensors to some max_seq_len. It is eather that or compile in every iteration of decode loop.

With this premise, I lay out three issues I've encountered while debugging data mismatches in Llama 3B token generation. 
In order to KV cache to work with padding we have to understand and solve these in our compiler:

image

Here is the link for this visualization together with issues: https://excalidraw.com/#json=jBvSa7lieemcWAwXBLyAc,qU8bzz4x8-8kMoELaCbYqw

Issues:

1. Rotarry embedings for each token depend on sequence length which is changed by padding. Possible action: track actual sequence length and calculate rotary embeddings based on that...
2. Llama attention updates K matrix with k_previous_token by APPENDING it to the end. After that, it applies rotary embeddings to each row in K by rotating the vector in position "i" for Theta*i where theta is some angle (depends on seq len - issue1), and "i" is index that is for k_previous_token equal to:max_seq_len but should be 6. Possible action: instead of appending new k, our compiler should remap append to inserting new k to first free index...(tracking actual sequence length)
3. Without padding, there is no need for providing attention mask, because new token is influenced by all previous. Here, we have to have attention mask because we don't want zeros from K matrix to influence softmax. Possible action: we need to create attention mask that takes into account where padding starts.

 I wish to avoid hacking these to enable Llama, but solve KV cache more generally.

chandrasekaranpradeep:
1. Add support for Llama 3B decode with and without cache:
      Fixes #422, #128  
      - No Cache (i.e test_llama_prefill_on_cpu_decode_on_tt_no_cache):
          - **Prefill stage:**
              - It will be executed on CPU. The input_ids and attention_mask are padded to max sequence length and then passed to the llama 3b model in prefill stage and return logits.
          - **Decode stage:**
              - It will be executed on TT devices. In this decode stage, the input_ids first padding token id(i.e 0) is replaced with predicted next token from the prefill/previous decode stage and attention_mask first padding mask(i.e 0) is replaced with non padding token mask (i.e 1) and the passed to the llama 3b model in decode stage for predicting the next token untill the a specified maximum number of new tokens is reached or an end-of-sequence token is encountered
      -  Cache (i.e test_llama_prefill_on_cpu_decode_on_tt_cache):
         - **Prefill stage:**
             - It will be executed on CPU. The input_ids and attention_mask are passed to the llama 3b model in prefill stage and return logits and past key values. Before passing to the decode stage, the past key values from prefill are zero padded on key_values_seq_length dim (i.e -2).
          - **Decode stage:**
              - It will be executed on TT devices. In this decode stage, the predicted next token from the prefill/previous decode stage, padded past key values, attention_mask(i.e will of shape (batch_size, key_value_seq_length +1), attention mask contains 0 or 1 as values. 0 means masked, 1 means not masked. The mask value 1 is applied upto the non padded past key value index and last index in past key values. The mask value 0 is applied to the padding past key values index) and postion_ids (i.e will be of shape (batch_size, 1), it contains the first non padding past key values index) are passed to the llama 3b model in decode stage for predicting the next token untill the a specified maximum number of new tokens is reached or an end-of-sequence token is encountered. 
 2. Add support for the cast op(i.e TTIR TypeCastOp) in Forge: - #336 
     - Currently tested the cast op conversion between float32 and int32.
 3. Remove the bypass_embedding_input_nops pass in run_pre_lowering_passes compilation stage:
     - Earlier the tvm relay cast op is mapped to Forge Identity Op(i.e Nop) in tvm_to_python.py file, though we are currently supporting cast op in Forge, there will be no introduction of nop op as input_id operand for the embedding op, so removing the pass.
 4. Enabling the verify_forge_codegen_vs_framework flag to compare Framework output and generated forge module output, because in the generated forge module some of the inputs passed to the forward function are mapped as unused input which leads to pcc drop once we compile and run the models.

### DIFF
--- a/forge/forge/op/eval/forge/cast.py
+++ b/forge/forge/op/eval/forge/cast.py
@@ -1,0 +1,29 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Union
+import torch
+
+from ..interface import PyEltwiseUnaryOp
+from ....tensor import forge_dataformat_to_pytorch_dtype, pytorch_dtype_to_forge_dataformat
+from forge._C import DataFormat
+
+
+class Cast(PyEltwiseUnaryOp):
+    @classmethod
+    def create(cls, dtype: Union[torch.dtype, DataFormat]):
+        self = cls("cast")
+        self.dtype = pytorch_dtype_to_forge_dataformat(dtype).to_json()
+        return self
+
+    def eval(self, tensors):
+        assert len(tensors) == 1, "Cast should have one input"
+        dtype = forge_dataformat_to_pytorch_dtype(DataFormat.from_json(self.dtype))
+        ret = tensors[0].to(dtype)
+        return ret
+
+    def shape(self, tensor_shapes):
+        assert len(tensor_shapes) == 1, "Cast should have one input"
+        shape = tensor_shapes[0]
+        return shape, []
--- a/forge/test/mlir/llama/tests/test_llama_decode.py
+++ b/forge/test/mlir/llama/tests/test_llama_decode.py
@@ -1,0 +1,324 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+import pytest
+
+import torch
+from transformers import LlamaConfig, LlamaForCausalLM, LlamaTokenizer
+from transformers.modeling_attn_mask_utils import _prepare_4d_causal_attention_mask
+
+import forge
+from forge.op.eval.common import compare_with_golden
+from test.mlir.llama.utils.utils import load_model
+
+
+class LlamaModelWrapper(torch.nn.Module):
+    """
+    In LlamaModelWrapper class, forward function takes input_ids (i.e list of input token or
+    or last predicted single token), attention_mask and return logits
+    Args:
+        1) use_cache = False
+            input_id (`torch.Tensor`) - shape of (batch_size, seq_length)
+            attention_mask (`torch.Tensor`) -  shape of (batch_size, seq_length)
+        2) use_cache = True
+            input_id (`torch.Tensor`) - shape of (batch_size, 1)
+            attention_mask (`torch.Tensor`) -  shape of (batch_size, key_value_seq_len + 1)
+            position_ids (`torch.Tensor`) -  shape of (batch_size, 1)
+            past_key_values (`List[List[torch.Tensor, torch.Tensor]]`)
+                        key/values shape: (batch_size, num_of_key_values_heads, key_value_seq_len, head_dim)
+
+    Returns:
+        1) use_cache = False
+            logits (`torch.Tensor`) - shape (batch_size, seq_length, vocab_size)
+        2) use_cache = True
+            logits (`torch.Tensor`) - shape (batch_size, 1, vocab_size)
+            past_key_values (`List[List[torch.Tensor, torch.Tensor]]`)
+                        key/values shape: (batch_size, num_of_key_values_heads, key_value_seq_len + 1, head_dim)
+    """
+
+    def __init__(self, model):
+        super().__init__()
+        self.model = model
+        self.embed_tokens = model.model.embed_tokens
+
+    def forward(self, input_ids, attention_mask, position_ids=None, past_key_values=None):
+
+        inputs_embeds = self.embed_tokens(input_ids)
+
+        # The causal_mask calculated inside _update_causal_mask method in LlamaModel module
+        # is not properly traced by torch.jit.trace while converting pytorch to relay ops in TVM PyTorch Frontend Conversion
+        # which leads to pcc drops, because the causal mask helps to avoid attention calculation on padding token in LlamaAttention module
+        # To overcome this jit trace issue, causal mask is calculated by using _prepare_4d_causal_attention_mask function.
+        past_key_values_length = past_key_values[0][0].shape[-2] if past_key_values is not None else 0
+        causal_attention_mask = _prepare_4d_causal_attention_mask(
+            attention_mask, input_ids.shape, inputs_embeds, past_key_values_length
+        )
+        model_outputs = self.model(
+            attention_mask=causal_attention_mask,
+            position_ids=position_ids,
+            past_key_values=past_key_values,
+            inputs_embeds=inputs_embeds,
+        )
+
+        if model_outputs.past_key_values is not None:
+            return model_outputs.logits, model_outputs.past_key_values
+
+        return model_outputs.logits
+
+
+def calculate_attention_mask_and_postion_ids(
+    padded_past_key_values_seq_length, non_padding_past_key_values_seq_length, input_seq_length
+):
+
+    # Calculate attention mask
+    attention_mask = torch.zeros(padded_past_key_values_seq_length + input_seq_length)
+    attention_mask[:non_padding_past_key_values_seq_length] = 1
+    attention_mask[-1] = 1
+    attention_mask = attention_mask.unsqueeze(0)
+
+    # Calculate position ids
+    position_ids = torch.arange(
+        non_padding_past_key_values_seq_length,
+        input_seq_length + non_padding_past_key_values_seq_length,
+        dtype=torch.long,
+    )
+    position_ids = position_ids.unsqueeze(0)
+
+    return attention_mask, position_ids
+
+
+@pytest.mark.parametrize("run_on_tt_device", [False, True])
+def test_llama_prefill_on_cpu_decode_on_tt_no_cache(run_on_tt_device):
+
+    if run_on_tt_device:
+        compiler_cfg = forge.config._get_global_compiler_config()
+        compiler_cfg.compile_depth = forge.CompileDepth.SPLIT_GRAPH
+
+    # Load Llama 3B model and tokenizer
+    model_path = "openlm-research/open_llama_3b"
+    model, tokenizer = load_model(model_path, return_dict=True, use_cache=False, use_fast=False)
+    framework_model = LlamaModelWrapper(model)
+    framework_model.eval()
+    tokenizer.pad_token_id = model.config.pad_token_id
+
+    # Prepare input sentence
+    max_sequence_length = 58
+    prompt = "Q: What is the largest animal?\nA:"
+    inputs = tokenizer(
+        prompt,
+        return_tensors="pt",
+        max_length=max_sequence_length,
+        pad_to_max_length=True,
+        truncation=True,
+    )
+    input_ids = inputs.input_ids
+    attention_mask = inputs.attention_mask
+
+    non_padding_seq_len = int(torch.sum(attention_mask))
+    padding_seq_len = attention_mask.shape[1] - non_padding_seq_len
+
+    # Run Prefill on CPU with no cache to get the initial logits
+    logits = framework_model(input_ids=input_ids, attention_mask=attention_mask)
+
+    # Take the last non padding token index in logits for the next token
+    next_token_logits = logits[:, non_padding_seq_len - 1, :]
+    next_token = torch.argmax(next_token_logits, dim=-1)
+
+    # Update the input_ids with predicted token from prefill in the first padding token index
+    input_ids[:, non_padding_seq_len] = next_token
+    attention_mask[:, non_padding_seq_len] = 1
+
+    non_padding_seq_len = int(torch.sum(attention_mask))
+    padding_seq_len = attention_mask.shape[1] - non_padding_seq_len
+
+    if run_on_tt_device:
+
+        # Compile the model on TT
+        compiled_model = forge.compile(framework_model, sample_inputs=[input_ids, attention_mask])
+        pytest.xfail("Found Unsupported operations while lowering from TTForge to TTIR in forward graph.")
+
+    # Run decode stage on TT device and generate tokens by appending predicted token into sequence of input tokens
+    # untill the a specified maximum number of new tokens is reached or an end-of-sequence token is encountered.
+    max_new_tokens = padding_seq_len
+    for idx in range(max_new_tokens):
+
+        # CPU Inference
+        framework_output = framework_model(input_ids=input_ids, attention_mask=attention_mask)
+
+        if run_on_tt_device:
+
+            model_inputs = [input_ids, attention_mask]
+
+            # Run on TT device
+            tt_output = compiled_model(*model_inputs)
+            tt_output = [tt_out.to("cpu") for tt_out in tt_output]
+
+            framework_output = [framework_output] if isinstance(framework_output, torch.Tensor) else framework_output
+
+            # Validate TT result with Framework
+            assert all(
+                [
+                    compare_with_golden(golden=fw_out, calculated=tt_out, pcc=0.99)
+                    for fw_out, tt_out in zip(framework_output, tt_output)
+                ]
+            )
+
+            logits = tt_output[0]
+
+        else:
+
+            logits = framework_output
+
+        next_token_index = non_padding_seq_len + idx
+        next_token_logits = logits[:, next_token_index - 1, :]
+        next_token = torch.argmax(next_token_logits, dim=-1)
+
+        if next_token == tokenizer.eos_token_id:
+            break
+
+        input_ids[:, next_token_index] = next_token
+        attention_mask[:, next_token_index] = 1
+
+    # Generated text
+    generated_text = tokenizer.decode(input_ids[0], skip_special_tokens=True)
+    print("generated_text=", generated_text)
+
+
+@pytest.mark.parametrize("run_on_tt_device", [False, True])
+def test_llama_prefill_on_cpu_decode_on_tt_cache(run_on_tt_device):
+
+    if run_on_tt_device:
+        compiler_cfg = forge.config._get_global_compiler_config()
+        compiler_cfg.compile_depth = forge.CompileDepth.SPLIT_GRAPH
+
+    # Load Llama 3B model and tokenizer
+    model_path = "openlm-research/open_llama_3b"
+    model, tokenizer = load_model(model_path, return_dict=True, use_cache=True, use_fast=False)
+    framework_model = LlamaModelWrapper(model)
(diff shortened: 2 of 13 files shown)


forge/test/mlir/test_ops.py
```diff
@@ -13,6 +13,48 @@
 from forge.op.eval.common import compare_with_golden_pcc, compare_with_golden
 from forge.tensor import to_forge_tensors, to_pt_tensors
 
+
+@pytest.mark.parametrize("operand_and_cast_dtype", [(torch.float32, torch.int32), (torch.int32, torch.float32)])
```
nvukobratTT, line 17: @chandrasekaranpradeep can you create issues to test out cast op with other dtypes and assignee it to the Bringup milestone? Just as reminder to check other cases as well (no need to push it with this change)

forge/forge/op/eval/forge/cast.py
```diff
@@ -0,0 +1,29 @@
...
+
+from typing import Union
+import torch
+
+from ..interface import PyEltwiseUnaryOp
+from ....tensor import forge_dataformat_to_pytorch_dtype, pytorch_dtype_to_forge_dataformat
+from forge._C import DataFormat
+
+
+class Cast(PyEltwiseUnaryOp):
```
nvukobratTT, line 13: cc @dgolubovicTT This one will be a good addition to testing data formats. 

After the push of ResNet 50, let's checkout how to test out and properly model data formats through TVM and FFE. If we don't do it sooner rather than latter, it'll probably bite us for other models as well

forge/test/mlir/llama/tests/test_llama_decode.py
```diff
@@ -0,0 +1,335 @@
...
+            inputs_embeds=inputs_embeds,
+        )
+
+        if model_outputs.past_key_values is not None:
+            return model_outputs.logits, model_outputs.past_key_values
+
+        return model_outputs.logits
+
+
+def load_model_and_tokenizer(model_path="openlm-research/open_llama_3b", **kwargs):
```
nvukobratTT, line 68: Should this one be a common function "part of utilis.py" for Llama models?

nvukobratTT (approved):
Change looks good! Few minor commits, but not a blocker for approval :))

forge/test/mlir/test_ops.py
chandrasekaranpradeep: Sure Nikola.
I will create a issue for cast op for checking other dtypes with Bringup milestone.
FYI, There is issue for the cast support in forge (i.e https://github.com/tenstorrent/tt-forge-fe/issues/336) I will link the issue in this PR for current implementation I have done for cast op

forge/test/mlir/llama/tests/test_llama_decode.py
chandrasekaranpradeep: I have created the function inside the test_llama_decode.py purposely because the openlm-research/open_llama_3b variant will not generate correct token when we use autotokenizer, @mstojkovicTT mentioned fix https://github.com/tenstorrent/tt-forge-fe/issues/422#issuecomment-2441188184 here, but I can't find out the fix.  I don't want to distrub the existing code. so why I created function inside the test file.

nvukobratTT: Can we merge the two to capture these details in one place? 

More precisely, having conditions per variant and applying different tokenizers is okay. Having those in a single place will make it much more easier to review and understand why there are these small differences

forge/test/mlir/test_ops.py
chandrasekaranpradeep: Created issue for cast op: https://github.com/tenstorrent/tt-forge-fe/issues/610

forge/test/mlir/llama/tests/test_llama_decode.py
chandrasekaranpradeep: Will update it

chandrasekaranpradeep: Updated.

forge/test/mlir/llama/tests/test_llama_decode.py
```diff
@@ -0,0 +1,321 @@
...
+            [
+                v,
+                torch.zeros(v.shape[-4], v.shape[-3], max_new_tokens, v.shape[-1]).to(k.dtype),
+            ],
+            dim=-2,
+        )
+
+    if run_on_tt_device:
+        # Calculate attention mask and postion_ids
+        padded_past_key_values_seq_length = model_inputs[1][0][0].shape[2]
```
dgolubovicTT, line 240: It's a nit, but we should use negative indexing whenever we can.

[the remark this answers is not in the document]
chandrasekaranpradeep: Will update it

forge/test/mlir/llama/tests/test_llama_decode.py
```diff
@@ -0,0 +1,321 @@
...
+):
+
+    # Calculate attention mask
+    attention_mask = torch.zeros(padded_past_key_values_seq_length + input_seq_length)
+    attention_mask[:non_padding_past_key_values_seq_length] = 1
+    attention_mask[-1] = 1
+    attention_mask = attention_mask.unsqueeze(0)
+
+    # Calculate position ids
+    position_ids = torch.arange(
```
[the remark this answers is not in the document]
chandrasekaranpradeep: @dgolubovicTT Yes We can calculate position_ids from attention mask but we need to call this prepare_inputs_for_generation method inside the LlamaModelWrapper. After including position_id calculation(i.,e by calling prepare_inputs_for_generation) inside the  LlamaModelWrapper, when I compile the model,  hitting the error in TVM PyTorch frontend conversion. 

```
            # Compile the model on TT
>           compiled_model = forge.compile(framework_model, sample_inputs=[input_ids, attention_mask])

forge/test/mlir/llama/tests/test_llama_decode.py:138: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
forge/forge/compile.py:245: in compile_main
    return forge_compile_from_context(compile_context)
forge/forge/compile.py:287: in forge_compile_from_context
    next_stage = stage_to_func[current_stage](context)
forge/forge/compile.py:626: in generate_initial_graph
    module, module_inputs = convert_to_forge_module(
forge/forge/compile.py:980: in convert_to_forge_module
    forge_module, dev_types, module_inputs = generate_forge_module(
forge/forge/tvm_to_python.py:1994: in generate_forge_module
    module_writers, flattened_inputs = compile_tvm_to_python(
forge/forge/tvm_to_python.py:2079: in compile_tvm_to_python
    json_graphs, flattened_pytorch_inputs, weights = load_tvm_graph(
third_party/tvm/python/tvm/contrib/forge_compile.py:106: in load_tvm_graph
    json_graphs, flattened_inputs = compile_tvm_graph(inputs, module, compiler_cfg, graph_name=graph_name, input_names=input_names, path=path, verify_cfg=verify_cfg, framework=framework)
third_party/tvm/python/tvm/contrib/forge_compile.py:146: in compile_tvm_graph
    json_graphs, inputs = compile_pytorch_for_forge(module, *inputs, graph_name=graph_name, compiler_cfg=compiler_cfg, verify_cfg=verify_cfg, input_names=input_names)
third_party/tvm/python/tvm/contrib/forge_compile.py:372: in compile_pytorch_for_forge
    mod, params = tvm.relay.frontend.from_pytorch(traced_model, input_structure, do_convert_params=convert_params)
third_party/tvm/python/tvm/relay/frontend/pytorch.py:6105: in from_pytorch
    outputs = converter.convert_operators(operator_nodes, outputs, ret_name, input_remap)
third_party/tvm/python/tvm/relay/frontend/pytorch.py:5196: in convert_operators
    self.record_output_type(relay_out)
third_party/tvm/python/tvm/relay/frontend/pytorch.py:258: in record_output_type
    self.infer_type_with_prelude(output)
third_party/tvm/python/tvm/relay/frontend/pytorch.py:194: in infer_type_with_prelude
    body = self.infer_type(val, self.prelude.mod)
third_party/tvm/python/tvm/relay/frontend/pytorch.py:187: in infer_type
    new_mod = transform.InferType()(new_mod)
third_party/tvm/python/tvm/ir/transform.py:163: in __call__
    return _ffi_transform_api.RunPass(self, mod)
third_party/tvm/python/tvm/_ffi/_ctypes/packed_func.py:242: in __call__
    raise_last_ffi_error()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def raise_last_ffi_error():
        """Raise the previous error from FFI
    
        This should be used instead of `raise get_last_ffi_error()`, as it
        handle propagation of errors across an FFI boundary.  For example,
        if Python passes a callback to a C++ function, and the callback
        raises an exception, the re-thrown exception should contain the
        full stack trace, not just the stack frames that are above the
        outermost FFI call.
        """
    
        _LIB.TVMGetLastPythonError.restype = ctypes.c_void_p
        _LIB.TVMGetLastBacktrace.restype = ctypes.c_char_p
        py_err = _LIB.TVMGetLastPythonError()
        if py_err is None:
            c_err_msg = py_str(_LIB.TVMGetLastError())
            py_err_msg, err_type = c2pyerror(c_err_msg)
            if err_type is not None and err_type.startswith("tvm.error."):
                err_type = err_type[10:]
            py_err = ERROR_TYPE.get(err_type, TVMError)(py_err_msg)
    
        else:
            # TVMGetLastPythonError returns a PyObject*, with NULL when
            # there is no such value.  If we annotated the restype as
            # ctypes.py_object, we would need to return Py_None from the
            # C++ implementation.  This would require introducing a
            # dependency on libpython that we want to avoid when not in a
            # Python environment.  Therefore, casting the resulting void*
            # pointer to PyObject* using ctypes.
            py_err = ctypes.cast(ctypes.c_void_p(py_err), ctypes.py_object).value
    
        tb = py_err.__traceback__
    
        # The py_err.__traceback__ only goes from the location thrown
        # up to the next FFI handoff.  To have the stacktrace also
        # include the C++ side, we need to adjust the __traceback__
        # before re-throwing.
        backtrace = _LIB.TVMGetLastBacktrace()
        if backtrace:
            frames = re.split(r"\n\W+\d+:\W+", py_str(backtrace))
            frames = frames[1:]  # Skip "Stack trace: "
    
            for frame in frames:
                if " at " in frame:
                    func_name, frame = frame.split(" at ", 1)
                    filename, lineno = frame.rsplit(":", 1)
                    func_name = func_name.strip()
                    filename = filename.strip()
                    lineno = int(lineno.strip())
    
                    tb = _append_traceback_frame(tb, func_name, filename, lineno)
    
        # Remove stack frames that provide little benefit to
        # debugging.  These are only removed from the stack frames
        # contained within the exception we are re-raising, and not to
        # the stack frames that it will continue to collect.
        # Therefore, there may still be a single instance of these
        # frames in the outermost Python-to-FFI call.
        filter_funcs = [
            lambda code: "tvm/_ffi/_ctypes/packed_func.py" in code.co_filename,
            lambda code: "tvm/_ffi/base.py" in code.co_filename,
        ]
        tb = _filter_traceback_frames(tb, filter_funcs)
    
        py_err = py_err.with_traceback(tb)
    
        # The exception PyObject may contain a large amount of state,
        # including all stack frames that may be inspected in a later
        # PDB post-mortem.  Therefore, we must make sure to remove the
        # underlying PyObject* from the C++ side after we retrieve it.
        _LIB.TVMDropLastPythonError()
    
>       raise py_err
E       tvm._ffi.base.TVMError: Traceback (most recent call last):
E         8: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::IRModule (tvm::transform::Pass, tvm::IRModule)>::AssignTypedLambda<tvm::transform::__mk_TVM9::{lambda(tvm::transform::Pass, tvm::IRModule)#1}>(tvm::transform::__mk_TVM9::{lambda(tvm::transform::Pass, tvm::IRModule)#1}, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, tvm::runtime::TVMRetValue)
E         7: tvm::transform::Pass::operator()(tvm::IRModule) const
E         6: tvm::transform::Pass::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
E         5: tvm::transform::ModulePassNode::operator()(tvm::IRModule, tvm::transform::PassContext const&) const
E         4: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<tvm::IRModule (tvm::IRModule, tvm::transform::PassContext)>::AssignTypedLambda<tvm::relay::transform::InferType()::{lambda(tvm::IRModule, tvm::transform::PassContext const&)#1}>(tvm::relay::transform::InferType()::{lambda(tvm::IRModule, tvm::transform::PassContext const&)#1})::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*)
E         3: tvm::DiagnosticContext::Render()
E         2: tvm::DiagnosticRenderer::Render(tvm::DiagnosticContext const&)
E         1: tvm::runtime::PackedFuncObj::Extractor<tvm::runtime::PackedFuncSubObj<tvm::runtime::TypedPackedFunc<void (tvm::DiagnosticContext)>::AssignTypedLambda<tvm::TerminalRenderer(std::ostream&)::{lambda(tvm::DiagnosticContext const&)#1}>(tvm::TerminalRenderer(std::ostream&)::{lambda(tvm::DiagnosticContext const&)#1})::{lambda(tvm::runtime::TVMArgs const&, tvm::runtime::TVMRetValue*)#1}> >::Call(tvm::runtime::PackedFuncObj const*, tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*)
E         0: tvm::ReportAt(tvm::DiagnosticContext const&, std::ostream&, tvm::Span const&, tvm::Diagnostic const&)
E         File "/proj_sw/user_dev/pchandrasekaran/latest/Forge/tt-forge-fe/third_party/tvm/src/ir/diagnostic.cc", line 267
E       TVMError: The source maps are not populated for this module. Please use `tvm.relay.transform.AnnotateSpans` to attach source maps for error reporting.
E       Error: The Relay type checker is unable to show the following types match.
E       In particular `Tensor[(1, 58), float32]` does not match `Tensor[(1, 58), int64]`

```
The error is raised because of the position_id calculation inside the LlamaModelWrapper 
Repro:
```
def test_postion_id_calculation():
    
    class PositionIdModelWrapper(torch.nn.Module):
        def __init__(self):
            super().__init__()
            
        def forward(self, attention_mask):
            position_ids = attention_mask.long().cumsum(-1) - 1
            position_ids.masked_fill_(attention_mask == 0, 1)
            return position_ids
     
    framework_model = PositionIdModelWrapper()
    framework_model.eval()
    
    attention_mask = torch.zeros((1, 20)).to(torch.int64)
    attention_mask[: , :12] = 1

    framework_out = framework_model(attention_mask)
    compiled_model = forge.compile(framework_model, sample_inputs=[attention_mask])
```

forge/test/mlir/llama/tests/test_llama_decode.py
```diff
@@ -0,0 +1,321 @@
...
+        if run_on_tt_device:
+            # Run on TT device
+            tt_inputs = [model_inputs[0], attention_mask, position_ids, model_inputs[1]]
+            tt_output = compiled_model(*tt_inputs)
+            tt_output = [tt_out.to("cpu") for tt_out in tt_output]
+
+            # Validate TT result with Framework
+            assert all(
+                [
+                    compare_with_golden_pcc(golden=fw_out, calculated=tt_out, pcc=0.99)
```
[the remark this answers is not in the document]
chandrasekaranpradeep: Will replace `compare_with_golden_pcc` with `compare_with_golden`

forge/test/mlir/llama/tests/test_llama_decode.py
```diff
@@ -0,0 +1,321 @@
...
+            break
+
+        model_inputs = [next_token.unsqueeze(0)]
+
+        # Formate the past key values from List(Key1, Values1, ... , Key26, Values26) to
+        # List(List(Key1, Values1), ... , List(Key26, Values26))
+        model_inputs.append([past_key_values[idx : idx + 2] for idx in range(0, len(past_key_values), 2)])
+
+        for idx in range(len(model_inputs[1])):
+            model_inputs[1][idx][0][:, :, non_padding_past_key_values_seq_length, :] = model_inputs[1][idx][0][
```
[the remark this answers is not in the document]
chandrasekaranpradeep: Sure Will add a comment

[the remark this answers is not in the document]
chandrasekaranpradeep: Added Comment

forge/test/mlir/llama/tests/test_llama_decode.py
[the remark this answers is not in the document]
chandrasekaranpradeep: Updated with `compare_with_golden`

forge/test/mlir/llama/tests/test_llama_decode.py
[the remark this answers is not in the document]
chandrasekaranpradeep: Updated with negative index

forge/test/mlir/llama/tests/test_llama_decode.py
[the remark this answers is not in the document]
chandrasekaranpradeep: Sure Will create a issue

Issue #520, opened 2024-10-22
Relay type checker is unable to show the following types match:   Tensor[(1, 255, 15, 16), float32]   Tensor[(1, 255, 16, 16), float32]


kamalrajkannan78:
## Description

-  Relay type checker is unable to show the following types match:   Tensor[(1, 255, 15, 16), float32]   Tensor[(1, 255, 16, 16), float32] in yolov3_tiny_holli_pytorch variant

## Reproduce

```
git checkout kkannan/forge_model_issues
git submodule update --recursive
export BACKEND_ARCH_NAME=wormhole_b0 ARCH_NAME=wormhole_b0 PYBUDA_FORCE_EMULATE_HARVESTED=1 
pytest forge/test/model_demos/high_prio/cnn/pytorch/test_yolo_v3.py::test_yolov3_tiny_holli_pytorch[no_device] -svv 
```

## Observed Behaviour
```
The Relay type checker is unable to show the following types match:
  Tensor[(1, 255, 15, 16), float32]
  Tensor[(1, 255, 16, 16), float32]
In particular:
  dimension 2 conflicts: 15 does not match 16.
The Relay type checker is unable to show the following types match.
In particular `Tensor[(1, 255, 16, 16), float32]` does not match `Tensor[(1, 255, 15, 16), float32]`
2024-10-22 10:16:49.600 | ERROR    | tvm.relay.op.contrib.forge.forge_passes:run_pattern_callbacks:4012 - Failed on "DecomposeDynamicResize2d" TVM callback
```

## Logs

- [yolov3_hollli.log](https://github.com/user-attachments/files/17474527/oct22_yolov3_hollli.log)

PR #616, opened 2024-11-05
Fix: Invalid padding calculation in FuseConvAndPoolPadding PatternCallback

ashokkumarkannan1:
Fix https://github.com/tenstorrent/tt-forge-fe/issues/520

Faced below issues
1. Conv2d - Failed to run MLIR compiler pass pipeline - https://github.com/tenstorrent/tt-forge-fe/issues/618
2. MaxPool2d
    a. interleaved-buffers page size should be divisible by buffer size- https://github.com/tenstorrent/tt-forge-fe/issues/617
    b. Page size must be divisible by sizeof(uint32_t) because buffers hold uint32_t values - https://github.com/tenstorrent/tt-forge-fe/issues/627

### DIFF
--- a/forge/test/mlir/test_ops.py
+++ b/forge/test/mlir/test_ops.py
@@ -1364,3 +1364,86 @@ def test_avg_pool2d():
     co_out = [co.to("cpu") for co in co_out]
     fw_out = [fw_out] if isinstance(fw_out, torch.Tensor) else fw_out
     assert all([compare_with_golden(golden=fo, calculated=co, pcc=0.99) for fo, co in zip(fw_out, co_out)])
+
+
+@pytest.mark.parametrize("shape", [(1, 3, 32, 32)])
+@pytest.mark.parametrize(
+    "padding",
+    [
+        pytest.param(
+            (1, 1, 1, 1),
+            marks=pytest.mark.xfail(reason="'ttnn.conv2d' op Bias must only have data on the final dimenstion"),
+        ),
+        pytest.param(
+            (1, 1, 2, 2),
+            marks=pytest.mark.xfail(reason="'ttnn.conv2d' op Bias must only have data on the final dimenstion"),
+        ),
+        pytest.param(
+            (1, 2, 1, 2),
+            marks=pytest.mark.xfail(
+                reason="TTNN only supports padding height/width attributes. Thus, padding_top "
+                "must equal padding_bottom for the op to execute as expected."
+            ),
+        ),
+    ],
+)
+@pytest.mark.xfail(reason="'ttnn.conv2d' op Bias must only have data on the final dimenstion")
+def test_conv2d_with_padding(shape, padding):
+    class PaddingAndConv2d(nn.Module):
+        def __init__(self, padding):
+            super().__init__()
+            self.padding = padding
+            self.conv = nn.Conv2d(3, 3, kernel_size=3, stride=1, padding=0)
+
+        def forward(self, x):
+            x = nn.functional.pad(x, self.padding, mode="constant", value=0)
+            return self.conv(x)
+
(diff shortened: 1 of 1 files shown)


forge/test/mlir/test_ops.py
```diff
@@ -1032,3 +1032,65 @@ def forward(self, x):
...
+
+@pytest.mark.parametrize("shape", [(1, 3, 32, 32)])
+@pytest.mark.parametrize(
+    "padding",
+    [
+        (1, 1, 1, 1),
+        (1, 1, 2, 2),
+    ],
+)
+@pytest.mark.xfail(reason="Only BFLOAT16 supported for now")
```
nvukobratTT, line 1076: Can we use bfloat16 in this test instead of marking it as xfailed?

forge/test/mlir/test_ops.py
```diff
@@ -1032,3 +1032,65 @@ def forward(self, x):
...
+
+@pytest.mark.parametrize("shape", [(1, 3, 32, 32)])
+@pytest.mark.parametrize(
+    "padding",
+    [
+        (1, 1, 1, 1),
+        (1, 1, 2, 2),
+    ],
+)
+@pytest.mark.xfail(reason="'ttnn.conv2d' op Bias must only have data on the final dimenstion")
```
nvukobratTT, line 1045: Can you share the trace error log for this test?

forge/test/mlir/test_ops.py
ashokkumarkannan1: On converting both framework_model and inputs to torch.bfloat16,

- Testcase 1 hit with `For valid non-interleaved buffers page size 22 must equal buffer size 724. For interleaved-buffers page size should be divisible by buffer size`
- Testcase 2 hit with `Page size must be divisible by sizeof(uint32_t) because buffers hold uint32_t values`

Shall I create issues respectively for both issues @nvukobratTT ?

### Logs:
[test_padding_with_maxpool2d.log](https://github.com/user-attachments/files/17630509/test_padding_with_maxpool2d.log)

forge/test/mlir/test_ops.py
```diff
@@ -1320,3 +1320,73 @@ def forward(self, x):
...
     fw_out = [fw_out] if isinstance(fw_out, torch.Tensor) else fw_out
     assert all([compare_with_golden(golden=fo, calculated=co, pcc=0.99) for fo, co in zip(fw_out, co_out)])
+
+
+@pytest.mark.parametrize("shape", [(1, 3, 32, 32)])
+@pytest.mark.parametrize(
+    "padding",
+    [
+        (1, 1, 1, 1),
+        (1, 1, 2, 2),
```
nvukobratTT, line 1374: Let's add cases with different padding shapes, but mark those as xfail with a proper comment :)) E.g. unsupported on TTNN op level or something similar

forge/test/mlir/test_ops.py
```diff
@@ -1320,3 +1320,73 @@ def forward(self, x):
...
+    ],
+)
+@pytest.mark.xfail(reason="'ttnn.conv2d' op Bias must only have data on the final dimenstion")
+def test_padding_with_conv2d(shape, padding):
```
nvukobratTT, line 1378: Let's rename it to `test_conv2d_with_padding`. Let's focus on the op type first for easier search. And if not already, put this test next to the existing conv2d test :))

forge/test/mlir/test_ops.py
ashokkumarkannan1: Updated

forge/test/mlir/test_ops.py
ashokkumarkannan1: Updated

PR #647, opened 2024-11-07
Dumping ttir and ttnn into json files

### LINKED ISSUE #149: [debugging] save lowered TTIR into a file
After lowering to `TTIR` save the `mlir` input in a file, which can be seen by `reportify`.

Also, it could be useful to save the resulting `TTNN` mlir output (maybe even the generated flatbuffer).

vkovinicTT:
This PR addresses issue #149 by adding functionality to `reportify` for formatting and storing **MLIR input** (`ttir`) and **MLIR output** (`ttnn`) in a JSON file. With this update, running any module will save MLIR information within the path `testify/ll-sw/%MODULE_NAME%/mlir_reports` as `ttir.mlir` and `ttnn.mlir`.

Fix #149

### DIFF
--- a/forge/csrc/passes/mlir_compiler.cpp
+++ b/forge/csrc/passes/mlir_compiler.cpp
@@ -78,7 +78,7 @@ runtime::Binary run_mlir_compiler(tt::ForgeGraphModule& module)
 
     // save what's dumped to a file named "{name}.mlir"
     auto mlir_modulee = mlir_module.get();
-    reportify::dump_mlir("ttnn", &mlir_modulee);    
+    reportify::dump_mlir("ttnn", &mlir_modulee);
 
     // Generate binary from the MLIR module.
     auto binary = mlir::tt::ttnn::ttnnToFlatbuffer(mlir_module.get());
--- a/forge/csrc/passes/mlir_compiler.cpp
+++ b/forge/csrc/passes/mlir_compiler.cpp
@@ -34,6 +34,9 @@
 #include "ttmlir/Dialect/TTNN/IR/TTNN.h"
 #include "ttmlir/Target/TTNN/TTNNToFlatbuffer.h"
 
+// Reportify headers
+#include "reportify/reportify.hpp"
+
 namespace tt::passes
 {
 /// Public API for lowering to MLIR, running MLIR passes and generate runtime binary.
@@ -72,6 +75,10 @@ runtime::Binary run_mlir_compiler(tt::ForgeGraphModule& module)
     tt::log_info(LogMLIRCompiler, "MLIR passes run successfully.");
 
     mlir_module->dump();
+
+    // save what's dumped to a file named "{name}.mlir"
+    auto mlir_modulee = mlir_module.get();
+    reportify::dump_mlir("ttnn", &mlir_modulee);    
 
     // Generate binary from the MLIR module.
     auto binary = mlir::tt::ttnn::ttnnToFlatbuffer(mlir_module.get());
--- a/forge/csrc/passes/mlir_compiler.cpp
+++ b/forge/csrc/passes/mlir_compiler.cpp
@@ -77,7 +77,7 @@ runtime::Binary run_mlir_compiler(tt::ForgeGraphModule& module)
     mlir_module->dump();
 
     // save what's dumped to a file named "{name}.mlir"
-    reportify::dump_mlir("ttnn",  mlir_module->getName()->str(), mlir_module.get());
+    reportify::dump_mlir("ttnn", mlir_module->getName()->str(), mlir_module.get());
 
     // Generate binary from the MLIR module.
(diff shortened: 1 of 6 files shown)


forge/csrc/passes/mlir_compiler.cpp
```diff
@@ -73,6 +76,10 @@ runtime::Binary run_mlir_compiler(tt::ForgeGraphModule& module)
 
     mlir_module->dump();
 
+    // save what's dumped to a file named "{name}.mlir"
+    auto mlir_modulee = mlir_module.get();
```
nvukobratTT, line 80: Can we use concreate types, instead of `auto`?

forge/csrc/reportify/reportify.cpp
```diff
@@ -425,5 +458,29 @@ void dump_graph(
     std::string default_dir = get_default_reportify_path("");
     dump_graph(default_dir, test_name, graph_prefix, graph, report_path);
 }
+
+void dump_mlir(const std::string& name, mlir::ModuleOp* module)
```
nvukobratTT, line 462: Let's add a brief function description

forge/csrc/reportify/reportify.cpp
```diff
@@ -425,5 +458,29 @@ void dump_graph(
...
+{
+    if (env_as<bool>("FORGE_DISABLE_REPORTIFY_DUMP"))
+        return;
+
+    std::string path = get_default_reportify_path("");
+    std::string report_path = get_mlir_reports_relative_directory();
+
```
nvukobratTT, line 469: Let's remove the new line here to get a cleaner function blocking (e.g. first block is to build the report path).

forge/csrc/reportify/reportify.cpp
```diff
@@ -425,5 +458,29 @@ void dump_graph(
...
+    std::string full_report_path = build_report_path(path, module->getName()->str(), report_path);
+
+    JsonNamePairs json_pairs = create_jsons_for_mlir(name, module);
+    json root_json = json_pairs.back().first;
+
```
nvukobratTT, line 474: Same here, the second block is building json name or full json path. 

Not major requirement, mostly personal preferences for code readability :))

forge/csrc/reportify/reportify.cpp
```diff
@@ -425,5 +458,29 @@ void dump_graph(
...
+    std::string root_json_name = json_pairs.back().second;
+    std::transform(root_json_name.begin(), root_json_name.end(), root_json_name.begin(), ::tolower);
+
+    std::string root_json_path = full_report_path + root_json_name;
+
+    std::filesystem::create_directories(std::filesystem::path(full_report_path));
```
nvukobratTT, line 480: Final block can bo json storing

forge/csrc/passes/mlir_compiler.cpp
pilkicTT: I'm more concerned about what `OwningOpRef<T>::get()` does under the hood... If it copies the underlaying object, then it could be problematic.

We are also using it for running `ttnnToFlatbuffer()`, so I would at least use the same `ModuleOp` object there (`milr_modulee`).

pilkicTT (commented):
Looks good in general, I'll take a more detailed look into `OwningOpRef<T>::get()` and get back to you...

FYI you have a merge commit in your commits; not exactly important, since we will squash it before merging the PR, but in general stick to rebase... :smile:

Also, when merging, take care so that the squashed commit contains the PR description and title. Personally, when sending the PR, I squash everything myself and write the entire PR description into my commit. That way, when creating the PR, all of the details will be filled automatically and will also ensure, that when I finally merge the PR, it will contain the correct message. In most cases this is a good practice IMO...

forge/csrc/passes/mlir_compiler.cpp
pilkicTT: I'm not sure why the `OwningOpRef<T>::get()` doesn't return by ref, but anyhow... After looking into the mlir implementation, it looks like all `mlir::Op`s (including `mlir::ModuleOp`) have a pointer inside of them and they are using default copy construction, so there's no deep copies being performed. No problems there...

However, I don't like the need of having this additional variable `mlir_modulee` just to be able to pass it by ref. So, by looking at the `ttnnToFlatbuffer()` (which can take `mlir_module.get()`), it might be good idea for `dump_mlir()` to take an `mlir::Operation*` as well. This enables us to just call `reportify::dump_mlir("ttnn", mlir_module.get())` and also to dump any operation in the future (e.g. a single function or a single operator).

pilkicTT (approved):

nvukobratTT (approved):

nvukobratTT:
@vkovinicTT minor comment, can we open the following issues with the tech dept label: Visualizing stored MLIR IR in Visualizer (e.g. reportify).

There is no need to work on it as a next task, but let's keep it as something to take a look in the future :))

PR #655, opened 2024-11-08
Resolve bugs in generated single op tests

chandrasekaranpradeep:
Fixes https://github.com/tenstorrent/tt-forge-fe/issues/643
1. Add input node dtypes as data member in Operation class
2. In `_determine_node_dtype` function before checking **framework_dtype is N/A**, add a condition to check whether the framework_dtype key is present in the node attrs because the kernel/op node doesn't have framework_dtype as key inside the node attrs.
3. In `write_pytest_function` method in `python_codegen.py file`, currently we are creating rand input tensor for all the dtypes, instead for floating dtype we will use torch.rand function and for integer datatype we will use torch.randint with high value equal to 1000. 
Note: For torch.randint high value is set to 1000 instead of the respective integer dtype max value because there is some special case like passing input_ids to embeding op and input_ids will be integer datatype. In that case there is possibility of generating higher integer value by torch.randint which is greater than vocab_size in embedding op, which leads error in embedding op execution.
4. In generated pytest function by write_pytest_function method, framework_output will be of TensorFromTrace, converted the TensorFromTrace to the particular framework and then passed to the compare_with_golden function
5. Instead of passing dumping input tensor to the forge module forward function, if all the inputs for the particular op is parameter or const, we will pass the parameter/const as input to the forge module forward function and in the pytest function instead generating rand values for the parameter/const as input to the forge module forward function, will load the actual values from the model parameter/buffers.For the particular op, if only the some inputs are parameter/const, we will not pass parameter/const as input to the forge module forward function. Eg: For matmul op, if operand_A is forward function input and operand_B is parameter, it will remain same for single generated op.

### DIFF
--- a/forge/forge/python_codegen.py
+++ b/forge/forge/python_codegen.py
@@ -7,6 +7,8 @@ import torch
 import numpy as np
 import tensorflow as tf
 from loguru import logger
+import forge
+from forge.tensor import forge_dataformat_to_pytorch_dtype
 
 
 def forge_df_str_from_str(df: str, name: str):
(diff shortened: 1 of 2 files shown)


forge/forge/python_codegen.py
```diff
@@ -53,6 +55,51 @@ def forge_df_str_from_str(df: str, name: str):
     else:
         logger.warning(f"Invalid data format: {df} for constant/parameter '{name}', defaulting to float32")
         return "forge.DataFormat.Float32"
+    
+def forge_df_from_str(df: str):
```
nvukobratTT, line 59: Can we merge this one and the function above? Seems like they can be single-function pretty easily

forge/forge/python_codegen.py
```diff
@@ -953,7 +1000,7 @@ def write_param_parser(
         else:
             assert False, "TODO: Add other framework param parsers"
 
-    def write_pytest_function(self, module_name, input_shapes):
+    def write_pytest_function(self, module_name, framework, input_shapes, input_dtypes, activation_param_names, need_process_framework_parameters_func=False, param_file_name=None, names_params_file_name=None, named_buffers_file_name=None):
```
nvukobratTT, line 1003: If we have a long list of arguments, let's see to create a common object that we can use.

forge/forge/python_codegen.py
```diff
@@ -969,27 +1016,56 @@ def write_pytest_function(self, module_name, input_shapes):
...
         """
         self.wl("")
         self.wl("")
         self.wl("def test_module():")
         self.indent += 1
-        self.wl("inputs = [")
-        self.indent += 1
-        for shape in input_shapes:
-            self.wl(f"Tensor.create_from_torch(torch.rand({shape})),")
+        if names_params_file_name is not None and named_buffers_file_name is not None and len(activation_param_names)!=0:
```
nvukobratTT, line 1026: Let's add brief doc blocks explaining different approaches. Short sentences, no need to go into every aspect. Let's just capture a main point here, the rest is in the code.

forge/forge/tvm_to_python.py
```diff
@@ -2735,40 +2750,79 @@ def generate_op_tests(
...
+
+        forward_method_inputs = {}
+        test_inputs_shapes = []
+        test_inputs_dtypes = []
+        activation_param_names = []
+        needed_params = {}
+        needed_consts = {}
+        
+        # Lambda function to check whether the input for the single op is const or parameter based upon the input names.
+        is_params_const = lambda inp_name: True if (inp_name.startswith("const") or inp_name.endswith(".weight") or "." in inp_name) else False
```
nvukobratTT, line 2778: Next to checking name, can we have more robust way of determining if type is param or constant?

chandrasekaranpradeep: In the Operation class itself, I will add additional data member about the operand type info like Parameter/Constant/Activation by comparing the inputs node id with params and const dict(Key: node id, value: tuple(param_name, param_shape, require_grad, datatype))

chandrasekaranpradeep:
@nvukobratTT  FYI, the unique op generation has some changes with respect to this PR in write_pytest_functions in ForgeWriter. will update the PR with above comments and unique op generation too.

nvukobratTT:
> @nvukobratTT FYI, the unique op generation has some changes with respect to this PR in write_pytest_functions in ForgeWriter. will update the PR with above comments and unique op generation too.

Sure, just make sure to link this PR to both issues :))

nvukobratTT:
@chandrasekaranpradeep what is the state of this PR?

chandrasekaranpradeep:
@nvukobratTT Bug fixing in the generation of single op test is merged into main in  [this PR](https://github.com/tenstorrent/tt-forge-fe/pull/715), so closing this issue

PR #669, opened 2024-11-12
Fix codegen failure

meenakshiramanathan1:
Fix https://github.com/tenstorrent/tt-forge-fe/issues/506
This PR fixes the issue `AssertionError: Setting a tensor value of incorrect shape: (1, 256, 16, 2, 16) vs torch.Size([1, 256, 16, 16, 2])`, which was caused due to shape Vs eval functions output shape mismatch in stack op.
More precisely, shape function output was wrong because insertion was not being done correctly [here](https://github.com/tenstorrent/tt-forge-fe/blob/f7db298a933741a3ccbfd3e3da382d0992e67f07/forge/forge/op/eval/forge/eltwise_nary.py#L147).
To ensure proper insertion at the right axis, changes with axis has been made in shape and decompose function of stack op in eltwise_nary file.

After these changes, it was failing with `AttributeError: module 'torch' has no attribute 'uint1'` during evaluation of cast op , for which uint1 has been mapped to bool dtype.

### DIFF
--- a/forge/forge/op/eval/forge/eltwise_nary.py
+++ b/forge/forge/op/eval/forge/eltwise_nary.py
@@ -144,7 +144,13 @@ def shape(type, attr, ops) -> Tuple[Tuple, List]:
         axis = attr[0]
 
         output_shape = list(ops[0])
-        output_shape.insert(axis, len(ops))
+        if axis == -1:
+            output_shape.append(len(ops))
+        elif axis < -1:
+            # axis + 1 is added because insertion at the correct position requires adjusting for negative axes to ensure proper behavior.
+            output_shape.insert(axis + 1, len(ops))
+        else:
+            output_shape.insert(axis, len(ops))
         return output_shape, []
 
     elif type == "interleave":
@@ -246,11 +252,17 @@ def decompose(type, attr, dc, inputs):
         new_inputs = []
         for inp in inputs:
             inp_shape = inp.shape.as_list()
-            inp_shape.insert(axis, 1)
-            new_inp = dc.op("reshape", [inp], (*inp_shape,))
+            if axis == -1:
+                inp_shape.append(1)
+            elif axis < -1:
+                # axis + 1 is added because insertion at the correct position requires adjusting for negative axes to ensure proper behavior.
+                inp_shape.insert(axis + 1, 1)
+            else:
+                inp_shape.insert(axis, 1)
+            new_inp = dc.op_with_named_attrs("reshape", [inp], {"shape": (*inp_shape,)}, (*inp_shape,))
             new_inputs.append(new_inp)
 
-        output = dc.op("concatenate", new_inputs, (axis,))
+        output = dc.op_with_named_attrs("concatenate", new_inputs, {"dim": (axis)}, (axis,))
         dc.fuse(output)
 
     if type == "concatenate":
--- a/forge/forge/python_codegen.py
+++ b/forge/forge/python_codegen.py
@@ -76,6 +76,8 @@ def pytorch_df_str_from_str(df: str, name):
         return "torch.int16"
     elif df == "int64":
         return "torch.int64"
+    elif df == "uint1":
+        return "torch.bool"
     else:
         logger.warning(f"Invalid data format: {df} for constant/parameter '{name}', defaulting to float32")
         return "torch.float32"
--- a/forge/forge/tvm_to_python.py
+++ b/forge/forge/tvm_to_python.py
@@ -23,7 +23,7 @@ import os
 import sys
 import importlib
 
-from forge.python_codegen import PyTorchWriter, ForgeWriter, PythonWriter
+from forge.python_codegen import PyTorchWriter, ForgeWriter, PythonWriter, pytorch_df_str_from_str
 
 
 def populate_torch_all_to_args(graph, nid, compiler_cfg):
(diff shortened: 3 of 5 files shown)


forge/forge/tvm_to_python.py
```diff
@@ -1239,6 +1239,8 @@ def populate_cast_args(graph, nid, compiler_cfg):
     node = graph["nodes"][nid]
     args = []
     dtype = node["attrs"]["dtype"][0][0]
+    if dtype == "uint1":
```
nvukobratTT, line 1242: Why can't we use uint1, and have to use bool precisely? Those are the same, but why true/false vs 1/0? 

Can we do this in TVM? AFAIK there is the logic that converts dtypes from pytorch to Forge dtypes (and vice verse). It seems like that would be the proper place to do so.

meenakshiramanathan1: @nvukobratTT ,We can't use uint1 as it was throwing `AttributeError: module 'torch' has no attribute 'uint1'` and currently there is support only for the dtypes mentioned [here](https://github.com/tenstorrent/tt-tvm/blob/e91c07cfe4710c72a9976c791205032aff60d52b/python/tvm/relay/frontend/pytorch.py#L2451).

meenakshiramanathan1: @nvukobratTT , it has been changed that way because uint1 is not supported and it was throwing error `AttributeError: module 'torch' has no attribute 'uint1'`

I will check it out and let you know about the tvm part.

meenakshiramanathan1:
> @meenakshiramanathan1 can we please have a proper overview of changes in PR description and how they solve the issue. Also issue contains only repro and the message but fails to explain what is the cause of the issue.

It has been updated now. Please let me know if any other details are required.

dgolubovicTT (approved):
Thanks for providing detailed explanation. It really speeds up the review process.

forge/forge/op/eval/forge/eltwise_nary.py
```diff
@@ -144,7 +144,12 @@ def get_eltwise_shape_and_broadcast():
         axis = attr[0]
 
         output_shape = list(ops[0])
-        output_shape.insert(axis, len(ops))
+        if axis == -1:
+            output_shape.append(len(ops))
+        elif axis < 0:
+            output_shape.insert(axis + 1, len(ops))
```
dgolubovicTT, line 150: can we put elif axis < -1 just to be precise. Also can you add just one line of comment why we use + 1 to handle inserting new axis. This is just to help someone that lands in this part of the code and doesn't have the context...

forge/forge/op/eval/forge/eltwise_nary.py
```diff
@@ -246,11 +251,16 @@ def decompose(type, attr, dc, inputs):
         new_inputs = []
         for inp in inputs:
             inp_shape = inp.shape.as_list()
-            inp_shape.insert(axis, 1)
-            new_inp = dc.op("reshape", [inp], (*inp_shape,))
+            if axis == -1:
+                inp_shape.append(1)
+            elif axis < 0:
+                inp_shape.insert(axis + 1, 1)
```
dgolubovicTT, line 257: same as above...

forge/forge/op/eval/forge/eltwise_nary.py
meenakshiramanathan1: Sure @dgolubovicTT , I will make the changes.

forge/forge/python_codegen.py
```diff
@@ -76,6 +76,8 @@ def pytorch_df_str_from_str(df: str, name):
         return "torch.int16"
     elif df == "int64":
         return "torch.int64"
+    elif df == "uint1":
+        return "torch.bool"
     else:
         logger.warning(f"Invalid data format: {df} for constant/parameter '{name}', defaulting to float32")
```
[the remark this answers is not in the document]
dgolubovicTT: Noted, but can we expand on idea offline during planning of dataformats effort

forge/forge/op/eval/forge/eltwise_nary.py
ashokkumarkannan1: updated.

forge/forge/python_codegen.py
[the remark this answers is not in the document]
nvukobratTT: Ofc, we should create a brief propose doc that will capture what we're going to do, and what it's going to impact potentially. This will be a great mediator in our further planning of for data formats effort

forge/test/mlir/test_ops.py
```diff
@@ -1663,3 +1663,44 @@ def forward(self, inp_1):
     co_out = [co.to("cpu") for co in co_out]
     fw_out = [fw_out] if isinstance(fw_out, torch.Tensor) else fw_out
     assert all([compare_with_golden(golden=fo, calculated=co, pcc=0.99) for fo, co in zip(fw_out, co_out)])
+
+
+pytest.mark.xfail(
```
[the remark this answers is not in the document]
kamalrajkannan78: Yes we have https://github.com/tenstorrent/tt-mlir/issues/1097 . 
TTIR graphs are present in attached log files @nvukobratTT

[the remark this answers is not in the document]
nvukobratTT: Can we extract TTIR files from the log and place them in the issue? Those are the most important details for MLIR folks, and for quicker reference, let's keep TTIR in the issue description as well :))

Ofc, please keep the log files for more granular overview

PR #672, opened 2024-11-12
Fix const eval input tensor stride and update tests

### LINKED ISSUE #609: Runtime error  : stride mismatch
## Description

- Stride mismatch runtime error in wideresnet 

## Reproduce

```
git checkout kkannan/stride_issue
pytest forge/test/mlir/test_ops.py -k " stride" -svv
```

## Logs

- [wideresnet_stride_issue.log](https://github.com/user-attachments/files/17615850/xx_nov4_stride_s_2.log)

### LINKED ISSUE #673: Const eval input tensor stride mismatch
When transpose is part of the const eval graph, it causes const eval input tensor to have different stride than compiled model expects. The issue is that compiled model expects the stride that is consistent with current shape of tensor. We should align constant input stride with its shape.

dgolubovicTT:
Closes #673. Closes #609.
- Fix const eval input tensor stride.
- Change resnet test: Run compiled ResNet model
- Remove xfail from Llama mlp and lm_head tests since they pass now.

### DIFF
--- a/forge/forge/tensor.py
+++ b/forge/forge/tensor.py
@@ -1339,7 +1339,13 @@ def consteval_tensor(
 
 
 def consteval_input(consteval_trace, name: str, inputs: Dict[str, torch.Tensor], is_forge: bool) -> torch.Tensor:
-    return consteval_tensor(consteval_trace, name, inputs, is_forge, "Forward")
+    const_eval_tensor = consteval_tensor(consteval_trace, name, inputs, is_forge, "Forward")
+    # This: "torch.empty(const_eval_tensor.shape).copy_(const_eval_tensor)" will create tensor with contiguous memory layout consistent with its current shape.
+    # We are doing this because constant input tensors should have memory layout consistent with their shape.
+    # Sometimes, the stride is inconsistent with shape because some consteval operations might change the shape but not the stride.
+    # For example, if we had transpose in consteval graph, output tensor would have stride same as input.
+    # However, since we store that input as constant tensor, its shape defines its stride.
+    return torch.empty(const_eval_tensor.shape, dtype=const_eval_tensor.dtype).copy_(const_eval_tensor)
 
 
 def consteval_shape(compiled_graph_state, name: str, tensor: torch.Tensor, is_forge: bool = False) -> torch.Tensor:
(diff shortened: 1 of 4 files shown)


forge/test/mlir/resnet/test_resnet_inference.py
```diff
@@ -21,8 +22,14 @@ def test_resnet_inference():
     input_image = torch.rand(1, 3, 224, 224)
 
     # Sanity run
-    generation_output = framework_model(input_image)
-    print(generation_output)
+    fw_out = framework_model(input_image)
+    # print(fw_out)
```
nvukobratTT, line 26: Redundant line?

forge/test/mlir/resnet/test_resnet_inference.py
```diff
@@ -21,8 +22,14 @@ def test_resnet_inference():
...
 
     # Compile the model
     compiled_model = forge.compile(framework_model, input_image)
+    co_out = compiled_model(input_image)
+    print(co_out)
```
nvukobratTT, line 31: Do we need to print outputs?

forge/forge/tensor.py
```diff
@@ -1413,6 +1423,8 @@ def get_post_const_eval_tensors(
 
     for input_name in ordered_input_names:
         # Load input constant tensors for consteval
+        if input_name == "input_1_add_18":
```
nvukobratTT, line 1426: Redundant block?

forge/forge/tensor.py
```diff
@@ -1339,7 +1339,12 @@ def eval_op(op_type, inputs):
 
 
 def consteval_input(consteval_trace, name: str, inputs: Dict[str, torch.Tensor], is_forge: bool) -> torch.Tensor:
-    return consteval_tensor(consteval_trace, name, inputs, is_forge, "Forward")
+    const_eval_tensor = consteval_tensor(consteval_trace, name, inputs, is_forge, "Forward")
+    # We need to change stride of new tensor explicitly by creating new tensor with shape of const_eval_tensor,
```
nvukobratTT, line 1343: Can we layer this explanation? In sum, to answer the following questions in order:
1. What we're doing?
2. Why we're doing it?
3. Specifics, Notes, Comments, etc. about how we implement it for the compiler.

Explanation is good, but I think it'll be good just to structure it a bit more to be easier to follow :))

forge/test/mlir/resnet/test_resnet_inference.py
dgolubovicTT: yeah, we don't need it...

forge/forge/tensor.py
dgolubovicTT: Updated the comments.

nvukobratTT (approved):

pilkicTT (changes requested):
Let's sync on this offline, this doesn't seem to me like the correct approach... Maybe i'm missing something...

dgolubovicTT:
@pilkicTT I would merge this one to unblock consteval cases. I created a tech debt issue to follow up with investigation for general cases of inputs not having stride consistent with their shape

nvukobratTT:
> @pilkicTT I would merge this one to unblock consteval cases. I created a tech debt issue to follow up with investigation for general cases of inputs not having stride consistent with their shape

Let's link the new issues here, just to have a ref :))

pilkicTT (approved):
Let's follow up then

Issue #505, opened 2024-10-21
RuntimeError: Unhandled attribute type 


kamalrajkannan78:
## Description

-  RuntimeError: Unhandled attribute type

## Reproduce

```
git checkout kkannan/forge_model_issues_nov12
git submodule update --recursive
pytest forge/test/model_demos/high_prio/cnn/pytorch/test_ssd300_resnet50.py -svv
```

## Observed Behaviour
```
context = CompileContext(modules=[Module pt_ssd300_resnet50], graph_name='pt_ssd300_resnet50', compiler_cfg=CompilerConfig(enabl...unt=0, target_cycles_offset=0, forge_module=<forge._C.ForgeGraphModule object at 0x7f9f18f196f0>, compiled_binary=None)

    def run_mlir_compiler(context: CompileContext) -> CompileDepth:
        assert context.forge_module is not None
    
>       context.compiled_binary = forge._C.run_mlir_compiler(context.forge_module)
E       RuntimeError: Unhandled attribute type
```

## Logs

- [ssd300_resnet50.log](https://github.com/user-attachments/files/17716212/nov12_ssd300_resnet50_issue.log)

ashokkumarkannan1:
This issue seems resolved now and currently blocked with Unsupported clip op.

### Op Config
```
Found Unsupported operations while lowering from TTForge to TTIR in forward graph
Unsupported Ops at: RUN_MLIR_COMPILER stage
clip
		 Input_shape: [{2, 1, 7, 7}, ]
					 Attributes: clip{max: 1.000000e+00, min: 0.000000e+00}
```

[clip.log](https://github.com/user-attachments/files/17630023/clip.log)

nvukobratTT:
Good, I assume we can close up this issue?

ashokkumarkannan1:
yes @nvukobratTT 

Compile Depth for this clip model is changed to SPLIT_GRAPH in this patch. https://github.com/tenstorrent/tt-forge-fe/pull/636

ashokkumarkannan1:
@kamalrajkannan78  Can you update the reproduce step in Description for models you faced this issue here?

kamalrajkannan78:
description updated now with repro steps @ashokkumarkannan1

Issue #618, opened 2024-11-05
RuntimeError: Failed to run MLIR compiler pass pipeline.


ashokkumarkannan1:
### Description

MLIR compiler passes failed for ForgeModule constructed with Conv2D op

### Reproduce

```bash
git checkout akannan/fix-invalid-maxpool-pad
git submodule update --recursive
pytest -svv forge/test/mlir/test_ops.py::test_padding_with_conv2d
```

### Observed Behaviour

```
>       context.compiled_binary = forge._C.run_mlir_compiler(context.forge_module)
E       RuntimeError: Failed to run MLIR compiler pass pipeline.

```

[conv2d.log](https://github.com/user-attachments/files/17628305/conv2d.log)

PR #681, opened 2024-11-13
Fix: Unhandled attribute type for std::tuple<int, int, int>

ashokkumarkannan1:
Fix https://github.com/tenstorrent/tt-forge-fe/issues/505

The `std::tuple<int, int, int>` type handled and mapped to `llvm::SmallVector<mlir::Attribute, 3>` type

### Logs
Below five models blocked in this issue and now its fixed

1. Blazepose
- [test_blazepose_detector_pytorch_without_fix.log](https://github.com/user-attachments/files/17734013/test_blazepose_detector_pytorch_without_fix.log)
- [test_blazepose_detector_pytorch_with_fix.log](https://github.com/user-attachments/files/17734011/test_blazepose_detector_pytorch_with_fix.log) -  https://github.com/tenstorrent/tt-forge-fe/issues/682
2. Llama3
- [test_llama3_sequence_classification_instruct_without_fix.log](https://github.com/user-attachments/files/17734015/test_llama3_sequence_classification_instruct_without_fix.log)
- [test_llama3_sequence_classification_instruct_with_fix.log](https://github.com/user-attachments/files/17734014/test_llama3_sequence_classification_instruct_with_fix.log) - https://github.com/tenstorrent/tt-forge-fe/issues/683
- [test_llama3_sequence_classification_without_fix.log](https://github.com/user-attachments/files/17734017/test_llama3_sequence_classification_without_fix.log)
- [test_llama3_sequence_classification_with_fix.log](https://github.com/user-attachments/files/17734016/test_llama3_sequence_classification_with_fix.log) - https://github.com/tenstorrent/tt-forge-fe/issues/683
3. Mistral
- [test_mistral_without_fix.log](https://github.com/user-attachments/files/17734020/test_mistral_without_fix.log)
- [test_mistral_with_fix.log](https://github.com/user-attachments/files/17734019/test_mistral_with_fix.log) - https://github.com/tenstorrent/tt-forge-fe/issues/683
4. Mobilenet V1 SSD
- [test_mobilenet_v1_ssd_pytorch_1x1_without_fix.log](https://github.com/user-attachments/files/17734022/test_mobilenet_v1_ssd_pytorch_1x1_without_fix.log)
- [test_mobilenet_v1_ssd_pytorch_1x1_with_fix.log](https://github.com/user-attachments/files/17734021/test_mobilenet_v1_ssd_pytorch_1x1_with_fix.log) - https://github.com/tenstorrent/tt-forge-fe/issues/618
5. SSD 300 resnet 50
- [test_ssd300_resnet50_without_fix.log](https://github.com/user-attachments/files/17734025/test_ssd300_resnet50_without_fix.log)
- [test_ssd300_resnet50_with_fix.log](https://github.com/user-attachments/files/17734023/test_ssd300_resnet50_with_fix.log) - https://github.com/tenstorrent/tt-forge-fe/issues/618

### DIFF
--- a/forge/csrc/lower_to_forge/common.hpp
+++ b/forge/csrc/lower_to_forge/common.hpp
@@ -56,7 +56,6 @@ using ForgeOpAttr = ::std::variant<
     bool,
     int,
     float,
-    std::tuple<int, int, int>,
     std::vector<int>,
     std::vector<std::tuple<int, int, int>>,
     std::vector<std::tuple<int, int, int, int>>,


forge/csrc/passes/lower_to_mlir.cpp
```diff
@@ -187,6 +187,14 @@ class MLIRGenerator
                     }
                     return builder_.getArrayAttr(attributes);
                 }
+                else if constexpr (std::is_same_v<T, std::tuple<int, int, int>>)
```
nvukobratTT, line 190: Does this mean that we're only handling rank 3 tuples? Can we extend it to hand N-dimensional tuples :))

forge/test/model_demos/high_prio/cnn/pytorch/test_blazepose.py
```diff
@@ -28,8 +28,12 @@ def test_blazepose_detector_pytorch(test_device):
 
     # Load BlazePose Detector
     pose_detector = BlazePose()
-    pose_detector.load_weights("mediapipepytorch/blazepose.pth")
-    pose_detector.load_anchors("mediapipepytorch/anchors_pose.npy")
+    pose_detector.load_weights(
```
nvukobratTT, line 31: I don't think we'll be able to load Blaze Pose from the CCM repo at the moment. 

Let me double-check if we have a better mechanism to handle this type of input. In the meantime, let's do the following:
- Add a note here and run this model without weights
- See if there is a model version that has the option to download and cache weights. Maybe it'll be a good option to define this logic on our side. E.g. function to download and cache model weights from certain URLs, similarly to how HuggingFace is caching their weights (as part of .cache/* dir)

forge/csrc/passes/lower_to_mlir.cpp
ashokkumarkannan1: I too explored on this and couldn't find an exact way. All the models which I checked hit with this `std::tuple<int, int, int>`. I created a separate issue to explore on this  N-dimensional tuples after merging this PR. https://github.com/tenstorrent/tt-forge-fe/issues/705

pilkicTT (changes requested):
Which ops are hitting this issue? I would like to address this in a different manner, by fixing the source of the problem. If we want to lower the attribute as a vector, there is no reason why it shouldn't be a vector in the first place (as far as I know, most attributes are vectors, so this is probably a bug within a single op).

Basically, it would be ideal if we can keep this mapping 1-1.

At least, we should add a regression test with this change, so if anyone hits a problem with this in the future, it will make the problem more understandable.

nvukobratTT:
> To answer your question, For all 5 models I tested, Its coming from Reshape Op and the attribute is shape.

@ashokkumarkannan1 I agree with this one. From the information I see here, it seems like `ForgeOpAttr` with `std::tuple<int, int, int>,` is an edge case rather that we're getting as a consequence of a silent bug somewhere in the code. 

That said, I would suggest finding the root cause and correcting it (e.g. reshaping attributes should be `std::vector<std::tuple<int, int, int>>` in this case). 

_Note: There is a chance that we'll need tuple support from multiple ranks. However, for this case, this doesn't seem like it's needed._

ashokkumarkannan1:
@nvukobratTT @pilkicTT 

On further investigation it is found that only reshape to 3d is hit with type `std::tuple<int, int, int>` and for other dimensional reshapes like 1D, 2D, 4D, and 5D, it is hit with type `std::vector<int, std::allocator<int> >` and I confirmed this with sanity test and here is the [log](https://github.com/user-attachments/files/17762938/test_reshape.log) for the same.

The reason for this is order of arrangement of std::variant for declaring `ForgeOpAttr`. As 3D shape attribute have only 3 int values and `std::tuple<int, int, int>` present first at this variant, it is falling back to `std::tuple<int, int, int>` type instead std::vector<int>.

#### Before:
```cpp
using ForgeOpAttr = ::std::variant<
    std::string,
    bool,
    int,
    float,
    std::tuple<int, int, int>,
    std::vector<int>,
    std::vector<std::tuple<int, int, int>>,
    std::vector<std::tuple<int, int, int, int>>,
    ForgeOpAttrQueueDramLocs,
    ForgeKernelBroadcastInputs>;
```
And just by changing this order, This unhandled attribute issue was now solved.
#### After:
```cpp
using ForgeOpAttr = ::std::variant<
    std::string,
    bool,
    int,
    float,
    std::vector<int>,
    std::tuple<int, int, int>,
    std::vector<std::tuple<int, int, int>>,
    std::vector<std::tuple<int, int, int, int>>,
    ForgeOpAttrQueueDramLocs,
    ForgeKernelBroadcastInputs>;
```

NOTE: If we change this order, There will not be any case hit this `std::tuple<int, int, int>` type.


### Logs
[test_reshape_with_fix.log](https://github.com/user-attachments/files/17764817/test_reshape_with_fix.log)
[test_reshape.log](https://github.com/user-attachments/files/17764822/test_reshape.log)

nvukobratTT:
@ashokkumarkannan1 @pilkicTT what si the value of keeping the `std::tuple`? 

Can we fully remove the tuple, and use only vector as a reference?

ashokkumarkannan1:
> @ashokkumarkannan1 @pilkicTT what si the value of keeping the `std::tuple`?
> 
> Can we fully remove the tuple, and use only vector as a reference?

I too have the same point. I will update this patch accordingly @nvukobratTT . 

@pilkicTT  Can we know your thoughts in this?

forge/test/model_demos/high_prio/cnn/pytorch/test_blazepose.py
nvukobratTT: I send a request to get some sort of storage for these .pth, and future ONNX files. I'm not sure when we're going to get it, so let's put this variant on pause if there isn't any other suitable.

nvukobratTT (approved):

nvukobratTT:
Please create tech dept issue as reminder to explori this topic as discussed @ashokkumarkannan1

ashokkumarkannan1:
> Please create tech dept issue as reminder to explori this topic as discussed @ashokkumarkannan1

Sure @nvukobratTT . I have created a tech debt issue as discussed. https://github.com/tenstorrent/tt-forge-fe/issues/705

pilkicTT (approved):

PR #685, opened 2024-11-13
Add new PyTorch unary operators to test

vobojevicTT:
Add new unary operators: abs, cos, exp, neg, rsqrt, sin, and square.

### DIFF
--- a/forge/test/operators/pytorch/eltwise_unary/test_unary.py
+++ b/forge/test/operators/pytorch/eltwise_unary/test_unary.py
@@ -165,7 +165,21 @@ class TestCollectionData:
     __test__ = False
 
     implemented = TestCollection(
-        operators=["relu", "sqrt", "reciprocal", "sigmoid"],
+        operators=[
+            "relu",
+            "sqrt",
+            "reciprocal",
+            "sigmoid",
+            "abs",
+            # "absolute",     # alias for abs
+            "cos",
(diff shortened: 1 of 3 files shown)


forge/test/operators/pytorch/eltwise_unary/test_unary.py
```diff
@@ -234,6 +248,63 @@ class TestCollectionData:
...
+                DataFormat.Int32,
+            ],
+            math_fidelities=[MathFidelity.HiFi4],
+            failing_reason=FailingReasons.DATA_MISMATCH,
+        ),
+        TestCollection(
+            operators=["rsqrt"],
+            input_sources=[InputSource.FROM_HOST],
+            input_shapes=[(1, 2, 3, 4)],
+            dev_data_formats=[
```
dgolubovicTT, line 266: How do we construct these sets of dev_data_formats based on op?

dgolubovicTT (approved):

forge/test/operators/pytorch/eltwise_unary/test_unary.py
vobojevicTT: We use all data formats with fixed values for the input source, input shape, and math fidelity, and we test data formats for all operators. This one TestCollection is a part of failing rules - tests for testing data formats fail for these data format values.

PR #694, opened 2024-11-13
[ops] lower equal op

pilkicTT:


forge/test/mlir/test_ops.py
```diff
@@ -583,7 +583,45 @@ def forward(self, x, y):
...
     co_out = [co.to("cpu") for co in co_out]
     fw_out = [fw_out] if isinstance(fw_out, torch.Tensor) else fw_out
-    assert all([compare_with_golden_pcc(golden=fo, calculated=co, pcc=0.99) for fo, co in zip(fw_out, co_out)])
+    assert all([compare_with_golden(golden=fo, calculated=co) for fo, co in zip(fw_out, co_out)])
+
+
+@pytest.mark.parametrize(
+    "shape_x, shape_y",
+    [
+        ((1, 128, 28, 28), (1, 128, 28, 28)),
```
dgolubovicTT, line 592: It is a nit, but since torch.eq requires same shapes of operands, can we use single shape parameter.

forge/test/mlir/test_ops.py
```diff
@@ -583,7 +583,45 @@ def forward(self, x, y):
...
     co_out = [co.to("cpu") for co in co_out]
     fw_out = [fw_out] if isinstance(fw_out, torch.Tensor) else fw_out
-    assert all([compare_with_golden_pcc(golden=fo, calculated=co, pcc=0.99) for fo, co in zip(fw_out, co_out)])
+    assert all([compare_with_golden(golden=fo, calculated=co) for fo, co in zip(fw_out, co_out)])
+
+
+@pytest.mark.parametrize(
+    "shape",
+    [
+        (1, 128, 28, 28),
```
nvukobratTT, line 592: Can we test out different ranks as well? If some of them are not working, let's mark them as xfail.

forge/test/mlir/test_ops.py
```diff
@@ -583,7 +583,45 @@ def forward(self, x, y):
...
+        (1, 128, 14, 14),
+        (1, 128, 56, 56),
+        (1, 32, 64, 64),
+        (1, 512, 7, 7),
+        (1, 32, 32, 32),
+    ],
+)
+@pytest.mark.push
+def test_equal(shape):
+    class equal(nn.Module):
```
nvukobratTT, line 604: Can we use the [CapWords](https://peps.python.org/pep-0008/#class-names) convention for naming classes? 

I see this is not the case for all tests, but should be

forge/test/mlir/test_ops.py
```diff
@@ -574,7 +574,7 @@ def forward(self, x, y):
     x = torch.rand(shape_x)
     y = torch.rand(shape_y)
 
-    framework_model = not_equal()
+    framework_model = NotEqual()
```
nvukobratTT, line 577: ❤️

nvukobratTT (approved):

Issue #703, opened 2024-11-14
mnist test failure


gfengTT:
The test `forge/test/mlir/mnist/test_inference.py` failed:

```
E   AttributeError: module 'cv2.dnn' has no attribute 'DictValue'
```

Restoring the change to opencv version in the following commit fixed it. 
https://github.com/tenstorrent/tt-forge-fe/commit/ec8ee2c558240c1e5c4938e64c8d08750033a306#diff-6429007449318f592b43cd2c86b2139e98bad52a3712d67581f43a2fa07bea3d

PR #666, opened 2024-11-11
Configure missings tests in pytest collect

ashokkumarkannan1:
Fix https://github.com/tenstorrent/tt-forge-fe/issues/659

### Logs

1. Push: collect only logs - [push_collect_only.log](https://github.com/user-attachments/files/17699792/push_collect_only.log)
2. Push: tests run log - [push_tests.log](https://github.com/user-attachments/files/17699793/push_tests.log)
3. Nightly: collect only logs - [nightly_collect_only.log](https://github.com/user-attachments/files/17699791/nightly_collect_only.log)

CI failures in `reduce_sum` and `reduce_max` not replicated in latest main

1. [test_reduce_max.log](https://github.com/user-attachments/files/17699797/test_reduce_max.log)
2. [test_reduce_sum.log](https://github.com/user-attachments/files/17699798/test_reduce_sum.log)

PR #706, opened 2024-11-14
Revert OpenCV version change

ashokkumarkannan1:
Fix: https://github.com/tenstorrent/tt-forge-fe/issues/703

Previously CI was blocked with `ImportError: ERROR: recursion is detected during loading of "cv2" binary extensions. Check OpenCV installation.` issue. As per the suggestion [here](https://github.com/opencv/opencv-python/issues/680), downgraded the `opencv-python-headless` version from `4.6.0.66` to `4.5.5.64` in this [PR](https://github.com/tenstorrent/tt-forge-fe/pull/666).

Now this was blocking the MNIST Model with this [issue](https://github.com/tenstorrent/tt-forge-fe/issues/703). Hence reverting this change to see whether it still affect the CI.

### DIFF
--- a/env/core_requirements.txt
+++ b/env/core_requirements.txt
@@ -29,7 +29,7 @@ networkx==2.8.5
 numpy==1.23.1
 onnx>=1.15.0
 onnxruntime>=1.16.3
-opencv-python-headless==4.5.5.64
+opencv-python-headless==4.10.0.84
 # This is needed to avoid issue https://yyz-gitlab.local.tenstorrent.com/devops/devops/-/issues/95
 pandas==1.5.3
 prettytable==3.0.0


dgolubovicTT (approved):

nvukobratTT:
@gfengTT @ashokkumarkannan1 can we try to use some newer version of OpenCV that doesn't hit this issue? 

This more seems like bug in one of their version, so let's just test out newer one and see will that fix the issue, rather than downgrading it back again :))

_Note: [Here](https://github.com/facebookresearch/nougat/issues/40) is an example of a similar failure. I'm not sure if it's the same issues, but let's try to keep our libraries on newer versions, rather than downgrading them :))_

nvukobratTT (changes requested):
@gfengTT @ashokkumarkannan1 can we try to use some newer version of OpenCV that doesn't hit this issue? 

This more seems like bug in one of their version, so let's just test out newer one and see will that fix the issue, rather than downgrading it back again :))

_Note: [Here](https://github.com/facebookresearch/nougat/issues/40) is an example of a similar failure. I'm not sure if it's the same issues, but let's try to keep our libraries on newer versions, rather than downgrading them._

ashokkumarkannan1:
> @gfengTT @ashokkumarkannan1 can we try to use some newer version of OpenCV that doesn't hit this issue?
> 
> This more seems like bug in one of their version, so let's just test out newer one and see will that fix the issue, rather than downgrading it back again :))
> 
> _Note: [Here](https://github.com/facebookresearch/nougat/issues/40) is an example of a similar failure. I'm not sure if it's the same issues, but let's try to keep our libraries on newer versions, rather than downgrading them._


I agree with your point @nvukobratTT . From this [Release History](https://pypi.org/project/opencv-python-headless/#history) page, it seems `4.10.0.84` is the latest stable version. I will update this change here in this PR to verify is there any CI failures here. 

@gfengTT  Can you please confirm from your side whether `4.10.0.84` version works from your side?

gfengTT:
Great suggestion @nvukobratTT . 

@ashokkumarkannan1 Confirming `4.10.0.84` works.

nvukobratTT:
@ashokkumarkannan1 make sure to update PR name to reflect new changes :))

nvukobratTT (approved):

Issue #628, opened 2024-11-05
AssertionError: Setting a tensor value of incorrect shape: (1, 64, 55, 55) vs torch.Size([1, 64, 56, 56]) in maxpool2d in googlenet


kamalrajkannan78:
## Description

-  maxpool2d in googlenet faces AssertionError: Setting a tensor value of incorrect shape: (1, 64, 55, 55) vs torch.Size([1, 64, 56, 56]) 

## Reproduce

```
git checkout kkannan/kkannan/maxpool2d_issue
git submodule update --recursive
# To run model specific sanity
pytest  forge/test/model_demos/high_prio/cnn/pytorch/test_googlenet.py -svv  
# To run op specific(maxpool2d) sanity
pytest forge/test/mlir/test_ops.py -k "maxpool2d" -svv
```

## Observed Behaviour
> AssertionError: Setting a tensor value of incorrect shape: (1, 64, 55, 55) vs torch.Size([1, 64, 56, 56])

## Logs

- [googlenet_whole_model_issue.log](https://github.com/user-attachments/files/17637717/xxx_nov5_googlenet_debug.log)

- [model_specific_sanity.log](https://github.com/user-attachments/files/17637719/xxx_nov5_googlenet_debug_2.log)

- [op_specific_sanity.log](https://github.com/user-attachments/files/17637723/xxx_nov6_maxpool_s.log)

PR #718, opened 2024-11-15
fix shape mismatch issue in maxpool2d when ceil model is true

kamalrajkannan78:
## Summary

- This PR fixes [#628](https://github.com/tenstorrent/tt-forge-fe/issues/628) 
- The shape mismatch issue arises due to the way the shape method is computing the output height (h_out) and width (w_out) in maxpool2d.Earlier it used torch.floor even when ceil mode = true which causes rounding down the result instead of rounding up which leads to shape mismatch.
- Now based on ceil mode (false/true) , operation (torch.floor/torch.ceil) will be determined .
- Compile depth of the corresonding blocked model are updated.
- Tracks these [TVM changes](https://github.com/tenstorrent/tt-tvm/compare/be49f296bf30220584469cc4715ef9c97ee19b17...b332bb7031c17a81d2629683f6b7f0cec8deabf9#diff-afd1b9179c4d052038fb665278188afce38f9a0b290efcfb20ef9d0598c8f39e) and sanity tests for maxpool2d (with ceil_mode=True) and power=0.75 are added 

- Note : model and input data type are changed to bfloat16 in maxpool2d sanity to avoid  `ttnn.max_pool2d currently only supports an input type of bfloat16. Recieved 'f32'.` issue.

# Logs

- [googlenet_after_fix.log](https://github.com/user-attachments/files/17776887/nov15_z_googlenet_after_fix.log)

- [googlenet_before_fix.log](https://github.com/user-attachments/files/17776893/nov15_z_googlenet_before_fix.log)

- [maxpool2d_before_fix.log](https://github.com/user-attachments/files/17811369/nov19_maxpool2d_c4.log)

- [maxpool2d_after_fix.log](https://github.com/user-attachments/files/17813137/nov19_maxpool2d_8.log)

- [vovnet_after_fix.log](https://github.com/user-attachments/files/17776897/nov15_z_vovnet_after_fix.log)

- [vovnet_after_fix.log.zip](https://github.com/user-attachments/files/17777450/nov15_z_vovnet_after_fix.log.zip)

### DIFF
--- a/forge/test/mlir/test_ops.py
+++ b/forge/test/mlir/test_ops.py
@@ -15,6 +15,170 @@ from forge.tensor import to_forge_tensors, to_pt_tensors
 
 
 @pytest.mark.parametrize(
+    "input_shape, kernel_size, stride_size, padding, ceil_mode",
+    [
+        pytest.param(
+            (1, 96, 54, 54),
+            3,
+            2,
+            0,
+            True,
+            marks=pytest.mark.xfail(
+                reason="Runtime Error  : Invalid sharding configuration: For Row Major layout with element size of 2 bytes, the innermost dimension must align to 2 bytes."
+            ),
+        ),
+        pytest.param(
+            (1, 64, 55, 54),
+            3,
+            2,
+            0,
+            True,
+            marks=pytest.mark.xfail(
+                reason="Runtime Error  : Invalid sharding configuration: For Row Major layout with element size of 2 bytes, the innermost dimension must align to 2 bytes."
+            ),
+        ),
+        pytest.param(
+            (1, 128, 26, 26),
+            3,
+            2,
+            0,
+            True,
+            marks=pytest.mark.xfail(
+                reason="Runtime Error  : Invalid sharding configuration: For Row Major layout with element size of 2 bytes, the innermost dimension must align to 2 bytes."
+            ),
+        ),
+        pytest.param(
+            (1, 256, 26, 26),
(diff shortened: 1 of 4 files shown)


forge/test/mlir/test_ops.py
```diff
@@ -14,6 +14,74 @@
...
+        (1, 1, 32, 32),
+        (1, 128, 26, 26),
+        (1, 256, 26, 26),
+    ],
+)
+@pytest.mark.xfail(
+    reason="error: 'ttnn.max_pool2d' op ttnn.max_pool2d currently only supports an input type of bfloat16. Recieved 'f32'."
+)
+@pytest.mark.push
+def test_maxpool2d_with_ceil_mode(input_shape):
```
nvukobratTT, line 31: Can we have a single maxpool 2d test and keep cail as pytest parameters?

kamalrajkannan78: are you referring to combine this [maxpool2d](https://github.com/tenstorrent/tt-forge-fe/blob/13d2afcc7a6f28fef74623b29803845b87a028f9/forge/test/mlir/test_ops.py#L1616) with my test?

nvukobratTT: Ideally, it should be single test_max_pool_2d, and rest should be parameters if possible :))

Kamalrajkannan: Modified !

PR #719, opened 2024-11-15
Remove transpose cast to float32 and update test_transpose

### LINKED ISSUE #690: Casting transpose to bfloat16 still gives float32 on output
Repro:

checkout: dgolubovic/repro-tensor-mismatch-due-to-bfloat16

Run: pytest -svv forge/test/mlir/test_ops.py::test_transpose[params0]

ERROR    | forge.op.eval.common:compare_with_golden_pcc:245 - Tensor mismatch

What we get is tensor missmatch between framework output and compiled model. What is strange that initial graph of compiler has output df float32 even though we casted it to bfloat16:

Image

TODO: Investigate the cause of missmatch and why output dataformat is float32...

This is not currently blocker to anything but it may increase priority if we want to transfer all models to compile at bfloat16 dataformat...

dgolubovicTT:
Closes #690.
Changes in this PR:

1. Remove transpose output cast to float32 and transpose op argument data_format. Transpose shouldn't have data_format attribute! It should give the same data_format on output as it gets on input.
2. Update test_transpose to vary between torch.float32 and torch.bfloat16 inputs. 
3. Caught tensor mismatch for transpose when using torch.bfloat16 inputs, and marked as xfail untill resolved on metal side.

Effort to test transpose on bfloat16 inputs emerges from the need to run Llama 3B on bfloat16 since embedding only has support for bfloat16.

### DIFF
--- a/forge/test/mlir/test_ops.py
+++ b/forge/test/mlir/test_ops.py
@@ -693,22 +693,47 @@ def test_add():
     assert all([compare_with_golden_pcc(golden=fo, calculated=co, pcc=0.99) for fo, co in zip(fw_out, co_out)])
 
 
-@pytest.mark.parametrize(
-    "params",
-    [
-        ((1, 32, 64), (-1, -2)),
-        ((1, 64, 32), (1, 2)),
-        ((1, 32, 64, 128), (3, 2)),
-        ((32, 128), (0, 1)),
-        ((18, 65), (1, 0)),
-        ((6, 33, 34), (-1, 1)),
-        ((1, 32, 64), (-2, -3)),
-        ((6, 33, 34), (-1, -3)),
-        ((32, 128, 24), (1, -3)),
-    ],
-)
+params = [
+    ((1, 32, 64), (-1, -2)),
+    ((1, 64, 32), (1, 2)),
+    ((1, 32, 64, 128), (3, 2)),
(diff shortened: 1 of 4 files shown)


nvukobratTT (commented):
Seems like you have some unexpected changes on this PR. Can you double check?

dgolubovicTT:
Wow, yeah... Addressing

forge/test/mlir/test_ops.py
```diff
@@ -718,16 +743,13 @@ def forward(self, a):
             return torch.transpose(a, *self.dims)
 
     input_shape, dims = params
-    inputs = [torch.rand(input_shape)]
-
-    framework_model = Transpose(dims)
+    inputs = [torch.rand(input_shape, dtype=data_format)]  # Use data_format instead of hardcoded dtype
+    # Initialize the model with data_formats
+    framework_model = Transpose(dims).to(data_format)
```
nvukobratTT, line 748: Do we need to push the model to the desired DF if is already applied to input?

nvukobratTT (approved):

nvukobratTT (changes requested):
Accidently approved.. Seems like push CI is failing. Can you check if there are few more places where transpose attribute needs to be updated?

PR #736, opened 2024-11-20
uplift tt-mlir @84341179

pilkicTT:


nvukobratTT (approved):

pilkicTT:
From today, tests marked as `xfail` fail when they pass... so removing the mark from the tests which are passing with latest `tt-mlir`

Issue #639, opened 2024-11-07
Unified verification logic for ops and models


nvukobratTT:
## Summary

We want to have a generic way of verifying all op and model tests. Main point is to focus on generalizing how we're processing CPU and compiled model results, and how we're comparing those. 

## Details
At the moment, most of the verification is done this way:
```py
    co_out = [co.to("cpu") for co in co_out]
    fw_out = [fw_out] if isinstance(fw_out, torch.Tensor) else fw_out
    assert all([compare_with_golden_pcc(golden=fo, calculated=co, pcc=0.99) for fo, co in zip(fw_out, co_out)])
```

However, there are a few cons to doing verification this way:
-  There can be different cases where this approach isn't valid (e.g. scalar comparison)
- In most cases, we're not comparing everything properly (e.g. we didn't have `all` used previously) which can lead to invalid test cases
- It's not so readable to have everything defined here for each test (a bunch of redundant code)
- Etc.

In sum, we want to unify this logic and provide a good interface that will:

1. Make tests cleaner (single entry point for verification)
2. Have generalized logic that can be applied to different input cases (e.g. scalars, single tensor input, N tensors as input, etc.)
3. Be able to utilize verification configurations and properly define them if needed (e.g. usage of `VerifyConfig`)

Few notes:
- This interface should only be used for verification. No need to run compile through it. Let's be intentional about only using it for verification. 
The focus is on inference, but keep in mind that this verification might be expended for training scenarios as well. There is no need to develop logic for this case, be keep the code structure expandable for this case as well.
- It might be good to forward both the Original Module (e.g. PyTorch Model), and the Compiled Module as input to the verification interface. This way, we can have finer control if verification is done properly (e.g. user that defines the test isn't doing any modification before calling verify methods; squeezing the tensor, etc.)

PR #739, opened 2024-11-20
Unified verification logic

vkovinicTT:
This PR addresses issue [#639](https://github.com/tenstorrent/tt-forge-fe/issues/639) by introducing a new `verify` function that centralizes all verification logic in one place. Additionally, a new `VerifyConfig` class is introduced, allowing for custom verification settings per test. The previous configuration class has been renamed to `DeprecatedVerifyConfig` to indicate it is now obsolete.

For more details on the changes, refer to the [PR documentation](https://docs.google.com/document/d/1owE6RS5CoL2b6vmbwYtgfa53CjIGElIsKh3fq9axi-Y/edit?usp=sharing).

Fix [#639](https://github.com/tenstorrent/tt-forge-fe/issues/639)


forge/forge/verify/verify.py
```diff
@@ -254,7 +256,84 @@ def verify_golden(
...
+    verify_cfg: DepricatedVerifyConfig,
 ):
 
     assert False  # Run ttnn golden
+
+
+def verify(
+    inputs: List[Union[torch.Tensor, tf.Tensor, tf.Variable]],
+    framework_model: Union[torch.nn.Module, tf.Module, tf.keras.Model],
+    compiled_model: Any,
```
nvukobratTT, line 268: Is this one indeed "Any"?

forge/forge/verify/verify.py
```diff
@@ -254,7 +256,84 @@ def verify_golden(
...
+
+    co_out = [co.to("cpu") for co in co_out]
+    fw_out = [fw_out] if isinstance(fw_out, torch.Tensor) else fw_out
+
+    # 3rd step: verifications of outputs
+    # - dtype check
+    # - shape check
+    # - compare with golden
+    if verify_cfg.verify_size:
+        assert len(fw_out) == len(
```
nvukobratTT, line 321: Can we raise exceptions instead of doing asserts? Exceptions are a bit more descriptive, and easier to manage compared to assertions. 

This function has a mix of both. Do you have specific reason for choosing ether/or in this function?

forge/forge/verify/verify.py
```diff
@@ -254,7 +256,84 @@ def verify_golden(
...
+            co_out
+        ), f"Number of outputs from framework model and compiled model do not match: framework model has {len(fw_out)} outputs, compiled model has {len(co_out)} outputs"
+
+    if verify_cfg.verify_dtype:
+        dtype_mismatches = [(fw.dtype, co.dtype) for fw, co in zip(fw_out, co_out) if fw.dtype != co.dtype]
+        assert not dtype_mismatches, f"Dtype mismatches found: {dtype_mismatches}"
```
nvukobratTT, line 327: Can we outline which dtype is from where? E.g. Framework vs Compiled dtype. Let's have similar granularity for other errors as well

forge/forge/verify/verify.py
vkovinicTT: I've used `exceptions` for input errors (e.g. someone passing the model that we don't support). On the other hand, for errors caused by our compiler `assertions` are used. Should I raise exceptions for everything instead?

nvukobratTT: Behavior is similar, but I'm more in favor of using Exceptions. Main reasons:
1. You have more granularity by exception type (either existing or custom exceptions)
2. Better control over them (e.g. custom logic when hit)
3. We can catch them during the program execution and try to do recovery if possible
4. They can't be ignored with specific Python optimizations (-O)

forge/test/mlir/test_ops_tf.py
```diff
@@ -162,12 +163,9 @@ def call(self, x):
...
 
     framework_model = DualConv2d()
-    fw_out = to_pt_tensors(framework_model(*inputs))
-
     compiled_model = forge.compile(framework_model, sample_inputs=inputs)
-    co_out = compiled_model(*inputs)
-    co_out = [co.to("cpu").to(fw_out[0].dtype) for co in co_out]
-    assert compare_tensor_to_golden("dual_conv2d", fw_out[0], co_out[0].reshape(fw_out[0].shape))
+
+    verify(inputs=inputs, compiled_model=compiled_model, framework_model=framework_model)
```
nvukobratTT, line 168: Minor comment, do we need to run these as named attributes? Can we simplify it like this:
from 
```py
verify(inputs=inputs, compiled_model=compiled_model, framework_model=framework_model)
```
to:
```py
verify(inputs, compiled_model, framework_model)
```

It looks a bit more cleaner, as these attributes are required for every verification. Therefore, we don't need to explicit mention them like so. For other cases, if we expand verify with some "non-common" attributes, we can use named attributes to provide some additional arguments, while not specifying others by order :))

nvukobratTT (approved):
One minor comment. Once that's resolved, feel free to merge changes :)) 💪

forge/forge/verify/verify.py
```diff
@@ -254,7 +257,80 @@ def verify_golden(
...
 ):
 
     assert False  # Run ttnn golden
+
+
+def verify(
+    inputs: List[Union[torch.Tensor, tf.Tensor, tf.Variable]],
+    framework_model: Union[torch.nn.Module, tf.Module, tf.keras.Model],
+    compiled_model: CompiledModel,
+    verify_cfg: VerifyConfig = global_verify_config,
```
pilkicTT, line 270: Please, remove the `global_verify_config`.

There is no need for this you can just initialize with the default constructor. We should avoid any kind of global state, if possible...

forge/forge/verify/verify.py
```diff
@@ -254,7 +257,80 @@ def verify_golden(
...
+    # 1st step: run forward pass for the networks
+    fw_out = framework_model(*inputs)
+    co_out = compiled_model(*inputs)
+
+    # 2nd step: apply preprocessing (push tensors to cpu, perform any reshape if necessary,
+    #  cast from tensorflow tensors to pytorch tensors if needed)
+    if not isinstance(fw_out, torch.Tensor):
+        fw_out = to_pt_tensors(fw_out)
+
+    if not isinstance(co_out, torch.Tensor):
```
pilkicTT, line 311: This will never be true, since we always return a list of tensors when running the compiled model.

pilkicTT (changes requested):

pilkicTT (changes requested):
In general, I think it's fine to merge this, but I have couple suggestions. Check out the suggestions and then let's sync offline, to see if we are going to merge this right away, and then follow up with additional improvements (to avoid resolving merge conflicts).

pilkicTT (approved):
As we agreed offline, let's follow up on the suggested changes after this initial change is merged.

PR #747, opened 2024-11-22
Nightly_sweeps pytest mark for nightly CI run

vobojevicTT:


pytest.ini
```diff
@@ -11,6 +11,9 @@ xfail_strict=true
 markers =
     push: marks tests as push
     nightly: marks tests as nightly
+    nightly_sweeps: marks tests as nightly_sweeps
+    slow: marks tests as slow               # Slow tests, should not be run in push pipeline
+    run_in_pp: marks tests as run_in_pp     # Tests that should run in push pipeline
```
vmilosevic, line 16: How do run_in_pp test relate to push tests?
How are nightly_sweeps related to nightly tests? Is nightly_sweeps subset of nightly?
Can you share few examples on how we should select tests for CI runs?

For example
pytest -m run_in_pp  # Run this group on every pull request as gate check
pytest -m nightly_sweeps  # Run this group every night 
...

nvukobratTT (approved):

pytest.ini
nvukobratTT: Sunced offline, for now, nightly tests (modela) and nightly sweeps will run as ywo disjoint nightlies.

vobojevicTT: `slow` and `run_in_pp` are deprecated marks - added to the comment.
`pytest -m nightly_sweeps` - run all single op tests.

PR #770, opened 2024-11-26
Add new unary operators

vobojevicTT:
- clip # alias for clamp - as not implemented
 - clamp as not implemented
 - pow as implemented


vbrkicTT (approved):

forge/test/operators/pytorch/eltwise_unary/test_unary.py
```diff
@@ -160,6 +165,31 @@ class TestParamsData:
...
+
+    kwargs_clamp = [
+        {"min": 0.0, "max": 0.5},
+        {"min": 0.5, "max": 0.0},
+        {"min": 0.2},
+        {"max": 0.2},
+    ]
+
+    kwargs_pow = [
+        {"exponent": 0.5},
```
dgolubovicTT, line 180: do we test negative exponents?

vobojevicTT: Currently we don't. Should we include negative exponents?

dgolubovicTT: I would, that is the half of the domain of legal exponents.

dgolubovicTT: Also I would test non integer exponents. torch.pow supports non integer exponents. If we don't support them, we should be aware. I would test something like 1.5, 1.2...

vobojevicTT: Added mentioned exponent values.

forge/test/operators/pytorch/eltwise_unary/test_unary.py
```diff
@@ -160,6 +165,35 @@ class TestParamsData:
...
+    ]
+
+    kwargs_pow = [
+        {"exponent": 0.0},
+        {"exponent": 0.5},
+        {"exponent": 2.0},
+        {"exponent": 10.0},
+        {"exponent": -2.0},
+        {"exponent": -1.26},
+        {"exponent": 1.52},
```
vbrkicTT, line 186: To limit total number of tests, full list of kwargs should be combined with subset of input shapes. We probably need introduction of an another TestCllection and separation of lists of kwargs.
Subset of input shapes for the value testing should be considered. Input shapes from quick collection is maybe not sufficient.
Otoh full list of kwargs should be tested with all data formats.

vbrkicTT (approved):

dgolubovicTT (approved):

forge/test/operators/pytorch/eltwise_unary/test_unary.py
vobojevicTT: Added required changes.

PR #757, opened 2024-11-23
Uplift MLIR

nvukobratTT:

PR #781, opened 2024-11-27
Enable conv2d test case and clip tests

nvukobratTT:


forge/test/mlir/test_ops.py
```diff
@@ -497,7 +497,6 @@ def test_gelu(shape):
         ((8,), -3.0, -1.0),
     ],
 )
-@pytest.mark.xfail(reason="'ttnn.clamp' op input and output must have same shape")
```
dgolubovicTT, line 500: Is this a ttnn constraint that is planned to be addressed? If yes, is this test case that xfails part of some model?

dgolubovicTT (approved):

forge/test/mlir/test_ops.py
ashokkumarkannan1: We fixed this issue in tt-mlir - https://github.com/tenstorrent/tt-mlir/pull/1357

As we uplifted tt-mlir recently in this [PR](https://github.com/tenstorrent/tt-forge-fe/pull/757), this issue is fixed and tests are passing now.

Issue #784, opened 2024-11-27
Fix cross entropy loss


pilkicTT:
The cross entropy implementation is wrong. Reimplement it and add tests.

PR #785, opened 2024-11-27
[ops] lower log op

pilkicTT:
Lowering + basic op tests.

Needed for cross entropy loss. Issue #784


forge/test/mlir/test_ops.py
```diff
@@ -701,6 +701,40 @@ def forward(self, x):
     assert all([compare_with_golden_pcc(golden=fo, calculated=co, pcc=0.99) for fo, co in zip(fw_out, co_out)])
 
 
+@pytest.mark.parametrize(
```
nvukobratTT, line 704: Can we test different ranks as well?

nvukobratTT (approved):

forge/test/mlir/test_ops.py
pilkicTT: Done...

PR #786, opened 2024-11-27
convert const scalars into tensors

pilkicTT:
Currently, neither `tt-mlir` or our runtime can handle scalars properly. So, when creating constant scalars, create them as a 1d tensors of shape (1,).

Pre-requisite for #784

### DIFF
--- a/forge/forge/op/eval/common.py
+++ b/forge/forge/op/eval/common.py
@@ -87,7 +87,9 @@ def create_constant_tensor_from_value(
         tensor[:, :, 0:dim_r, 0:dim_c] = value
     else:
         if tensor_c == 0 and tensor_r == 0:
-            tensor = torch.tensor(value, dtype=dtype)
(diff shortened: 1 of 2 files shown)


forge/forge/op/eval/common.py
```diff
@@ -87,7 +87,9 @@ def create_constant_tensor_from_value(
         tensor[:, :, 0:dim_r, 0:dim_c] = value
     else:
         if tensor_c == 0 and tensor_r == 0:
-            tensor = torch.tensor(value, dtype=dtype)
+            # Unsqueezing to make this a 1d tensor.
+            # Currently, the runtime expects a 1d tensor for scalar values.
```
nvukobratTT, line 91: Should this be by design for the runtime? If no, we should track this as separate issues on MLIR side

nvukobratTT (approved):

PR #787, opened 2024-11-27
[train] fix implementation of cross entropy loss

### LINKED ISSUE #784: Fix cross entropy loss

pilkicTT:
Rewrite the cross entropy loss, since the existing implementation is more complex and wrong.

Fixes #784

### DIFF
--- a/forge/forge/op/loss.py
+++ b/forge/forge/op/loss.py
@@ -2,41 +2,41 @@
 
 # SPDX-License-Identifier: Apache-2.0
 from ..module import ForgeModule
-from .reduce import ReduceSum, ReduceAvg
-from .eltwise_unary import Exp, Log, Abs
-from .eltwise_binary import Subtract, Multiply
 from .constant import Constant
+from .eltwise_unary import Log, Abs
+from .eltwise_binary import Subtract, Multiply
+from .nn import Softmax
+from .reduce import ReduceSum, ReduceAvg
 
 
 class CrossEntropyLoss(ForgeModule):
     """
-    Cross-Entropy Loss
+    Cross-Entropy Loss (with mean reduction)
+
+    loss = reduce_avg(-1 * sum(labels * log(softmax(predictions)), dim=-1), dim=0)
     """
 
     def forward(self, predictions, labels):
-        exponential_predicted = Exp("exp", predictions)
-        reduced_exponential_predicted = ReduceSum("reduce_sum", exponential_predicted, dim=-1)
-        ln_reduced_exponential_predicted = Log("log", reduced_exponential_predicted)
-        output_minus_ln_reduced_exponential_predicted = Subtract(
-            "subtract", predictions, ln_reduced_exponential_predicted
-        )
+        assert predictions.ndim() == 2, f"Predictions must be a 2D tensor. Got {predictions.ndim()}."
+        assert (
+            predictions.shape == labels.shape
+        ), f"Shapes of predictions and labels must match. predictions.shape={predictions.shape} labels.shape={labels.shape}."
+
+        softmax = Softmax("softmax", predictions, dim=-1)
+        log_softmax = Log("log", softmax)
+
+        product = Multiply("products", labels, log_softmax)
+        log_loss = ReduceSum("log_loss", product, dim=-1)
+
         negative_one_constant = Constant("negative_one_const", constant=-1.0)
-        negative_one_times_output_minus_ln_reduced_exponential_predicted = Multiply(
-            "multiply",
-            output_minus_ln_reduced_exponential_predicted,
+        negative_log_loss = Multiply(
+            "negative_log_loss",
+            log_loss,
             negative_one_constant,
         )
-        word_losses = Multiply(
-            "word_losses",
-            negative_one_times_output_minus_ln_reduced_exponential_predicted,
-            labels,
-        )
-        token_losses = ReduceSum("token_losses", word_losses, dim=-1)
-        avg_sequence_loss = ReduceAvg("avg_sequence_loss", token_losses, dim=-2)
 
-        out = avg_sequence_loss
-
-        return out
+        reduction_avg = ReduceAvg("reduction_avg", negative_log_loss, dim=0)
+        return reduction_avg
 
 
 class L1Loss(ForgeModule):
--- a/forge/forge/tensor.py
+++ b/forge/forge/tensor.py
@@ -115,6 +115,9 @@ class TensorShape:
     def __getitem__(self, i):
         return self.dims[i]
 
+    def __eq__(self, other):
+        return self.dims == other.dims
+
 
 class TensorBase:
     def _create_const_tensor(self, value):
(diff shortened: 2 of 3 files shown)


forge/forge/op/loss.py
```diff
@@ -2,41 +2,36 @@
...
 
 class CrossEntropyLoss(ForgeModule):
     """
-    Cross-Entropy Loss
+    Cross-Entropy Loss (with mean reduction)
+
+    loss = reduce_avg(-1 * sum(labels * log(softmax(predictions)), dim=-1), dim=0)
     """
 
     def forward(self, predictions, labels):
```
ndrakulicTT, line 19: Maybe add assert that predictions and labels needs to have same shape so that in case labels are in ints with shape (batch_size) model does not compile, as this is possible in torch.CrossEntropyLoss

vladimirjovanovicTT: I would add to assert that predictions and labels need to be 2D, since we are covering this case here, and torch covers more stuff, e.g. 'shape' section here:

https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html

forge/test/mlir/test_loss.py
```diff
@@ -0,0 +1,66 @@
...
+    target_forge = forge.tensor.Tensor.create_from_torch(target)
+
+    forge_loss = forge.compile(forge_loss, sample_inputs=[prediction_forge, target_forge])
+    forge_loss_out = forge_loss(prediction, target)
+    torch_loss_out = torch_loss(prediction, target)
+
+    print(f"forge_loss_out: {forge_loss_out}")
+    print(torch_loss_out)
+
+    assert torch.allclose(torch_loss_out, forge_loss_out[0], rtol=11e-3)
```
vladimirjovanovicTT, line 66: 11e-3? :)

vladimirjovanovicTT (approved):
Looks good!

forge/test/mlir/test_loss.py
pilkicTT: exactly, 1e-2 is not enough... 😄

forge/forge/op/loss.py
```diff
@@ -2,41 +2,41 @@
...
-        )
+        assert predictions.ndim() == 2, f"Predictions must be a 2D tensor. Got {predictions.ndim()}."
+        assert (
+            predictions.shape == labels.shape
+        ), f"Shapes of predictions and labels must match. Got {predictions.shape} and {labels.shape}."
+
+        softmax = Softmax("softmax", predictions, dim=-1)
+        log_softmax = Log("log", softmax)
+
+        product = Multiply("products", labels, log_softmax)
```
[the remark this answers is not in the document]
pilkicTT: I am asserting that the shapes are the same, that should be enough?

forge/forge/op/loss.py
```diff
@@ -2,41 +2,41 @@
...
-        exponential_predicted = Exp("exp", predictions)
-        reduced_exponential_predicted = ReduceSum("reduce_sum", exponential_predicted, dim=-1)
-        ln_reduced_exponential_predicted = Log("log", reduced_exponential_predicted)
-        output_minus_ln_reduced_exponential_predicted = Subtract(
-            "subtract", predictions, ln_reduced_exponential_predicted
-        )
+        assert predictions.ndim() == 2, f"Predictions must be a 2D tensor. Got {predictions.ndim()}."
+        assert (
+            predictions.shape == labels.shape
+        ), f"Shapes of predictions and labels must match. Got {predictions.shape} and {labels.shape}."
```
nvukobratTT, line 23: Can we clarify the message?

More precisely, to explicitly say which shape is prediction, and which label

forge/forge/tensor.py
```diff
@@ -223,6 +226,12 @@ def __repr__(self):
             return f"Forge Tensor: {self.value()}, {self.data_format}"
         return f"Forge Empty Tensor: {self.shape}"
 
+    def ndim(self) -> int:
```
nvukobratTT, line 229: Should we just name this function `rank()`? 

E.g. `... Got { predictions.rank() } ...`

forge/test/mlir/test_loss.py
```diff
@@ -0,0 +1,63 @@
...
+    prediction_forge = forge.tensor.Tensor.create_from_torch(prediction)
+    target = torch.empty(prediction_shape[0], dtype=torch.long).random_(prediction_shape[1])
+    target = nn.functional.one_hot(target, num_classes=prediction_shape[1]).float()
+    target_forge = forge.tensor.Tensor.create_from_torch(target)
+
+    forge_loss = forge.compile(forge_loss, sample_inputs=[prediction_forge, target_forge])
+    forge_loss_out = forge_loss(prediction, target)
+    torch_loss_out = torch_loss(prediction, target)
+
+    assert torch.allclose(torch_loss_out, forge_loss_out[0], rtol=11e-3)
```
nvukobratTT, line 36: @vkovinicTT once you merge your PR, please checkout if there are new tests that are not incorporating your verification feature.

Also, once merged, make sure to post the details on #fore-fe channel, in order to communicate further for everyone to use this function for future tensor verifications.

forge/forge/tensor.py
pilkicTT: torch uses ndim, so i've used it also; i will rename it to rank

forge/forge/op/loss.py
```diff
@@ -2,41 +2,41 @@
...
+from .eltwise_unary import Log, Abs
+from .eltwise_binary import Subtract, Multiply
+from .nn import Softmax
+from .reduce import ReduceSum, ReduceAvg
 
 
 class CrossEntropyLoss(ForgeModule):
     """
-    Cross-Entropy Loss
+    Cross-Entropy Loss (with mean reduction)
```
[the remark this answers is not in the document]
pilkicTT: I'll file an issue.

For adding the `reduction` attribute, I think it will only create unnecessary confusion... The loss module doesn't take any arguments, so the user knows that there is no option of configuration reduction (or disabling it). If I add an argument with default value "mean", then the user will think that it can change it, but he will be greeted with an assert instead. So, I won't add this now; it should be added once proper reduction configuration is added.

forge/forge/tensor.py
vladimirjovanovicTT: Actually jax also uses ndim.
As I think about it, 'rank' is kind of an overloaded notation - if you ask for a rank of a 2d tensor, do you want to get the matrix rank or nr of dimensions?
I would stick with ndim. Thoughts? @pilkicTT @nvukobratTT

pilkicTT: I am fine either way 😄 

I also thought that `ndim` seems more aligned with rest of the world...

forge/forge/op/loss.py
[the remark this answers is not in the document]
pilkicTT: Created issue #792

forge/forge/tensor.py
pilkicTT: Leaving as `ndim`...

nvukobratTT (approved):

PR #788, opened 2024-11-27
[consteval] enable during training

### LINKED ISSUE #293: run_consteval_graph_pass is being called twice in forge_compile_from_context
Once in run_post_initial_graph_pass and once in run_consteval_pass.
We need to understand why it's being called in run_post_initial_graph_pass and find a better solution. If there are no reasons for it, we can simply remove it and leave it in run_consteval_pass.

pilkicTT:
Enables consteval pass when compiling for training, by disabling consteval on inputs which require gradients (when training is on).

Fixing up the const eval during runtime - removing the unnecessary (legacy) checks for epoch type.

Moving the consteval pass to be executed just before `pre_lowering` pass.

Closes #293

### DIFF
--- a/forge/csrc/graph_lib/utils.cpp
+++ b/forge/csrc/graph_lib/utils.cpp
@@ -2155,8 +2155,8 @@ void ConstEvalGraph::autograd()
 bool is_consteval_capable_input_type(Node *node)
 {
     graphlib::InputNode *input = dynamic_cast<graphlib::InputNode *>(node);
-    return input and (input->is_parameter() or input->is_constant()) and
-           not node->as<graphlib::TaggedNode>()->has_tag("dont_consteval");
+    return input && (input->is_parameter() || input->is_constant()) &&
+           !node->as<graphlib::TaggedNode>()->has_tag("dont_consteval");
 }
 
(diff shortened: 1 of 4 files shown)


pilkicTT:
Created an issue for further refactor of the `consteval` pass: #796

forge/csrc/graph_lib/utils.cpp
```diff
@@ -2245,6 +2245,11 @@ bool is_consteval_capable_input_no_operand_forks(Graph *graph, InputNode *input)
     std::vector<Node *> users = graph->data_users(input);
     std::vector<Edge> user_edges = graph->user_data_edges(input);
 
+    if (input->requires_grad() && graph->training())
```
nvukobratTT, line 2248: Do we need to check for both? 

What is the case where these can differ?

nvukobratTT (approved):

forge/csrc/graph_lib/utils.cpp
pilkicTT: Generally, `requires_grad` is set depending on how the framework module is defined. By default, all `torch.nn.Parameter`s in torch are marked as `requires_grad` - regardless of training or inference mode.

So, if I only check for `requires_grad` then I'll disable consteval for params during inference (which is not ideal).

PR #793, opened 2024-11-28
update compile depth for forge models

kamalrajkannan78:
## summary 

- update the compile depth of models based on the results on latest main


nvukobratTT:
Is this PR still active?

PR #738, opened 2024-11-20
Update named attrs in commute_utils

meenakshiramanathan1:
Fixes #704 
overwrite_op_attrs has been deprecated and replaced with op specific named attr functions.
overwrite_op_named_attrs has been created to update both attrs and named attrs , the specific attribute to be updated in named attrs will be passed through update function of corresponding op (reshape, concat, select, reduce, matmul etc.)

update_reshape_attr -> Updates shape attribute in attrs and named attrs
update_select_attr ->  Updates select_dim, begin, length, and stride attributes in attrs and named_attrs
update_reduce_attr -> Updates the reduce_dim attribute in attrs and named_attrs
update_concat_attr -> Updates the dim attribute in attrs and named_attrs
update_grouped_reduce_avg -> Updates reduce_dim and other attributes in attrs and named_attrs
update_conv_attr -> Updates padding attributes in attrs and named_attrs for a conv2d operation.
update_matmul_attr -> Updates the requant_zp attribute in attrs and named_attrs
update_vstack_attr -> Updates the slice_size attribute in attrs and named_attrs

PR #814, opened 2024-12-02
Add log type that is missing

dgolubovicTT:
Adding log type in logger_trace calls that was missing. This caused issues with build in debug mode. Introduced in [PR](https://github.com/tenstorrent/tt-forge-fe/pull/738)


forge/csrc/passes/commute_utils.cpp
```diff
@@ -3,6 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 #include "passes/commute_utils.hpp"
 
+#include <utils/logger.hpp>
```
pilkicTT, line 6: use quotes, i.e.:
```
#include "utils/logger.hpp"
```

pilkicTT (approved):

PR #819, opened 2024-12-03
Add Weekly job for model analysis

### LINKED ISSUE #662: [Model Analysis] Weekly job that will automatically update model and op details
Based on existing scripts, create a weekly job that will generate all markdowns and update them automatically. This one should be run over weekends.

For reference, we already have a sample of doing so, just for the FFE/MLIR uplifts. We can utilize sam structure to generate and run these.

chandrasekaranpradeep:
Fixes #662 
1. Add new marker `model_analysis` for the models inference test present in the forge/test/models/pytorch path
2. Modified the script to collect the test based upon the marker `model_analysis`
3. Created a workflow(i.e model-analysis-weekly.yml) for running the model analysis script  which will generate markdown files and then create a PR for updating the model_analysis_docs(Markdown files).
4. Moved the build-image job in Build and Test workflow(i.e build-and-test.yml) to separate workflow, so that I can reuse the workflow for the Model Analysis Weekly workflow(i.e model-analysis-weekly.yml) 

Note: The workflow job will trigger every week on Friday at 11:00 PM UTC (Saturday at 12:00 AM Serbia time).

### DIFF
--- a/.github/workflows/model-analysis-weekly.yml
+++ b/.github/workflows/model-analysis-weekly.yml
@@ -1,0 +1,112 @@
+name: Model Analysis Weekly
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 23 * * 5' # 11:00 PM UTC Friday (12:00 AM Saturday Serbia)
+
+jobs:
+
+  docker-build:
+    uses: ./.github/workflows/build-image.yml
+    secrets: inherit
+
+  model-analysis:
+    needs: docker-build
+    runs-on: runner
+
+    container:
+      image: ${{ needs.docker-build.outputs.docker-image }}
+      options: --device /dev/tenstorrent/0
+      volumes:
+        - /dev/hugepages:/dev/hugepages
+        - /dev/hugepages-1G:/dev/hugepages-1G
+        - /etc/udev/rules.d:/etc/udev/rules.d
+        - /lib/modules:/lib/modules
+        - /opt/tt_metal_infra/provisioning/provisioning_env:/opt/tt_metal_infra/provisioning/provisioning_env
+
+    steps:
+
+      - name: Set reusable strings
+        id: strings
+        shell: bash
+        run: |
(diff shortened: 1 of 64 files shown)


.github/workflows/model-analysis-weekly.yml
```diff
@@ -0,0 +1,111 @@
...
+          -DCMAKE_CXX_COMPILER=clang++ \
+          -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+          -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+          cmake --build ${{ steps.strings.outputs.build-output-dir }}
+
+      - name: Run Model Analysis Script
+        shell: bash
+        run: |
+          source env/activate
+          apt install -y libgl1-mesa-glx
```
milosevic81, line 75: You may want to set "set pipefail" here, otherwise it may report a false positive result

chandrasekaranpradeep: Sure Will update it

vmilosevic (approved):

.github/workflows/model-analysis-weekly.yml
chandrasekaranpradeep: Added the `set -o pipefail ` command to change the pipeline exit status to the first command that fail.

dgolubovicTT (approved):

forge/test/conftest.py
```diff
@@ -448,3 +448,23 @@ def pytest_runtest_logreport(report):
         for key, default_value in environ_before_test.items():
             if os.environ.get(key, "") != default_value:
                 os.environ[key] = default_value
+
+
+def pytest_collection_modifyitems(config, items):
+
+    marker = config.getoption("-m")  # Get the marker from the -m option
+
+    if marker and marker == "model_analysis":  # If a marker is specified
```
vmilosevic, line 457: Please check if this is correct, marker can also be an expression and "not model_analysis" or "push and nightly" are all valid -m options, how will 
```if marker and marker == "model_analysis```
work in that case?

PR #822, opened 2024-12-03
Remove one value from parameter 'number of operands' used in tests

kmilanovicTT:
Remove value '7' from test parameter 'number of operands' that cause problem with memory during nightly sweeps

Relates to Bug #867

### DIFF
--- a/forge/test/operators/pytorch/eltwise_nary/test_concatenate.py
+++ b/forge/test/operators/pytorch/eltwise_nary/test_concatenate.py
@@ -237,7 +237,7 @@ class TestCollectionData:
         numbers_of_operands=[
             2,
             3,
-            7,
+            # 7,  # consume too much memory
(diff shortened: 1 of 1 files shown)


vbrkicTT (approved):

nvukobratTT:
Is this PR still active?

vbrkicTT:
> Is this PR still active?

I would say it's not relevant anymore since process doesn't stop at Concatenate anymore.

Issue with the memory should be addressed in a generic way.  There is a bug #867 that document the issue.

nvukobratTT:
> > Is this PR still active?
> 
> I would say it's not relevant anymore since process doesn't stop at Concatenate anymore.
> 
> Issue with the memory should be addressed in a generic way. There is a bug #867 that document the issue.

Thanks! For now, let's close this PR as we already have a issues with relevant details to address to problem

Issue #816, opened 2024-12-02
update_reshape_attr called for a non-reshape operation in whisper and swin model


kamalrajkannan78:
repro steps will be added soon

Issue #772, opened 2024-11-26
'ttir.squeeze' op Dimensions of the output tensor must be the same as the input tensor except for dimension 0.


ashokkumarkannan1:
### Description:
During the bringup of the xglm model after the compilation we will be facing this issue in runtime of the compiled model.
### Log:
`loc("reshape_297.dc.squeeze.0_operand_commute_clone1810"("forward":4294967295:3440)): error: 'ttir.squeeze' op Dimensions of the output tensor must be the same as the input tensor except for dimension 0.`

[xglm_log.log](https://github.com/user-attachments/files/17917545/xglm_log.log)

PR #833, opened 2024-12-04
Fix: 'ttir.squeeze' op Dimensions must be same for input & output

ashokkumarkannan1:
Fix: https://github.com/tenstorrent/tt-forge-fe/issues/772, Fix https://github.com/tenstorrent/tt-forge-fe/issues/816

The issue arises from an incorrect incoming edge to the commute clone node. Previously, the node inherited the incoming edge from the old edge, which led to inconsistencies in shape and caused the recalculation of shape to fail. By updating the incoming edge after the node is attached, this inconsistency will be resolved.

# Before Commute and Bypass
Screenshot 2024-12-09 at 9 25 02 AM

# Before (without Fix)
Screenshot 2024-12-04 at 5 32 16 PM

# After (With Fix)
Screenshot 2024-12-04 at 5 32 53 PM


dgolubovicTT:
Can you please add a unit test in `forge/csrc/test/passes/test_erase_inverse_ops.cpp`?

forge/csrc/passes/erase_inverse_ops.cpp
```diff
@@ -306,37 +311,44 @@ void commute_and_bypass(graphlib::Graph *graph, std::vector<graphlib::Node *> co
...
-                dynamic_cast<graphlib::InputNode *>(graph->node_by_id(operand_edge.producer_node_id)))
+            if ((first->op_name() == "unsqueeze" or first->op_name() == "squeeze"))
             {
                 op->change_op_type("reshape");
-                graphlib::Shape op_shape = graphlib::Shape::create(std::vector<uint32_t>(consumer->shape().size(), 1));
-                auto input = dynamic_cast<graphlib::InputNode *>(graph->node_by_id(operand_edge.producer_node_id));
-
-                for (int i = -1; i >= -(int)input->shape().size(); i--)
+                std::vector<graphlib::Edge> clone_data_edges = graph->operand_data_edges(clone);
+                for (const auto &clone_data_edge : clone_data_edges)
```
dgolubovicTT, line 318: Let's not use auto if we can, instead use the actual type.

forge/csrc/passes/erase_inverse_ops.cpp
```diff
@@ -306,37 +311,44 @@ void commute_and_bypass(graphlib::Graph *graph, std::vector<graphlib::Node *> co
...
                 {
-                    if (i + (int)op_shape.size() >= 0)
-                        op_shape[i] = input->shape()[i];
-                    else
+                    auto input =
```
dgolubovicTT, line 320: same here...

forge/csrc/passes/erase_inverse_ops.cpp
```diff
@@ -270,6 +270,11 @@ void commute_and_bypass(graphlib::Graph *graph, std::vector<graphlib::Node *> co
                 name,
                 consumer->name(),
                 graph->node_by_id(operand_edge.producer_node_id)->name());
+
+            // Updating op type to reshape to support update_reshape_attr
+            if (op->op_name() == "squeeze" || op->op_name() == "unsqueeze")
```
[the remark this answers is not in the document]
dgolubovicTT: When I run this test you wrote `test_squeeze_operand_commute_clone` on main branch, reshape decomposes to squeeze/unsqueeze. Then, update_reshape_attr is called on suqeeze op which triggers this assert:

`TT_ASSERT(reshape->op_name() == "reshape", "update_reshape_attr called for a non-reshape operation");`

Is that why you are changing op type to reshape?

forge/csrc/passes/erase_inverse_ops.cpp
ashokkumarkannan1: Updated

forge/csrc/passes/erase_inverse_ops.cpp
ashokkumarkannan1: Updated

ashokkumarkannan1:
> Can you please add a unit test in `forge/csrc/test/passes/test_erase_inverse_ops.cpp`?

Done @dgolubovicTT .

Can you please review it?

forge/csrc/passes/erase_inverse_ops.cpp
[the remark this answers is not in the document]
ashokkumarkannan1: yes @dgolubovicTT

dgolubovicTT:
I am still looking into this... I want to make sure this is the right way to solve this one...

forge/csrc/passes/erase_inverse_ops.cpp
```diff
@@ -306,37 +311,44 @@ void commute_and_bypass(graphlib::Graph *graph, std::vector<graphlib::Node *> co
...
-                dynamic_cast<graphlib::InputNode *>(graph->node_by_id(operand_edge.producer_node_id)))
+            if ((first->op_name() == "unsqueeze" or first->op_name() == "squeeze"))
             {
                 op->change_op_type("reshape");
-                graphlib::Shape op_shape = graphlib::Shape::create(std::vector<uint32_t>(consumer->shape().size(), 1));
-                auto input = dynamic_cast<graphlib::InputNode *>(graph->node_by_id(operand_edge.producer_node_id));
-
-                for (int i = -1; i >= -(int)input->shape().size(); i--)
+                std::vector<graphlib::Edge> clone_data_edges = graph->operand_data_edges(clone);
+                for (const graphlib::Edge &clone_data_edge : clone_data_edges)
```
dgolubovicTT, line 318: Won't this clone_data_edges be empty since clone isn't added between any two nodes in the graph untill `insert_node_on_edge` call later? So will we ever enter this for loop?

dgolubovicTT:
@ashokkumarkannan1 this is my understanding of what is happening here:
We are trying to add squeeze op as an inverse to reshape ((16,256,256) -> (1,16,256,256)) on the operant of the current node we are commuting the reshape over.

The issue is that dimensions of the output tensor must be the same as the input tensor except for dimension we are squeezing.
In our case we have a broadcast on edge before squeeze. That broadcast isn't accounted for since MLIR doesn't have a representation for broadcast. Then, the only thing MLIR sees is the input of squeeze being (1,1,256,256) and output (16,256,256). 

Normally, this problem doesn't exist since broadcast is commuted through clone op (in our case squeeze). However, `try_commute_bcast_through_clone` only supports reshape and transpose, and fails silently for squeeze. 

Your solution is hacking this by changing clone op_type to reshape just to make commuting broadcast possible: In [this line](https://github.com/tenstorrent/tt-forge-fe/pull/833/files?diff=unified&w=0#diff-d72de50643cecb66351311940149d4fdc3ea69e286f2d5742d281bd0d26de97eR314) by removing the condition op is again changed op type to reshape `op->change_op_type("reshape");` and the whole [for loop](https://github.com/tenstorrent/tt-forge-fe/pull/833/files?diff=unified&w=0#diff-d72de50643cecb66351311940149d4fdc3ea69e286f2d5742d281bd0d26de97eR318) seems to be skipped.

This hopping between op types is wrong IMO. The logical thing is to add commute through squeeze and unsqueeze, but I would have to think a little more. 

Maybe I am missing some pieces. @ashokkumarkannan1 please give your thoughts on this so we can see the whole picture

ashokkumarkannan1:
> @ashokkumarkannan1 this is my understanding of what is happening here: We are trying to add squeeze op as an inverse to reshape ((16,256,256) -> (1,16,256,256)) on the operant of the current node we are commuting the reshape over.
> 
> The issue is that dimensions of the output tensor must be the same as the input tensor except for dimension we are squeezing. In our case we have a broadcast on edge before squeeze. That broadcast isn't accounted for since MLIR doesn't have a representation for broadcast. Then, the only thing MLIR sees is the input of squeeze being (1,1,256,256) and output (16,256,256).
> 
> Normally, this problem doesn't exist since broadcast is commuted through clone op (in our case squeeze). However, `try_commute_bcast_through_clone` only supports reshape and transpose, and fails silently for squeeze.
> 
> Your solution is hacking this by changing clone op_type to reshape just to make commuting broadcast possible: In [this line](https://github.com/tenstorrent/tt-forge-fe/pull/833/files?diff=unified&w=0#diff-d72de50643cecb66351311940149d4fdc3ea69e286f2d5742d281bd0d26de97eR314) by removing the condition op is again changed op type to reshape `op->change_op_type("reshape");` and the whole [for loop](https://github.com/tenstorrent/tt-forge-fe/pull/833/files?diff=unified&w=0#diff-d72de50643cecb66351311940149d4fdc3ea69e286f2d5742d281bd0d26de97eR318) seems to be skipped.
> 
> This hopping between op types is wrong IMO. The logical thing is to add commute through squeeze and unsqueeze, but I would have to think a little more.
> 
> Maybe I am missing some pieces. @ashokkumarkannan1 please give your thoughts on this so we can see the whole picture

Hi @dgolubovicTT ,

As you mentioned, due to broadcast TM and it is not reflected in MLIR, the output shape of reshape is wrong. ([1, 1, 256, 256] cannot be reshaped to [16, 256, 256]). So I made the shape calculation from previous node of this squeeze op. I too suspected there is a hidden bug here but not sure what it is. If you have any pointers, please let me know

forge/csrc/passes/erase_inverse_ops.cpp
dgolubovicTT: Any thoughts on this? @ashokkumarkannan1

ashokkumarkannan1: @dgolubovicTT   I'm checking on this. Now I moved insert_node_on_edge before this block and now I can get the proper data edges and its not empty. If you have any docs on this commute_and_bypass, please share it with me.

dgolubovicTT: I don't have any additional docs, just reading the code...

ashokkumarkannan1:
@dgolubovicTT I have updated the code now. can you review it ?

nvukobratTT:
Hey guys, what is the status of this PR?

ashokkumarkannan1:
> Hey guys, what is the status of this PR?

@nvukobratTT  the logic of the erase_inverese_ops for squeeze and unsqueeze ops in commute_clone is rectified and updated .

forge/csrc/passes/erase_inverse_ops.cpp
```diff
@@ -336,12 +339,25 @@ void commute_and_bypass(graphlib::Graph *graph, std::vector<graphlib::Node *> co
...
+                if (op->op_name() == "reshape")
+                    update_reshape_attr(op, op_shape);
             }
 
             auto [in_edge, out_edge] = insert_node_on_edge(graph, operand_edge, clone);
             // Set dataformat to match producer on operand edge
             clone->set_output_df(graph->node_by_id(in_edge.producer_node_id)->output_df());
+
+            // Remove braodcast tms from clone's data edges and recalculate shapes
+            std::vector<graphlib::Edge> clone_data_edges = graph->operand_data_edges(clone);
```
nvukobratTT, line 351: Why are we removing broadcast TMs here?

ashokkumarkannan1: Hi @nvukobratTT ,
Broadcast tms are removed only for edges which has consumer as squeeze or unsqueeze ops. If we have broadcast tms for theses edges, op shape is miscalculated, which is the root cause for this issue at MLIR stage.

As adding broadcast tms for squeeze & unsqueeze op is not valid, this is removed now.

nvukobratTT: Do you know when those broadcasts are added? We got them from somewhere, and my suggestion is to determine where, and fix this issue there, instead of bypassing the root problem this way.

Also, if this proves to be harder, let's just disable this pass for referencing models, and see what the will be results. If we can't figure out root of the broadcast, disabling this optimization is okay for time being.

dgolubovicTT:
Closing this one since new [PR](https://github.com/tenstorrent/tt-forge-fe/pull/1082) is solving the issues #772 and #816

PR #835, opened 2024-12-04
Convert flatbuffer binary to json

meenakshiramanathan1:


dgolubovicTT (approved):

pilkicTT (changes requested):
Why is this needed? We are already saving the generated TTNN IR with other reports in default reportify directory...

meenakshiramanathan1:
> Why is this needed? We are already saving the generated TTNN IR with other reports in default reportify directory...

This was added as a way to dump the generated binary and json so that it can be used to run generate_md script present in [this](https://github.com/tenstorrent/tt-torch/tree/dgolubovic/check-parsing-form-to-json) branch as an experiment.

CC: @dgolubovicTT

dgolubovicTT:
This change was made in order to generate json files for unique ops so that we could generate ttnn configurations for the ttnn folks. @tt-mpantic contacted me to check if this json generation works as input to [generate_md.py ](https://github.com/tenstorrent/tt-torch/blob/main/tt_torch/tools/generate_md.py) script for generating ttnn configurations. I verified that it had worked and generated ttnn configurations Mladen needed. 

If we are going to provide those configurations to ttnn folks, we should have this in main... Question for @tt-mpantic: Is this recurring and can you give more details on that ttnn effort? Are they using these for debugging, planning or something else?

nvukobratTT:
> recurring and can you give more details on that ttnn effort? Are they using these for debugging, planning or something

Thanks for the context! 

The information you shared, this will be valuable information to have for the long run. I.e. to have this data stored in our database. Let's discuss the details tomorrow and determine next steps :))

PR #836, opened 2024-12-04
Lower select and embedding bw ops

rpavlovicTT:
**SelectOp**
- Lower to select op directly to ttir::SelectOp.
- Don't decompose select in post autograd.
- Create select op with named attrs so lowering to mlir works.

**EmbeddingOp**
- embedding op creates specific embedding_bw op in autograd pass which lowers directly to ttir::EmbeddingBackwardOp.

**Llama backward**
- Added simple UT to test autograd pass and lowering to TTMLIR
- Currently doesn't work because of dtype constraints (opened issue in tt-mlir)

Misc: few changes related to `named_attrs` property of ForgeOp.

### DIFF
--- a/forge/forge/op/eval/forge/embedding_bw.py
+++ b/forge/forge/op/eval/forge/embedding_bw.py
@@ -7,6 +7,7 @@ from forge._C import DataFormat
 from forge._C.graph import RuntimeTensorTransform, RuntimeTensorTransformType
 from ....forgeglobal import TILE_DIM
 
+
 def eval(type, attr, ops):
     assert type == "embedding_bw"
     assert len(ops) == 2
@@ -27,12 +28,12 @@ def shape(type, attr, tensor_shapes):
 
 
 def lower(type, attr, lc, ops, outputs):
-    pass
+    assert False, "embedding_bw should not be lowered"
 
 
 def decompose(type, attr, dc, inputs):
-    pass
+    assert False, "embedding_bw should not be decomposed"
 
 
 def backward(type, attr, ac, operand, inputs, output, grad):
-    pass
+    assert False, "embedding_bw should not be backwarded"
--- a/forge/forge/op/eval/forge/embedding_bw.py
+++ b/forge/forge/op/eval/forge/embedding_bw.py
@@ -1,0 +1,38 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+import torch
+from ..common import to_torch_operands
+from forge._C import DataFormat
+from forge._C.graph import RuntimeTensorTransform, RuntimeTensorTransformType
+from ....forgeglobal import TILE_DIM
+
+def eval(type, attr, ops):
+    assert type == "embedding_bw"
+    assert len(ops) == 2
+    t_ops = to_torch_operands(*ops)
+    input = t_ops[0]
+    weight = t_ops[1]
+    grad = t_ops[1]
+
+    result = torch.zeros(weight.shape)
+    for i, idx in enumerate(input):
+        result[idx] = grad[i]
+    return result
+
+
+def shape(type, attr, tensor_shapes):
+    assert type == "embedding_bw"
+    return tensor_shapes[1], []
+
+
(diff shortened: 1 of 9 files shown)


forge/test/mlir/test_ops.py
```diff
@@ -2192,3 +2192,32 @@ def forward(self, x):
...
+    framework_model = Select()
+    fw_out = framework_model(*inputs)
+
+    compiled_model = forge.compile(framework_model, sample_inputs=inputs)
+
+    inputs = to_pt_tensors(inputs)
+    compiled_output = compiled_model(*inputs)
+
+    co_out = [co.to("cpu") for co in compiled_output]
+    assert compare_with_golden_pcc(golden=fw_out.value(), calculated=co_out[0], pcc=0.99)
```
dgolubovicTT, line 2223: use compare_with_golden as it is more general (calls compare_with_golden_pcc for vectors and allclose for scalars)

forge/test/mlir/llama/test_llama_backward.py
```diff
@@ -0,0 +1,30 @@
...
+from transformers import LlamaConfig, LlamaForCausalLM, LlamaTokenizer
+
+import forge
+from test.mlir.llama.utils.utils import load_model
+
+
+# TODO(tt-mlir issue #1503): This test is failing because the embedding op doesn't work with FP32.
+# It should be fixed in the tt-mlir compiler soon.
+@pytest.mark.parametrize("model_path", ["openlm-research/open_llama_3b"])
+@pytest.mark.xfail()
```
pilkicTT, line 16: Since you're only compiling the model in this test, should this be marked as xfail?

forge/forge/op/eval/forge/embedding_bw.py
```diff
@@ -0,0 +1,38 @@
...
+        result[idx] = grad[i]
+    return result
+
+
+def shape(type, attr, tensor_shapes):
+    assert type == "embedding_bw"
+    return tensor_shapes[1], []
+
+
+def lower(type, attr, lc, ops, outputs):
```
pilkicTT, line 29: Please add `assert False, "Not expected to be used"` or something like that, in `lower()`, `decompose()` and `backward()`

forge/forge/op/tm.py
```diff
@@ -85,7 +85,9 @@ def Reshape(name: str, operandA: Tensor, shape: Tuple[int, ...]) -> Tensor:
     return op("reshape", name, operandA, attrs=shape, shape=shape).get_tensor()
 
 
-def Index(name: str, operandA: Tensor, dim: int, start: int, stop: int = None, stride: int = 1) -> Tensor:
+def Index(
+    name: str, operandA: Tensor, dim: int, start: int, stop: int = None, stride: int = 1
```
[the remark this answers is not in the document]
rpavlovicTT: Interesting, I ran Black format and got these. I will run pre-commit and revert fix whitespaces accordingly.

forge/test/mlir/llama/test_llama_backward.py
rpavlovicTT: Yeah because compilation fails. We have checks in tt-mlir to verify if unexpected data format is used.

forge/test/mlir/test_ops.py
rpavlovicTT: Do you mean `VerifyUtils.verify` or some else `verify`?

pilkicTT: Here's the commit which introduced the `verify()` function: 17442e9

The common usage is:
```
verify(inputs, framework_model, compiled_model)
```

This will run the compiled model and compare its results with the golden (framework output).

forge/forge/op/tm.py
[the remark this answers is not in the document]
pilkicTT: maybe you've ran it with settings from `tt-mlir` which I think has lower limit for line length... 😄

forge/test/mlir/llama/test_llama_backward.py
pilkicTT: Ok, I assumed it fails in ttnn...

pilkicTT (approved):

PR #841, opened 2024-12-05
Add test without gradient accumulation

### LINKED ISSUE #806: Rework mnist training script
test_mnist_training() from forge/test/mlir/mnist/training/test_training.py should be reworked to showcase standard batched training.
It makes sense to have an implementation similar to the current one in order to test gradient accumulation - it just needs to be renamed and explained through comments.

pglusacTT:
Add a new test that does standard batched training without gradient accumulation.
Renames the existing test to specify that it has gradient accumulation.

Closes #806


forge/test/mlir/mnist/training/test_training.py
```diff
@@ -21,6 +21,92 @@
...
+            framework_optimizer.zero_grad()
+
+            # Create target tensor and leave on CPU
+            target = nn.functional.one_hot(target, num_classes=10).to(dtype)
+
+            # Forward pass (prediction) on device
+            pred = tt_model(data)[0]
+            golden_pred = framework_model(data)
+            assert golden_pred.dtype == dtype
+            assert compare_with_golden(golden_pred, pred, verify_cfg=VerifyConfig(pcc=0.95))
```
dgolubovicTT, line 71: After this [PR](https://github.com/tenstorrent/tt-forge-fe/pull/739), there is a unified method for comparing golden output with output of compiled model: [linkt to verify](https://github.com/tenstorrent/tt-forge-fe/commit/17442e9aa4303dd6189964cffa9d2fc9262fb472#diff-e20a777f7a22e33ebefaedc11a005cbb40c7ca37adbe814a5d84baeae45e775bR266). It calls this compare_with_golden but also performs dataformat and some other sanity checks...

In this PR above, you can find examples of its use in tests. It makes things more concise...

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -21,6 +21,92 @@
 def test_mnist_training():
     torch.manual_seed(0)
 
+    # Model and data type. To use bfloat16 follow the instructions found at test_forge_vs_torch() in this file.
```
vladimirjovanovicTT, line 24: test_forge_vs_torch is moving into tt-thomas - it makes sense to copy the instructions here.

forge/test/mlir/mnist/training/test_training.py
pglusacTT: I need the predictions from the models since I am calculating the loss after this.
Should I change the `verify` method to make it return the predictions?

forge/test/mlir/mnist/training/test_training.py
pglusacTT: Makes sense, done here de73a93e9f1d67ef233382203264d7929ea23cc4

forge/test/mlir/mnist/training/test_training.py
dgolubovicTT: Verify is under development at the moment (by @vkovinicTT), so I wouldn't change it. Let's leave `compare_with_golden` then...
@vkovinicTT did you plan on adding return for the predictions any time soon?

pilkicTT: Let's not diverge and extend the scope of planned changes for the verify refactor, yet. There are not so many training tests, so it's not a big problem at this moment IMO...

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -21,6 +21,96 @@
...
+    # And in file forge/forge/compile.py::compile_main forced bfloat 16 should be added compiler_cfg.default_df_override = DataFormat.Float16_b
+    dtype = torch.float32
+
+    # Set training hyperparameters
+    num_epochs = 3
+    batch_size = 64
+    learning_rate = 0.001
+
+    # Limit number of batches to run - quicker test
+    limit_num_batches = 1000
```
pilkicTT, line 37: How about removing this "parameter" and just run this with higher batch size, i.e. 2048?

pilkicTT (commented):
Also, I am not sure if we should keep the old test? We should probably test the gradient accumulation part differently (sort of like a unit test)...

vladimirjovanovicTT:
> Also, I am not sure if we should keep the old test? We should probably test the gradient accumulation part differently (sort of like a unit test)...

I agree that it should be reworked, but maybe in a different PR? We can create an issue for this.

forge/test/mlir/mnist/training/test_training.py
vladimirjovanovicTT: MNIST is 60k inputs so with a large batch size we will never reach limit_num_batches. I agree that limit_num_batches can be removed.

forge/test/mlir/mnist/training/test_training.py
vkovinicTT: I didn't plan on extending it, but I can do that if needed?

forge/test/mlir/mnist/training/test_training.py
pglusacTT: Done here: ae7bee0d7d39d5b4c95d22ea926d71c71998493f

pilkicTT (approved):

PR #843, opened 2024-12-06
Restructure unique op test generation

### LINKED ISSUE #723: Restructure the generating unique op test function
**Function Purpose**
The generate_unique_op_tests function in the tvm_to_python.py file is used for extracting unique op configuration based upon the operand node type(i.e Activation/Parameter/Constant), operand shape and operand datatype and then compare the operation argument if any and store operation names in association with operation argument and generate the  forge module and test.

**Restructure the generate_unique_op_tests function:**
     1. We are collecting process framework parameter function status when there is params or constants used in the forge module class defination and forward function and then use it in pytest parameter instead we declare the function as method with ForgeModule (i.e base class) and so we can avoid this in the generate_unique_op_tests and write_pytest_function in ForgeWritter.

### LINKED ISSUE #724: [Model Analysis] Move the unique ops generation to separate module
In tvm_to_python.py file, the models unique operation configuration are extracted using the UniqueOperations, UniqueOperationInfo, OpArgsOpNames, OperandsInfo and OpArgs class. Move the mentioned class to separate python module instead of having it in tvm_to_python.py file.

chandrasekaranpradeep:
Fixes #723 
Fixes #724 

1. Move the unique op extraction logic and generation of unique op forge module with test function to separate module.
2. Remove the generation of single op forge module with test function and the flags present in the forge config and confest.py file
3. Remove the dependency of process framework parameter function status collection in unique op forge module generation  logic and write_pytest_function in the ForgeWritter by having the function in the base class ForgeModule.
4. Modified the framework and compiled output comparision logic with new verify logic in the write_pytest_function in the ForgeWritter .


forge/forge/module.py
```diff
@@ -963,6 +963,12 @@ def _set_device(self, device: "Device"):
         for submodule in self.get_submodules().values():
             submodule._set_device(device)
 
+    def process_framework_parameters(self):
```
dgolubovicTT, line 966: What is the point of this one? I am not that familiar with unique op test generation, so maybe I am missing something...

dgolubovicTT: Great, thanks!

dgolubovicTT (approved):

PR #847, opened 2024-12-06
Verification enhanced v1

vkovinicTT:
Using enums instead of booleans for verification. So now user can just:
```
verify(inputs, fw_model, co_model, VerifyConfig(verifyTensorValues=VerifyTensorValues.AUTOMATIC))
verify(inputs, fw_model, co_model, VerifyConfig(verifyTensorValues=VerifyTensorValues.AUTOMATIC, pcc=0.95))
verify(inputs, fw_model, co_model)
```
But if user doesn't really know how verification works, he can also call:
```
verify(inputs, fw_model, co_model, VerifyConfig(verifyTensorValues=VerifyTensorValues.TORCH_ALL_CLOSE, pcc=0.95))
```
where he sets `pcc` which is never used in this type of verification :(


forge/forge/verify/compare.py
```diff
@@ -0,0 +1,269 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+
+
+from typing import Union
+import os
+
```
pilkicTT, line 8: Is there a reason why you didn't send out the PR for resolving the circular dependency issue? If there is none, please send it, so that we can do this in chunks...

nvukobratTT:
@vkovinicTT is this PR still applicable?

vkovinicTT:
> @vkovinicTT is this PR still applicable?

no we are done with it

PR #848, opened 2024-12-06
Verification enhanced v2

vkovinicTT:
In this verification we have abstract `ValueChecker` class which can be derived into:
`AutomaticValueChecker`, `AllCloseValueChecker`, `FullChecker`. Default is using `AutomaticValueChecker` so we don't have problems with `torch.allclose`.
Here if user wants to perform full verification, it can be done like this:
```
verify(inputs, fw_model, co_model, VerifyConfig(valueChecker=FullChecker()))
verify(inputs, fw_model, co_model, VerifyConfig(valueChecker=FullChecker(pcc=0.95, rtol=1e-05)))
```
The advantage of this approach is that user only sets things that will be used in verification.

**NOTE**: in this PR I also made function that will tell us which `atol` and `rtol` to use in order for check to pass.

### DIFF
--- a/forge/forge/python_codegen.py
+++ b/forge/forge/python_codegen.py
@@ -157,7 +157,6 @@ class ForgeWriter(PythonWriter):
             self.wl("from forge.verify.compare import compare_with_golden")
             self.wl("from forge.verify.verify import verify")
             self.wl("from forge.verify.config import VerifyConfig")
-            self.wl("from forge.verify.compare import compare_with_golden")
             self.wl("import pytest")
 
         if self.framework == "tensorflow":
--- a/forge/forge/python_codegen.py
+++ b/forge/forge/python_codegen.py
@@ -157,6 +157,7 @@ class ForgeWriter(PythonWriter):
             self.wl("from forge.verify.compare import compare_with_golden")
             self.wl("from forge.verify.verify import verify")
             self.wl("from forge.verify.config import VerifyConfig")
+            self.wl("from forge.verify.compare import compare_with_golden")
             self.wl("import pytest")
 
         if self.framework == "tensorflow":
--- a/forge/forge/verify/compare.py
+++ b/forge/forge/verify/compare.py
@@ -35,8 +35,8 @@ def compare_with_golden(
         all_close = torch.allclose(golden, calculated, rtol=rtol, atol=atol)
         if not all_close:
             req_atol, req_rtol = compute_required_tolerances(golden, calculated)
-            logger.error("Tensor mismatch. Current rtol={}, atol={}", rtol, atol)
-            logger.error("Required to pass the test rtol={}, atol={}", req_rtol, req_atol)
+            logger.error("Tensor mismatch. Required rtol={}, atol={}", rtol, atol)
+            logger.error("Observed maximum relative diff: {}, maximum absolute diff: {}", req_rtol, req_atol)
             logger.error("Golden: (shape = {}", golden.shape)
             logger.error(golden)
             logger.error("Calculated: (shape = {}", calculated.shape)
--- a/forge/forge/verify/compare.py
+++ b/forge/forge/verify/compare.py
@@ -12,39 +12,39 @@ import numpy as np
 from loguru import logger
 from scipy.spatial import distance
 
-from forge.verify.config import VerifyConfig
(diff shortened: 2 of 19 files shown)


forge/forge/verify/value_checkers.py
```diff
@@ -0,0 +1,104 @@
...
+            raise ValueError(
+                f"Data mismatch -> FullChecker (compare_with_golden): framework_model={fw_out}, compiled_model={co_out}"
+            )
+
+        all_close_check = True
+        if fw_out.dtype not in [
+            torch.int32,
+            torch.int64,
+            torch.bool,
+        ]:  # allclose doesn't make sense for integer/bool types
```
pilkicTT, line 91: we should have an assertion if the tensor data type is bool/int

vkovinicTT: I didn't put assertion on purpose, because most of the time people would want to call `FullChecker ` without thinking about dtypes or anything, so for that purpose I didn't put assertion here, but in `AllCloseValueChecker` where the whole purpose of the checker is to use `allclose` function I put the assertion regarding dtypes.

forge/forge/verify/compare.py
```diff
@@ -12,56 +12,69 @@
...
         golden = golden.flatten()[0]
         calculated = calculated.flatten()[0]
-        return torch.allclose(
-            golden, calculated, rtol=verify_cfg.rtol[golden.dtype], atol=verify_cfg.atol[golden.dtype]
-        )
+
+        all_close = torch.allclose(golden, calculated, rtol=rtol, atol=atol)
+        if not all_close:
+            req_atol, req_rtol = compute_required_tolerances(golden, calculated)
+            logger.error("Tensor mismatch. Current rtol={}, atol={}", rtol, atol)
```
pilkicTT, line 38: Please change the messages, it looks confusing to me. I would formulate it something like:
```python
logger.error("Tensor mismatch. Required rtol={}, atol={}", rtol, atol)
logger.error("Observed maximum relative diff: {}, maximum absolute diff: {}", req_rtol, req_atol)
```

forge/forge/python_codegen.py
```diff
@@ -157,6 +157,7 @@ def write_header(self, include_pytest_imports=False):
             self.wl("from forge.verify.compare import compare_with_golden")
             self.wl("from forge.verify.verify import verify")
             self.wl("from forge.verify.config import VerifyConfig")
+            self.wl("from forge.verify.compare import compare_with_golden")
```
pilkicTT, line 160: Remove this extra addition...

forge/forge/verify/value_checkers.py
```diff
@@ -0,0 +1,104 @@
...
+            torch.int32,
+            torch.int64,
+            torch.bool,
+        ], f"AllCloseValueChecker (all_close): all_close doesn't make sense for integer/bool types"
+
+        if not torch.allclose(fw_out, co_out, rtol=self.rtol, atol=self.atol):
+            atol, rtol = compute_required_tolerances(fw_out, co_out)
+            raise ValueError(
+                f"Data mismatch -> AllCloseValueChecker (all_close):\n"
+                f"- Current tolerances: rtol={self.rtol}, atol={self.atol}\n"
```
pilkicTT, line 57: See above comment about log messages

forge/forge/verify/value_checkers.py
pilkicTT: Yeah, my bad... Maybe update the description of `FullChecker`:

```python
class FullChecker():
    """ Performs PCC and all_close for float tensors. """
```

forge/forge/verify/value_checkers.py
```diff
@@ -0,0 +1,104 @@
...
+                f"- Required tolerances for test to pass: rtol={rtol}, atol={atol}\n"
+                f"- Framework output: ({fw_out.shape})\n"
+                f"{fw_out}\n"
+                f"- Compiled model output: ({co_out.shape})\n"
+                f"{co_out}"
+            )
+
+
+class FullChecker(AutomaticValueChecker):
```
pilkicTT, line 66: Can you inherit from `ValueChecker` instead?

pilkicTT (changes requested):

pilkicTT (approved):

PR #855, opened 2024-12-09
Increase the job execution timeout for model analysis weekly job

chandrasekaranpradeep:
[Model Analysis Weekly job](https://github.com/tenstorrent/tt-forge-fe/actions/runs/12207459164) is triggered on 11:00 PM UTC Friday (12:00 AM Saturday Serbia), the job is cancelled out while running the model analysis script because the default timeout for the job execution is 360 minutes (i.e 6 hours). So increase the job execution timeout to 7 days (i.e 10080 minutes).

### DIFF
--- a/.github/workflows/model-analysis-weekly.yml
+++ b/.github/workflows/model-analysis-weekly.yml
@@ -14,6 +14,7 @@ jobs:
   model-analysis:
     needs: docker-build
     runs-on: runner
+    timeout-minutes: 10080 # Set job execution time to 7 days(default: 6 hours)
 
     container:
       image: ${{ needs.docker-build.outputs.docker-image }}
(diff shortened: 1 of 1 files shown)


.github/workflows/model-analysis-weekly.yml
```diff
@@ -14,6 +14,7 @@ jobs:
   model-analysis:
     needs: docker-build
     runs-on: runner
+    timeout-minutes: 10080 # Set job execution time to 7 days(default: 6 hours)
```
vmilosevic, line 17: After 24h GH auto token will expire, so if you want to do even longer jobs we must make a PAT token for it
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

vmilosevic (approved):

vmilosevic (approved):

PR #862, opened 2024-12-10
Reshape operator test

vobojevicTT:


forge/test/operators/pytorch/tm/test_reshape.py
```diff
@@ -0,0 +1,178 @@
...
+            test_device=test_device,
+            input_shapes=input_shapes,
+            input_params=input_params,
+            input_source_flag=input_source_flag,
+            dev_data_format=test_vector.dev_data_format,
+            math_fidelity=test_vector.math_fidelity,
+            pcc=test_vector.pcc,
+            warm_reset=warm_reset,
+            value_range=ValueRanges.SMALL,
+        )
```
vbrkicTT, line 114: verify that output shape is same as target shape

vobojevicTT: Now we use new verify module that checks this.

forge/test/operators/pytorch/tm/test_reshape.py
```diff
@@ -0,0 +1,309 @@
...
+        num_dims = rng.randint(1, max_dims)  # Limit the maximum number of dimensions
+
+        # Attempt to divide total_elements into a random number of factors
+        target_shape = []
+        remaining = total_elements
+        for i in range(num_dims - 1):
+            if remaining == 1:
+                target_shape.extend([1] * (num_dims - len(target_shape)))
+                break
+            dim_size = rng.randint(1, remaining)
```
dgolubovicTT, line 52: This dim_size calculation is brute force. What do you thing about this: Factor remaining into prime factors and then choose some number of factors `k` to take and form a dim_size. Then choose `k` random factors from the list of factors. 

It should look something like this:

```
def random_factor(remaining):
    """Generate a random factor of the given number."""
    factors = prime_factors(remaining)
    len_factors = len(factors)

    if len_factors == 0:
        return 1  # If the number is 1 or prime, the only factor is 1.

    k = random.randint(1, len_factors)
    chosen_factors = random.sample(factors, k)

    # Calculate the product of the chosen factors to form a random factor
    random_factor = 1
    for f in chosen_factors:
        random_factor *= f

    return random_factor
```

To find prime factors, it would be something like:

```
def prime_factors(n):
    """Return the prime factors of a given number."""
    factors = []
    while n % 2 == 0:
        factors.append(2)
        n //= 2
    
    for i in range(3, int(sqrt(n)) + 1, 2):
        while n % i == 0:
            factors.append(i)
            n //= i
    
    if n > 2:
        factors.append(n)
    
    return factors
```

forge/test/operators/pytorch/tm/test_reshape.py
```diff
@@ -0,0 +1,361 @@
...
+    for factor in factors:
+        random_dim = rng.randint(0, num_dims - 1)
+        shape[random_dim] *= factor
+
+    # Step 4: Ensure the final shape has the correct volume
+    if math.prod(shape) != vol:
+        raise ValueError("Generated shape does not match the desired volume.")
+
+    # If we get the same shape as source_shape, we try to add one dimension with size 1
+    if tuple(shape) == source_shape and len(shape) < max_dims:
```
dgolubovicTT, line 94: Can we leave output shape as it is if it turns out that it is the same to input shape? So we would have some small percentage of identity reshapes...

vobojevicTT: @dgolubovicTT Thanks! Updated to keep the output shape if it matches the input.

vbrkicTT: Identity shapes happened rarely, I thing only one of shapes that we tested generated same shape for output, but logging doesn't hurt.

dgolubovicTT (approved):

Issue #761, opened 2024-11-25
Optimize the model analysis script and generate models ops test


chandrasekaranpradeep:
1) The model_analysis script is currently generating and running the unique op configuration test model wise instead collect the unique op configuration from all the models and then extract unique op configuration test from all the models by doing so we are avoid runing same configuration that are present in multiple models again and again. 
For example if add op contains two operands of  (1, 32)  and (1, 32) in resnet and mobilenet, will run the unique op test only once and populate the test details in both resnet and mobilenet

2) After extracting unique op test from all the models, store those unique op tests in a separate module and by utilizing it, create models ops tests and include them as part of push pipeline.

**Note:**
After generating those models ops test, need to create workflow for running those models ops tests and collect the test status. Based upon the test passing status, need to add them as part of push pipeline which will be done manually - [Reference](https://github.com/tenstorrent/tt-forge-fe/issues/795#issuecomment-2507525525)

**Limitation:**
- Need to refresh those generated models ops tests whenever there are any tvm changes being uplifted.
- We're not covering cases with multiple ops in the graph to exactly match the models.
- The generated models ops test doesn't contains one to one mapping with the models like using random tensor for weight/constant instead of loading actual parameter/constant from the saved model pt file which may leads issue like pcc drop which may not be arrive while using actual model parameter/constant. so first generate the nightly test and see is there any pcc drop or other issues arrives when the random tensor is passed in runtime. If something like pcc issue arrives, saved the model weights/constant that are needed for the unique op test instead of using whole model pt file.

3. Move the compiler_component and exception mapping list (i.e `common_failure_matching_rules_list`) to separate file for properly managing it

PR #874, opened 2024-12-11
Optimize the model analysis weekly pipeline

chandrasekaranpradeep:
The PR will optimise the model analysis weekly pipeline - #761 

The current model_analysis script will generate the unique op tests and running the unique op configuration test model wise instead collect the unique op test configuration from all the models and then extract unique op configuration test across all the models which will avoid running same configuration that are present in multiple models again and again.

For example, if we have add op with operands shape of (1, 32) and (1, 32) in resnet and mobilenet, will run the test only once and populate the test results in both resnet and mobilenet

### DIFF
--- a/forge/forge/python_codegen.py
+++ b/forge/forge/python_codegen.py
@@ -1085,7 +1085,7 @@ class ForgeWriter(PythonWriter):
         self.wl("")
         self.wl("compiled_model = compile(framework_model, sample_inputs=inputs)")
         self.wl("")
-        self.wl("verify(inputs, framework_model, compiled_model, VerifyConfig(verify_allclose=False))")
+        self.wl("verify(inputs, framework_model, compiled_model)")
         self.wl("")
         self.wl("")
         self.indent -= 1
--- a/forge/forge/tvm_unique_op_generation.py
+++ b/forge/forge/tvm_unique_op_generation.py
@@ -2,9 +2,10 @@
 
 # SPDX-License-Identifier: Apache-2.0
 import os
+import json
 from enum import Enum
 from loguru import logger
-from typing import Dict, List
+from typing import Any, Dict, List, Optional
 
 import torch
 
@@ -16,6 +17,14 @@ class NodeType(Enum):
     Activation = 1
     Parameter = 2
     Constant = 3
+
+    @classmethod
+    def to_json(cls, value):
+        return value.name
+
+    @classmethod
+    def from_json(cls, value):
+        return cls[value]
 
 
 class Operation:
@@ -34,12 +43,13 @@ class Operation:
         inputs_to_delete (list): A list of inputs to delete.
         loop_with (list): A list of loop variables.
         src_layer (optional): The source layer associated with the operation.
+        metadata (dict): It contains additional information associated with the operation like model, variant, framework
     """
 
     def __init__(
         self,
         function_name,
-        output_name,
+        output_name="",
         node_name="",
         input_names=[],
         args=[],
@@ -47,6 +57,7 @@ class Operation:
         input_shapes=[],
         input_dtypes=[],
         input_node_types=[],
+        metadata={},
     ):
         self.function_name = function_name
         self.node_name = node_name
@@ -60,6 +71,7 @@ class Operation:
         self.inputs_to_delete = []
         self.loop_with = []
         self.src_layer = src_layer
+        self.metadata = metadata
 
 
 class OpArgs(dict):
@@ -121,7 +133,7 @@ class OpArgs(dict):
         return len(self) == 0
 
     def __str__(self):
-        return f"Opargs({super().__str__()})"
+        return super().__str__()
 
 
 class OperandsInfo:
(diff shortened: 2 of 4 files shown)


nvukobratTT:
Great work @chandrasekaranpradeep! I don't see any major issues from code changes. However, I would like to confirm few details before proceeding with further review. That said, can you provide details about achieved results with this change? 

Here are a few questions as a guideline:
1. How long tests are executing now, vs before this change? Please provide details of machine on which you run this.
2. Summary of workflow changes? E.g. how generation and run work now (generate unique across all models; run only unique and populate details for all variants)
3. Do we have ability to separate test generation and test runs? E.g. to have a workflow where we generate these tests on weekly, but running them on nightly.
4. What is the process of including generated op tests as part of the push pipeline? 
5. More detail on how params/constants form model are handled? Do we rely on random tensors now, and can this be swapped easily per need? E.g. debugging PCC issues with original weights instead of ones generated in the runtime. 

Other:
1. @vmilosevic is it possible to have workflow run when the specific part of the code has changed? @chandrasekaranpradeep for example when TVM ref change, to run job to update these generated unique op tests.

chandrasekaranpradeep:
@nvukobratTT 

> 1. How long tests are executing now, vs before this change? Please provide details of machine on which you run this.

I have checked the behaviour of the implementation but haven't noted the execution time before and after the change, will check and let you know here.

> Summary of workflow changes? E.g. how generation and run work now (generate unique across all models; run only unique and populate details for all variants)

**Summary of previous workflow:**
1. Generate the unique op tests for the models present in the directory path specified by the user(i. e forge/test/models/pytorch):
    1. Collected all the pytest present in the specified directory path
    2. Run the pytest and extract unique op configuration and generate forge module with test functions and additionally export the unique op config test details as excel.
2. Run the generated unique op config tests and generate markdowns
     1. Run the generated unique op config forge test cases model  wise and generate sub markdown files contains unique operation configurations, pass/fail status for each compiler component, and failure reasons.
     2. After running all the models unique op test, generate root markdown file for summarising the result for all the models.
                  
**Summary of the new workflow:**
1. Generate the unique op tests for the models present in the directory path specified by the user(i. e forge/test/models/pytorch):
      1. Collecting and running the pytest and extracting unique op config and generating forge module is similar to the previous workflow but in addition to exporting unique op config test details as excel file, will export the model metadata information which includes module_name, framework name, model params and buffers file path which will be used for extracting unique op configuration across all the models.
2. Run the generated unique op config tests and generate markdowns.
      1. Instead of running the generated unique op config forge test cases model wise, by utilising the same UniqueOperations class(i.e which is used for extracting unique op config from model), will extract unique op configuration across all the models and run the extracted unique op configuration across all the models and populate the test result to the respective models. 
      2. After running extracted unique op config across all the models will generate sub markdown and root markdown files.


> 3. Do we have ability to separate test generation and test runs? E.g. to have a workflow where we generate these tests on weekly, but running them on nightly.
> 4. What is the process of including generated op tests as part of the push pipeline?
> 5. More detail on how params/constants form model are handled? Do we rely on random tensors now, and can this be swapped easily per need? E.g. debugging PCC issues with original weights instead of ones generated in the runtime.


All these question are related to generation of test for nightly from the extracted unique op configuration across all the models, the PR doesn't have implementation for generating test for nightly from extracted unique op configuration across all the models.

The [pchandrasekaran/generate_nightly_test](https://github.com/tenstorrent/tt-forge-fe/commit/0f3ca410912c5240f6a1d6a646d23fa7b80a8349) branch that I am working has the changes for generating test for nightly which will be later include as part of push pipeline

The idea is to generate nightly test by using the exported unique op config test details excel and models metadata info json files produced from the **Generate the unique op tests for the models present in the directory path specified by the user(i. e forge/test/models/pytorch) step** in the new workflow

In the model analysis script main function, I have added an option to generate nightly test.

When we enable the option of generating nightly test, after generating unique op test from all the models with xlsx and json files, we are extracting unique op configuration across all the models and then generate nightly test for extracted unique op configuration across models.

Instead of loading actual parameters and constants  from models with process_framework_paremeter method, now the parameters and constants are randomly generated  by  modifying the add_constants method in ForgeModule in forge/forge/module.py and Parameter class in forge/forge/parameter.py and also done some modification in the write_class_defination function in ForgeWritter.

Added recorded_property fixtures in the nightly test function which includes model_name, variant_name, framework_op_name and op_kind

In sum, the model_analysis script will additional generate nightly test with model_analysis_nightly marker and then create a separate PR for the generate nightly test and then we will have some github workflow to run the generated nightly based upon the test results we can remove the failed test cases from the same PR branch and check it and merge it into main.


Whats your thoughts on this nikola?

chandrasekaranpradeep:
Genereted nightly test from model analysis script:
[nightly.zip](https://github.com/user-attachments/files/18227452/nightly.zip)

nvukobratTT:
Thanks for the details! Sounds good! 

>The [pchandrasekaran/generate_nightly_test](https://github.com/tenstorrent/tt-forge-fe/commit/0f3ca410912c5240f6a1d6a646d23fa7b80a8349) branch that I am working has the changes for generating test for nightly which will be later include as part of push pipeline

Think about how we can include passing ones right away as part of push pipeline, while keeping other part of nightly. Ofc, this is more question for upcoming PR rather then this one. For other details related to futhre PR, let's discuss in our sync tomorrow.

nvukobratTT (approved):

PR #876, opened 2024-12-11
Add NLLLoss

pglusacTT:


forge/forge/op/loss.py
```diff
@@ -73,3 +73,18 @@ def forward(self, prediction, labels):
             return c_reduce
 
         raise RuntimeError("Unsupported reduce type: " + self.reduction)
+
+
+class NLLLoss(ForgeModule):
+    def __init__(self, name: str, reduction: str = "avg"):
+        super().__init__(name)
+        self.reduction = reduction
```
vladimirjovanovicTT, line 81: We are setting reduction, but we are not using it?

forge/test/mlir/test_loss.py
```diff
@@ -62,3 +62,36 @@ def test_cross_entropy_loss(prediction_shape):
...
+def test_nll_loss(prediction_shape):
+    forge_loss = forge.op.loss.NLLLoss("nll_loss")
+    torch_loss = torch.nn.NLLLoss()
+
+    prediction = torch.randn(prediction_shape, requires_grad=True).to(torch.float32)
+    prediction_forge = forge.tensor.Tensor.create_from_torch(prediction)
+    target = torch.empty(prediction_shape[0], dtype=torch.long).random_(prediction_shape[-1])
+
+    # Because of the following error
+    # RuntimeError: TT_FATAL @ ../embedding_device_operation.cpp:28: weights.get_dtype() == DataType::BFLOAT16
```
vladimirjovanovicTT, line 87: could you check whether this ties to an existing issue? Maybe mention the issue nr here if it does.

forge/forge/op/loss.py
```diff
@@ -73,3 +73,18 @@ def forward(self, prediction, labels):
...
+        self.is_loss = True
+
+    def forward(self, prediction, labels):
+        zeroed_prediction = Multiply("mul_pred_labels", prediction, labels)
```
vladimirjovanovicTT, line 85: I would suggest to use another name here - labels are one-hot in majority of cases, but can be an arbitrary distribution.

vladimirjovanovicTT (changes requested):
Looks good.

forge/forge/op/loss.py
```diff
@@ -119,3 +119,18 @@ def forward(self, predictions, labels):
...
+class NLLLoss(ForgeModule):
+    def __init__(self, name: str):
+        super().__init__(name)
+        self.is_loss = True
+
+    @validate_shapes(min_dim=2, max_dim=2)
+    def forward(self, prediction, labels):
+        weighted_prediction = Multiply("mul_pred_labels", prediction, labels)
+        rows = ReduceSum("r_sum", weighted_prediction, -1)
+        loss = ReduceAvg("r_avg", rows, -2)
```
pmarkovicTT, line 133: nit: can we move this at the end after Multiply and use dim=0 instead of -2. Just to align with CrossEntroyLoss since those are similar

forge/forge/op/loss.py
```diff
@@ -119,3 +119,18 @@ def forward(self, predictions, labels):
         square = Multiply("square", diff, diff)
         loss = reduce_loss(self.reduction, square)
         return loss
+
+
+class NLLLoss(ForgeModule):
+    def __init__(self, name: str):
```
pmarkovicTT, line 125: Add comment as in CrossEntropy as well, but with note that input should be in the log-space

forge/forge/op/loss.py
```diff
@@ -119,3 +119,18 @@ def forward(self, predictions, labels):
...
+        super().__init__(name)
+        self.is_loss = True
+
+    @validate_shapes(min_dim=2, max_dim=2)
```
pmarkovicTT, line 129: Check if this still holds (min_dim=2)

pmarkovicTT (approved):
lgtm

forge/test/mlir/test_loss.py
```diff
@@ -78,20 +76,24 @@ def test_cross_entropy_loss(prediction_shape):
         (128, 128),
     ],
 )
-@pytest.mark.parametrize("reduction", ["mean", "sum"])
-def test_mse_loss(prediction_shape, reduction):
```
pilkicTT, line 82: Why is the MSE test removed?

pglusacTT: Accidentally, thanks for noticing, returning it ASAP!

pilkicTT (approved):

PR #880, opened 2024-12-12
CI Fix - Replace libgl1-mesa-glx with libgl1 libglx-mesa0

kmabeeTT:
- Package was recently deleted gives error 404
 - Update build-and-test.yml and model-analysis-weekly.yml, and need an apt-get update too.


pilkicTT (approved):
Thanks, @kmabeeTT!

PR #900, opened 2024-12-12
Add resnet unique op tests that fail on main, marked as xfail.

dgolubovicTT:
I used unique op test generator tool to extract unique ops from ResNet model and see which ones don't work.
I've used those generated op tests to create repro tests in `forge/test/mlir/test_model_unique_ops.py`.

Tests that fail: avg_pool2d, max_pool2d, matmul, unsqueeze.


pilkicTT (approved):
Please move the added tests under `forge/test/mlir/resnet` directory

PR #903, opened 2024-12-12
Test for embedding op

kmilanovicTT:
Test torch embedding operator

Supported operand source models for Forge:

from dram
from host
from another op
const eval pass
mf and df tests

Closes https://github.com/tenstorrent/tt-forge-fe/issues/267

### DIFF
--- a/forge/test/operators/pytorch/nn/test_embedding.py
+++ b/forge/test/operators/pytorch/nn/test_embedding.py
@@ -1,0 +1,467 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+#
+# Tests for testing of embedding operators
+#
+# In this test we test pytorch embedding operator
+
+# GENERAL OP SUPPORT TEST PLAN:
+# 1. Operand type - any supported type
+# 2. Operand source(s):
+# (+)  2.1 From another op
+#       - Operator -> input
+# (+)  2.2 From DRAM queue
+#       - Operator is first node in network
+#       - Input_queue flag = false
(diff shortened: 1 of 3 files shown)


forge/test/operators/pytorch/nn/test_embedding.py
```diff
@@ -0,0 +1,422 @@
...
+
+    rng = random.Random(31)
+    INPUT_SHAPE_THRESHOLD = 100000000
+    MAX_EMBEDDING_DIM = 10000
+
+    embedding_dims = [1000, 3200, MAX_EMBEDDING_DIM]
+    # weight_dtypes = [torch.bfloat16, torch.float32]
+
+    @classmethod
+    def generate_kwargs(cls, test_vector: TestVector, weight_dtype: Type[torch.dtype]):
```
vbrkicTT, line 235: Currently test_single is not working because "parsing" of torch dtype is failing during kwargs eval.
Solution is to include `import torch` in `utils/plan.py`

kmilanovicTT:
Before:

forge/test/operators/pytorch/nn/test_embedding.py
```diff
@@ -0,0 +1,530 @@
...
+            input_sources=[
+                InputSource.FROM_HOST,
+                InputSource.FROM_DRAM_QUEUE,
+                InputSource.CONST_EVAL_PASS,
+                InputSource.FROM_ANOTHER_OP,
+            ],
+            input_shapes=[
+                (1, 9920, 1),
+                (1, 1, 9920, 1),
+            ],
```
vbrkicTT, line 515: Probably (9220, 1) should be escaped as well

PR #912, opened 2024-12-16
Add triplet margin loss

### LINKED ISSUE #911: TripletMarginLoss support

pglusacTT:
Add support for triplet margin loss

Closes #911


forge/forge/op/loss.py
```diff
@@ -119,3 +121,51 @@ def forward(self, predictions, labels):
         square = Multiply("square", diff, diff)
         loss = reduce_loss(self.reduction, square)
         return loss
+
+
+class TripletMarginLoss(ForgeModule):
```
pmarkovicTT, line 126: nit: description

pmarkovicTT (approved):
lgtm

forge/forge/op/loss.py
nvukobratTT: +1 for desc

forge/forge/op/loss.py
```diff
@@ -119,3 +121,51 @@ def forward(self, predictions, labels):
...
+    def __init__(self, name: str, margin: float = 1.0, reduction: str = "mean"):
+        super().__init__(name)
+        self.margin = margin
+        # pow is fixed to 2.0
+        # https://github.com/tenstorrent/tt-mlir/issues/1203
```
nvukobratTT, line 131: Seems like a metal issue that was blocking this one is solved? Can we look into it and enable it here if it works? 
- https://github.com/tenstorrent/tt-metal/issues/13857

nvukobratTT: If yes, let's add it as attribute to this loss

forge/forge/op/loss.py
```diff
@@ -119,3 +121,51 @@ def forward(self, predictions, labels):
         square = Multiply("square", diff, diff)
         loss = reduce_loss(self.reduction, square)
         return loss
+
+
+class TripletMarginLoss(ForgeModule):
+    def __init__(self, name: str, margin: float = 1.0, reduction: str = "mean"):
```
nvukobratTT, line 127: One more benefit of having a function description is that we can outline our limitations here. 

Are we limiting support of the following attributes:
- eps for numerical stability
- distance swap

forge/forge/op/loss.py
```diff
@@ -119,3 +121,51 @@ def forward(self, predictions, labels):
...
+        self.margin = margin
+        # pow is fixed to 2.0
+        # https://github.com/tenstorrent/tt-mlir/issues/1203
+        self.p = 2.0
+        self.reduction = reduction
+        self.is_loss = True
+
+    @validate_shapes(min_dim=1, max_dim=2)
+    def forward(self, anchor, positive, negative):
+        if anchor.ndim() == 1:
```
nvukobratTT, line 138: Do we want to do an implicit unsqueeze, or just assert if inputs are incorrect? 

My thinking comes from the angle that this can cause unexpected shape mismatches with other parts of the graph. E.g. fwd and bwd have 1D at this point, while loss returned [1, N] dim. Will this cause failures?

forge/test/mlir/test_loss.py
```diff
@@ -95,3 +95,36 @@ def test_mse_loss(prediction_shape, reduction):
...
+    positive = torch.randn(prediction_shape, requires_grad=True)
+    positive_forge = forge.tensor.Tensor.create_from_torch(positive)
+    negative = torch.randn(prediction_shape, requires_grad=True)
+    negative_forge = forge.tensor.Tensor.create_from_torch(negative)
+
+    forge_loss = forge.compile(forge_loss, sample_inputs=[anchor_forge, positive_forge, negative_forge])
+    forge_loss_out = forge_loss(anchor_forge, positive_forge, negative_forge)
+    torch_loss_out = torch_loss(anchor, positive, negative)
+
+    assert torch.allclose(torch_loss_out, forge_loss_out[0], rtol=5e-2)
```
nvukobratTT, line 130: Can we use our verify() function instead of direct torch.allclose? For example, feel free to check some model tests. E.g. forge/test/models/pytorch/*

pglusacTT: Yes, I think in this way we can. Since the other loss functions also don't use verify, is it okay to create an issue and refactor all of them at once in a separate PR?

nvukobratTT: Yup, ofc! :)))

nvukobratTT:
@pglusacTT is this one ready for review? If not, can we move it to draft?

forge/forge/op/loss.py
pglusacTT: I've added support for those attributes now

forge/forge/op/loss.py
pglusacTT: I agree that it feels wrong to return [1, N] if the input is N, I've changed it so we explicitly require 2D tensor

nvukobratTT (approved):
Looks good!

forge/test/mlir/test_loss.py
```diff
@@ -262,3 +262,38 @@ def test_bce_with_logits_loss(prediction_shape, reduction):
...
```
nvukobratTT, line 299: Suggestion to use verify instead of all close

pglusacTT: We talked about it here :) https://github.com/tenstorrent/tt-forge-fe/pull/912#discussion_r1919632564

I've created an issue https://github.com/tenstorrent/tt-forge-fe/issues/1317

nvukobratTT: This one is a bit different, as you're not doing any parameter verification related to training, but simple outputs (similar to inference). That said, you should be able to use `verify` here without issues.

nvukobratTT (approved):

PR #913, opened 2024-12-16
MNIST lora layer test

pmarkovicTT:
Add test to make sure LoRA layer can be implemented and used in a full training pipeline.

Problem encountered
Hitting `KeyError: 'gradient_lora1.b.consteval_graph.output'` when implementing LoRA layer with nn.Parameters.
```
class LoraLayer(nn.Module):
    def __init__(self, input_size, output_size, rank=8, alpha=4, dtype=torch.float32):
        super(LoraLayer, self).__init__()
        self.a = nn.Parameter(torch.empty(input_size, rank, dtype=dtype), requires_grad=True)
        self.b = nn.Parameter(torch.zeros(rank, output_size, dtype=dtype), requires_grad=True)
        self.alpha = alpha / rank

        nn.init.normal_(self.a, mean=0, std=1)

    def forward(self, x):
        return self.alpha * (x @ self.a @ self.b)
```

Raised issue: https://github.com/tenstorrent/tt-forge-fe/issues/929

### DIFF
--- a/forge/test/mlir/mnist/training/test_training.py
+++ b/forge/test/mlir/mnist/training/test_training.py
@@ -410,3 +410,84 @@ def test_loss_device():
             break
 
     print(f"Test (total) loss: {test_loss}")
+
+
+@pytest.mark.push
+def test_lora():
+    torch.manual_seed(0)
+
+    # Config
+    num_epochs = 3
+    batch_size = 1
+    learning_rate = 0.001
+
+    # Limit number of batches to run - quicker test
+    limit_num_batches = 1000
+
+    # Load dataset
+    test_loader, train_loader = load_dataset(batch_size)
+
+    # Load TensorBoard writer (for logging)
+    writer = load_tb_writer("forge_mnist")
+
+    framework_model = MNISTLinear(bias=False)
+    framework_optimizer = torch.optim.SGD(framework_model.parameters(), lr=learning_rate)
+
+    tt_model = forge.compile(framework_model, sample_inputs=[torch.rand(batch_size, 784)], training=True)
+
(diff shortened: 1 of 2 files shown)


forge/test/mlir/mnist/training/test_training.py
```diff
@@ -410,3 +410,80 @@ def test_loss_device():
...
     print(f"Test (total) loss: {test_loss}")
+
+
+@pytest.mark.push
+def test_lora():
+    torch.manual_seed(0)
+
+    # Config
+    num_epochs = 3
+    batch_size = 1
```
pglusacTT, line 421: We could use a batch size larger than 1 here and no gradient accumulation to simulate a standard training scenario?

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -410,3 +410,80 @@ def test_loss_device():
...
+        framework_optimizer.step()
+
+    test_loss = 0
+    for batch_idx, (data, target) in enumerate(test_loader):
+        pred = tt_model(data)[0]
+        target = nn.functional.one_hot(target, num_classes=10).float()
+
+        test_loss += tt_loss(pred, target)[0]
+
+        if batch_idx == 1000:
```
pglusacTT, line 487: Same here, or use `limit_num_batches` variable

pglusacTT (approved):
🚀

pilkicTT (approved):
For the issue you've encountered; it seems that we somehow end up constevaluating the `lora1.b` parameter even though it is trainable (`requires_grad=True`). Can you please create a branch with repro and file an issue for this?

pmarkovicTT:
Yep, done it.
https://github.com/tenstorrent/tt-forge-fe/issues/929

PR #915, opened 2024-12-17
Speed up test plan querying

vbrkicTT:
Speed up test plan querying via filtering test plans by OPERATORS filter prior to vector filtering.


dgolubovicTT (approved):
LEGIT

Issue #922, opened 2024-12-17
update Llama 3.2 1B tests


mstojkovicTT:
We have updated the requirements file, PR #904, to incorporate newer versions of various modules, including Llama 3.2. Consequently, tests that were previously skipped due to compatibility issues have had their `@pytest.mark.skip` decorators removed.

mstojkovicTT:
Waiting for either a Hugging Face token or storage to save models before removing `pytest.mark.skip`

nvukobratTT:
@mstojkovicTT token should be updated on our CI. Feel free to double-check if it's working now :))

PR #923, opened 2024-12-17
update Llama 3.2 1B tests

mstojkovicTT:
I have removed the `@pytest.mark.skip` decorators from tests related to Llama 3.2. Additionally, I have separated the Llama 3b and Llama 3.2 tests to allow for independent testing.

fixes: #922


mstojkovicTT:
Reverted skips, waiting for either a Hugging Face token or storage to save models.

pilkicTT (changes requested):
As we've talked offline, let's not duplicate test code...

PR #928, opened 2024-12-18
[train] optimizer on device

### LINKED ISSUE #176: Training MNIST based Linear model e2e on device
## Summary
In order to train this model, the following key details are required:
- Required `fwd`, `bwd`, `loss` and `opt` ops are supported e2e
  -  tenstorrent/tt-mlir#77
  -  tenstorrent/tt-mlir#145
- Inference is running e2e
  - tenstorrent/tt-mlir#149
  - tenstorrent/tt-mlir#150 
- Runtime API can handle model execution
  - tenstorrent/tt-forge-fe#179
  - tenstorrent/tt-forge-fe#174 
  - tenstorrent/tt-forge-fe#178 
  - tenstorrent/tt-forge-fe#175 
  - tenstorrent/tt-forge-fe#177 

Having the components mentioned above, pushing initial e2e training should be straightforward.

## Details
With all the mentioned components, most of the details should be in place. For easier understanding, here is the overview of how compile flow should work and look like:
1. We define our training model and training loop. Check `pybuda/test/mlir/mnist/test_training.py` for a sample. 
2. Based on the training loop, runtime push inputs to appropriate locations (e.g. host, dram) and executes model compile
3. Each model is compiled separately, but runtime is aware of inputs/outputs locations (e.g. l1, dram, etc). Utilize [runtime stitching](https://tenstorrent.github.io/tt-mlir/specs/runtime-stitching.html) as a concept to drive this
4. Depending on how the user defines the training loop, runtime executes graphs and updates specific memory locations as defined by flatbuffer metadata
  - MLIR shouldn't be aware of graph purpose, they should be independent (more details in follow-up sections)

### Model compile
It's important to note that each model is compiled separately. E.g. each `pybuda.compile` executes a separate compile workflow that generates flatbuffers and returns them to the PyBuda runtime together with appropriate metadata that runtime needs (e.g. where to find inputs)

During model compile, each component should be compiled separately and be unaware of each other. That way, we can strive to more streamlined support of other frameworks, and keep the flexibility of compilation. This ensures that users have an ability to use the [PyTorch training loop style](https://pytorch.org/tutorials/beginner/introyt/trainingyt.html#the-training-loop). That way, our customers can integrate our compiler as plug-and-play, where only a few lines of their existing workflows should be updated.

Likewise, we should utilize something like [runtime stiching](https://tenstorrent.github.io/tt-mlir/specs/runtime-stitching.html) in order to support independent module/components compile and runtime workloads. Only runtime for each frontend should be aware of inference/training specifics. MLIR should be agnostic to it (same applies for flatbuffer schemas).

### Details around model training
Main point of training is for model to "learn" how to do the desired task. You can view the model as a simple "mapper". I.e. "mapping" one type of input to desired output. For example, "mapping" an image of a cat, into the actual prediction string "cat". In this section, I'll focus only on training points that are related to the compiler.

In sum, once the model weights are initialized (automatically done when the model is created) we're running the training loop and doing the following (code sample here: `pybuda/test/mlir/mnist/test_training.py`):
1. Forward pass and get model predictions
2. Calculate loss by referencing forward pass outputs (predictions) and targets (ground truth labels)
3. Backward pass that uses loss output to calculate gradients for each model parameter
4. Optimization pass that uses calculated (or accumulated) gradients to update original model parameters

For a more precise example, here are a few graphs coming from how the training worked on the older PyBuda/BBE stack.

#### Forward graph with attached loss
image

#### Backward graph:
image

#### Optimizer graph:
image

### Training loop
To easily map above-mentioned concepts, here is the proposed sample of how a training loop can look like.
```py
    # Define model and instruct it to compile and run on TT device
    framework_model = MNISTLinear()
    tt_model = pybuda.compile(framework_model)
    tt_model.to("tt")

    # Create a torch loss and leave on CPU
    tt_loss_fn = torch.nn.L1Loss()

    # Define optimizer and instruct it to compile and run on TT device
    framework_optimizer = torch.optim.SGD(framework_model.parameters(), lr=learning_rate)
    tt_optimizer = pybuda.compile(framework_optimizer)
    tt_optimizer.to("tt")

    for epoch_idx in range(num_epochs):
        for batch_idx, (data, target) in enumerate(train_loader):
            # Put inputs on device
            data = data.to("tt")
            
            # Create target tensor and leave on CPU
            target = nn.functional.one_hot(target, num_classes=10).float()

            # Reset gradients (every batch)
            tt_optimizer.zero_grad()
            
            pred = tt_model(data)
            
            # Pull output back to CPU
            pred = pred.to("cpu")

            # Compute loss on CPU
            loss = tt_loss_fn(pred, target)
            
            # Run backward pass on device
            loss.backward()
            
            # Adjust weights (on device)
            tt_optimizer.step()
```

### Training components
In sum, this is still in exploration work, and we should strive to do as many R&D experiments as possible in order to choose the best approach for the long run. Therefore, some of these concepts might change over time.

Firstly, regarding the **forward** graph. In sum, it represents an inference graph. One caveat is that we previously once caveat is that we should determine how we can simply how loss is connected to it. In the previous stack, it was attached to the forward graph (like an extension). The details that need to be checked here are whether we have to be aware of the loss part of the graph, or if we can simply integrate it as a part of the forward graph, and compute backward on top of it. 

As already mentioned, the **loss** part is currently examined as an addition to the forward graph. If we prove that we can follow a similar mechanism and get valid training, it can simplify graph compilation a bit. However, this can become tricky for more advanced loss functions. As mentioned, we should explore this path in more detail. If it's not possible to merge those two, runtime will have to be aware of a loss component in the long run. To bridge that complexity, we can also run loss on the host/CPU for the initial training PoC.

Following is the **backward** part. It's generated based on the forward graph using the PyBuda autograd mechanism. It's used to compute gradients for each model parameter, which is latter used by the optimizer to properly update those gradients based on loss output. Overall, MLIR should view the backward graph as a plain graph, without any notion of which type is it. PyBuda runtime is the one that will orchestrate what is what and how to properly manage those components/data. 

And lastly, there is the **optimizer**. It's main point is to update model parameters using calculated gradients. Also, the optimizer can have additional inputs that define how parameters are updated (e.g. learning rate). Like each graph mentioned above, the optimizer should also be a standalone graph. The trick here is for runtime to know how to map gradients to the valid parameters. To handle this properly, we should ensure that the model is compiled first. That way, we can gather the list of model parameters, and match those to the calculated/accumulated gradients. Afterward, once we compile the optimizer and determine where the inputs/outputs are stored, we can just run a graph and update parameters based on how the training loop is defined by the user.

## Limitations
As expected, this initial overview has some limitations. The main one is that we're not exploring as part of the initial run how to scale out training. During this push, we should be aware that training will also run across multiple chips. However, for initial PoC it won't be forced. That mentioned, here is a list of the current limitations that we should think about:
- Training with recompute will not work for the initial version
    - Initial training will not include training with recompute. 
    The main point is that MLIR should be as agnostic as possible to the ML. I.e. it should only see different graphs that should be run. E.g. forwards, backward, optimizer. 
    - Also, the new BE is a temporal-based stack, so older implementation will not be easily applicable from spatial to temporal atp.
    - Therefore, we can't keep the concept of interconnected fwd and bwd ops in order to support recompute as done before. So, a new solution should be introduced that will support both PyBuda and Jax frontends. TBD.
- Multichip will not work for the initial version
    - Training on multichip has additional caveats 
- Not mandatory to be efficient from the start
    - For further experiments, we need to focus on multi-chip and larger models. E.g. for valid end sprint PoC, we need to demonstrate that we can scale the training across multiple machines in order to be able to train larger models

### Passing Criteria
- Training works e2e on the valid dataset
- We can save training checkpoints and load them for independent inference

### LINKED ISSUE #178: Support optimizer.step() e2e on device - running optimization step for updating model parameters
## Summary
Optimizer is required for weights update once gradients are calculated. Effort of this issue should be to map `optimizer.step()` as a runtime API to run optimization graph based on runtime metadata got after MLIR compile and flatbuffer generation.

## Details
Most of the optimizer related details can be found here tenstorrent/tt-forge-fe#176. In sum, optimizer should be separated flatbuffer as well (MLIR only sees the graph; not aware of training specifics). 

Key goal of this issue is to explore how we can map `optimizer.step()` to run optimizer flatbuffers. The flow can work like this:
1. Initialize model (fwd grph)
2. Mark it as training
3. Compile the model. It's important to compile model before optimizer (details below)
4. Compile the optimizer
5. Run training loop with optimizer

Few key points to note here are:
- For 3. point, we should enforce that model is compiled first. In order for optimizer to be able to map gradient accumulation inputs, and parameter outputs, runtime has to compile model first in order to get that informatin
- For 4. point, but similar to above mentioned point, optimizer needs to be aware of valid inputs and outputs. In sum, backward graph will contain information about where are gradients stored/accumulated. Before and after model compilation, runtime needs to keep parameter references and match those correctly: which grad accumulation input maps to which model parameter.

## Passing criteria
- Optimizer is separate flatbuffer
- Optimizer has a valid grad accumulation inputs mapped to model parameters
- Gradients are successfully updated
- Updated gradients (after X steps) are matching CPU (Golden)

pilkicTT:
With this change we can now run complete training loop on the device. The last missing part was the optimizer.

For now, only forge optimizers are supported, since the torch optimizers are not of `nn.Module` type (so we don't have a way to compile them afaik).

To run the optimizer on the device, pass the forge optimizer into the `forge.compile()` when compiling a model with trainable parameters.

The compile flow:
 - passed optimizer from the user is sent through to the autograd pass
 - the autograd constructs the optimizer graph by calling `generate_op_trace()` on the optimizer for each trainable parameter; this function creates a subgraph which defines optimizer step for particular parameter which is then merged into the main graph
 - before lowering to mlir we split the graph into multiple graphs (forward, backward, optimizer) as we did before this change
 - finally, all of the optimizer parameters are stored in the `CompiledModel` for the runtime, and the `CompiledModel` is linked to the optimizer - this enables user to call `optimizer.step()` which will in turn execute optimizer graphs for all linked models

Since we don't have a way to implement in-place updates yet, one major workaround in this change is the introducing of aliased tensor. This is done so that we can update the parameters' values after the execution of the optimizer graph. E.g. `updated_weight = weight - lr * grad`, where `updated_weight` output is aliased to the `weight` tensor, so that the runtime can swap out the original weight tensor's data with the updated ones.

Tests for compiling models with SGD, Adam, Adamw forge optimizers are added as well as e2e test for running MNIST training (with SGD optimizer) on the device is added.

Closes #176, closes #178

### DIFF
--- a/forge/csrc/passes/split_graph.cpp
+++ b/forge/csrc/passes/split_graph.cpp
@@ -4,6 +4,7 @@
 
 #include "split_graph.hpp"
 
+#include <algorithm>
 #include <memory>
 #include <utils/assert.hpp>
 #include <utils/logger.hpp>
@@ -21,41 +22,10 @@ using Graph = tt::graphlib::Graph;
 namespace tt::passes
 {
 
-bool is_parameter_node(const tt::graphlib::Node *node)
-{
-    return node->node_type() == graphlib::NodeType::kInput && node->as<graphlib::InputNode>()->is_parameter();
-}
-
-bool is_loss_input_node(const tt::graphlib::Node *node)
-{
-    return node->node_type() == graphlib::NodeType::kInput && node->as<graphlib::InputNode>()->is_loss();
-}
-
 void clone_and_add(const graphlib::Node *node, Graph *new_graph)
 {
     auto cloned_node = node->clone(node->name());
     new_graph->add_node(std::move(cloned_node), 0 /*subgraph_id=*/);
-}
-
-// 1. Clone a node from the source graph
-// 2. Add it to the dest graph
-// 3. Replicate all operand edges from the source graph to the dest graph
-void clone_and_connect_operands(const graphlib::Graph *src_graph, const graphlib::Node *src_node, Graph *dst_graph)
-{
-    auto cloned_node = src_node->clone(src_node->name());
-    dst_graph->add_node(std::move(cloned_node), 0 /*subgraph_id=*/);
-
-    auto &src_node_name = src_node->name();
-
-    for (auto operand : src_graph->operands(src_node))
-    {
-        TT_ASSERT(
-            dst_graph->has_node_with_name(operand->name()),
-            "Expected operand {} to be present in the destination graph",
-            operand->name());
-        auto dst_operand = dst_graph->get_node_by_name(operand->name());
-        dst_graph->add_edge(dst_operand, dst_graph->get_node_by_name(src_node_name));
-    }
 }
 
 bool has_users_in_bwd(const Graph *graph, const graphlib::Node *node)
@@ -66,6 +36,16 @@ bool has_users_in_bwd(const Graph *graph, const graphlib::Node *node)
         user_edges.begin(),
         user_edges.end(),
         [graph](const graphlib::Edge &edge) { return graph->node_by_id(edge.consumer_node_id)->is_backward(); });
+}
+
+bool has_users_in_opt(const Graph *graph, const graphlib::Node *node)
+{
+    auto user_edges = graph->user_data_edges(node);
+
+    return std::any_of(
+        user_edges.begin(),
+        user_edges.end(),
+        [graph](const graphlib::Edge &edge) { return graph->node_by_id(edge.consumer_node_id)->is_optimizer(); });
 }
 
 bool needs_intermediate_output(const Graph *graph, const graphlib::Node *node)
@@ -106,7 +86,26 @@ std::unique_ptr<Graph> extract_forward_graph(const Graph *graph, const std::vect
             continue;
         }
 
-        clone_and_connect_operands(graph, node, fwd_graph.get());
+        const std::string &node_name = node->name();
+
+        auto cloned_node = node->clone(node_name);
+        const graphlib::Node *new_node = fwd_graph->add_node(std::move(cloned_node), 0 /*subgraph_id=*/);
+
+        for (auto operand : graph->operands(node))
+        {
+            if (!operand->is_forward())
+            {
+                continue;
+            }
+
+            TT_ASSERT(
+                fwd_graph->has_node_with_name(operand->name()),
+                "Expected operand {} of the node {} to be present in the destination graph",
+                operand->name(),
+                node_name);
+            auto dst_operand = fwd_graph->get_node_by_name(operand->name());
+            fwd_graph->add_edge(dst_operand, new_node);
+        }
     }
 
     // Add all the module inputs to the forward graph (keeping the same order as in the original graph).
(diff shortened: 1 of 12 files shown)


forge/csrc/passes/split_graph.cpp
```diff
@@ -333,26 +332,147 @@ std::unique_ptr<Graph> extract_backward_graph(
...
+            grad_output_node->set_alias(param->as<graphlib::InputNode>());
+
+            opt_graph->add_node(std::move(grad_output_node), 0 /*subgraph_id=*/);
+            opt_module_outputs.push_back(opt_graph->get_node_by_name(param->name() + "_updated")->id());
+        }
+    }
+
+    std::vector<graphlib::NodeId> opt_module_inputs;
+
+    // All gradients computed in the backward graphs should be inputs to the optimizer graph - add them.
```
vladimirjovanovicTT, line 374: All *parameter* gradients computed in the bwd are inputs to opt, activation gradients aren't - this might well be the case here but it's not obvious to me.

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -469,3 +469,79 @@ def test_lora():
         test_loss += tt_loss(pred, target)[0]
 
     print(f"Test (total) loss: {test_loss}")
+
+
+@pytest.mark.push
+def test_e2e_device():
```
vladimirjovanovicTT, line 475: Just add '_gradient_accumulation' to the test name (or push optimizer.step() inside the batch loop).

vladimirjovanovicTT (approved):
Looks great!

forge/test/mlir/test_training.py
```diff
@@ -11,17 +11,18 @@
...
+        nn.init.xavier_uniform_(self.p)
+
+    def forward(self, x):
+        return torch.matmul(x, self.p)
 
-        def forward(self, x):
-            return torch.matmul(x, self.p)
 
+@pytest.mark.push
+def test_torch_training():
```
nvukobratTT, line 25: Should we update this test to run the optimizer on the device as well? 

Even better, I would suggest to have:
1. One test where loss and optimizer are no host
2. Second test where loss is on device, optimizer on host
3. Third test where both loss and optimizer are on device

This is good showcase how to manage each case, and can be helpful for future debugging.

forge/forge/compile.py
```diff
@@ -5,6 +5,7 @@
 from typing import Optional, List, Dict, Any, Tuple, Union
 from dataclasses import dataclass, field
 
+from forge.optimizers import Optimizer
```
nvukobratTT, line 8: Let's move this import a bit lower. 

It'll be good to strive to have unified import semantics:
- Built-in python libs
- Third-party libs
- Our libs

forge/forge/compile.py
```diff
@@ -114,7 +115,7 @@ class CompileContext:
     microbatch_size: int
     microbatch_count: int
     inputs: Union[torch.Tensor, List[torch.Tensor]]
-    optimizer: Optional[torch.optim.Optimizer] = None
+    optimizer: Optional[Union[torch.optim.Optimizer, forge.optimizers.Optimizer]] = None
     training: bool = False
```
nvukobratTT, line 119: Nit-picking, but do we need a training flag? 

If we provide loss or optimizer, it is expected to force training mode? Correct me if I'm missing something.

Ofc, this comment is not related to this change, just a question for further API design.

forge/csrc/passes/split_graph.cpp
```diff
@@ -333,26 +332,147 @@ std::unique_ptr<Graph> extract_backward_graph(
...
+    const Graph *graph, const Graph *fwd_graph, const std::vector<graphlib::Node *> &topo)
+{
+    auto opt_graph = std::make_unique<Graph>(tt::graphlib::IRLevel::IR_TT_FORGE, "optimizer");
+    opt_graph->set_training(graph->training());
+
+    // Adding all trainable params (which require gradients) from the forward graph as inputs to the optimizer graph.
+    std::vector<graphlib::NodeId> opt_module_outputs;
+    for (auto param : graph->get_parameter_nodes())
+    {
+        if (param->as<graphlib::InputNode>()->requires_grad())
```
nvukobratTT, line 345: Do we have a test covering cases where not all parameters are trainable? 

I'm mostly interested in e2e test which proves fine-tuning case with mixed required_grad options.

forge/csrc/passes/split_graph.cpp
```diff
@@ -333,26 +332,147 @@ std::unique_ptr<Graph> extract_backward_graph(
...
+            log_debug("Adding parameter node {} as input to opt graph", param->name());
+            clone_and_add(param, opt_graph.get());
+
+            // For each trainable parameter, we need to create the output node for the updated parameter (aliased to the
+            // input node). E.g. If the parameter is `w`, then the output node will be `w_updated`: `w_updated = w - lr
+            // * gradient`.
+            //
+            // The runtime will look for aliased outputs and will make sure that the appropriate tensors are updated.
+            // See `OutputNode::is_aliased_tensor()` for more details.
+            auto grad_output_node = graphlib::create_node<graphlib::OutputNode>(param->name() + "_updated");
```
nvukobratTT, line 360: Is this alias just a placeholder, or it stores copy of trainable parameters? I.e. `w` vs `w_updated`

forge/test/mlir/test_training.py
vladimirjovanovicTT: I agree, and we can have the 4th option as well (loss on host, opt on device)

vladimirjovanovicTT: This sounds like a good task for training team.

forge/forge/compile.py
vladimirjovanovicTT: A need for training (or similar) flag will be needed when we switch to two fwd passes. When doing validation during training, we will switch a flag to false (as in torch model.eval()), and turn it back on when training is resumed.

forge/csrc/passes/split_graph.cpp
nvukobratTT: Sure. I think we should have both. One simple case for mixed trainable params, and a specific one that targets the uniqueness of the LoRA case. 

For example, if a simple mix grad test fails, it's easier to debug as it has a single purpose compared to LoRA.

forge/forge/compile.py
nvukobratTT: Hmm. Valid point. Let's discuss this next week. I have a few unknowns from my side on how to do a seamless switch during training.

forge/csrc/passes/split_graph.cpp
pilkicTT: You were right, should be corrected now...

forge/csrc/passes/split_graph.cpp
pilkicTT: As explained in the description and in the comments, there is no way currently to have "in-place" updates of tensors in mlir (probably even in ttnn), i.e. every operation produces a new output.

So, for every weight that we're updating we'll have a new tensor representing new weight value. And using the introduced aliasing mechanism, the runtime can figure out that the tensor is aliased, and replace the data pointer in the original weight tensor to point to the data containing the new value.

Issue #931, opened 2024-12-18
[bug] Unexpected import from TVM to Forge


mstojkovicTT:
When loading parameters from TVM into Forge, we don't use relative paths. Instead, we rely on module imports, which depend on the `PYTHONPATH` or environment configuration. This approach can lead to unexpected results and should be avoided.

PR #932, opened 2024-12-18
fix module import in tvm_to_python

mstojkovicTT:
Instead of relying on module imports, which depend on the `PYTHONPATH` or environment configuration, we should use relative paths.

fixes: #931


pilkicTT (approved):
Thanks, much better!

PR #941, opened 2024-12-19
Support value range for tensors with default df

vbrkicTT:
Value range should not be ignored in tests when data format is not specified.

Preserve original behavior for unary, binary, matmul and softmax operators


vbrkicTT:
Check if bevavior changed for new operators

dgolubovicTT (approved):

PR #946, opened 2024-12-20
[bug] Fix for reduce avg backward op

pmarkovicTT:
In the reduce avg op, multiply op requires same dims between broadcast and constant operands, which causes `test_llama_backward.py` to fail.

I added a test for individual ops required for llama training.
https://github.com/tenstorrent/tt-thomas/issues/60

This PR also includes test to check if required dimensions for Llama are supported by mat mul op.
https://github.com/tenstorrent/tt-thomas/issues/61


forge/forge/op/eval/forge/reduce.py
```diff
@@ -321,7 +321,8 @@ def backward(type, attr, ac, operand, inputs, output, grad):
         dim = attr[0]
         size = ac.get_shape(inputs[0])[dim]
         broadcast = ac.op("broadcast", (grad,), (dim, size))
-        return ac.op("multiply", (broadcast, ac.constant(1 / size)))
+        consts = ac.tensor(torch.full(broadcast.shape.as_list(), 1 / size))
```
nvukobratTT, line 324: Remind me, this is a TTNN limitation that we expect to solve sooner or later? 

If yes, let's add a brief description above const creation outlining the main reason for doing torch.full here.

forge/test/mlir/llama/test_llama_train_ops.py
```diff
@@ -0,0 +1,124 @@
...
+from forge.verify.verify import verify
+from forge.op.loss import CrossEntropyLoss
+from forge.tensor import to_forge_tensors
+
+
+@pytest.mark.parametrize(
+    "input_shape, dim",
+    [
+        pytest.param(
+            (1, 12, 200),
```
nvukobratTT, line 20: Can we test out some "non-pretty" shapes like 7, 31, etc.

It's okay if they fail, we can mark them as xfail. But it'll be good to be aware of stability of our ops.

forge/test/mlir/llama/test_llama_train_ops.py
```diff
@@ -0,0 +1,124 @@
...
+            -1,
+        ),
+        pytest.param(
```
nvukobratTT, line 23: Also, how about checking different ranks as well? E.g. rank 4 :))

forge/test/mlir/llama/test_llama_train_ops.py
```diff
@@ -0,0 +1,124 @@
...
+
+    framework_model = MeanBwd(dim=dim)
+    framework_optimizer = torch.optim.SGD(framework_model.parameters(), lr=0.001)
+
+    compiled_model = forge.compile(framework_model, input_ids, optimizer=framework_optimizer, training=True)
+
+    verify([input_ids], framework_model, compiled_model)
+
+
+@pytest.mark.parametrize(
```
nvukobratTT, line 66: When we're not doing xfail on specific parameters, you can also use pytest params to generate combinations for you. 

E.g. 
```
feature_shapes = [8, 16, 32, ...]
@pytest.mark.parametrize("in_features", feature_shapes)
@pytest.mark.parametrize("out_features", feature_shapes)
```

Ofc, if you need a more granular approach like in current example, this one is valid as well.

forge/test/mlir/llama/test_llama_train_ops.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # Create target tensor and leave on CPU
+    target = torch.nn.functional.one_hot(torch.randint(0, 9, (12,)), num_classes=10).float()
+
+    # Forward pass (prediction) on device
+    input_ids = torch.randn((12, 3200))
+    pred = tt_model(input_ids)[0]
+    verify([input_ids], framework_model, tt_model)
+
+    tt_loss(pred, target)
```
nvukobratTT, line 118: What is the main purpose of this test? I.e. what do we want to confirm?

My main ask comes from this part as we're doing verification on the model run only. However, loss and optimizer are only run to do a functional check, but data correctness (shape, DF, values) is ignored.

forge/forge/op/eval/forge/reduce.py
pmarkovicTT: Added description for reference

forge/test/mlir/llama/test_llama_train_ops.py
pmarkovicTT: In this case, current setup makes more sense to me. If no other reasons I would leave it this way

forge/test/mlir/llama/test_llama_train_ops.py
pmarkovicTT: Here we want to make sure that required matmul dims for Llama works in both fwd and bwd passes, it's more functional/model test.

nvukobratTT (approved):

PR #951, opened 2024-12-23
update test.md docs

mstojkovicTT:
added pytest usage part to test.md doc

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -7,6 +7,7 @@
 # Project setup
 - [Building](./build.md)
 - [Testing](./test.md)
+    - [Pytest](./pytest.md)
 - [Tools](./tools.md)
 
 # Dev Notes
--- a/docs/src/pytest.md
+++ b/docs/src/pytest.md
@@ -1,0 +1,86 @@
+## Pytest
+
+Pytest is a powerful testing framework for Python that simplifies writing and executing test cases. It supports features like test discovery, fixtures, parameterized testing, and detailed assertions. For more details, visit the official [Pytest Documentation](https://docs.pytest.org/en/latest/).
+
+
+### Testing with multiple input sets
+
+The `@pytest.mark.parametrize` decorator allows you to run a single test function with multiple sets of inputs.
+
+#### Example
+
+```python
+@pytest.mark.parametrize("arg1, arg2, expected", [
+    (1, 2, 3),
+    (2, 3, 5),
+    (3, 5, 8),
+])
+def test_addition(arg1, arg2, expected):
+    assert arg1 + arg2 == expected
+```
+
+#### Explanation
+- This is particularly useful for testing a function with various combinations of arguments
+
+### Marking specific parameters
+
+You can use `pytest.param` to mark specific parameter combinations with additional metadata, such as expected failures (`xfail`).
+
+#### Example
+
+```python
+@pytest.mark.parametrize("inputs", [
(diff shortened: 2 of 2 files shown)


docs/src/test.md
```diff
@@ -46,6 +46,96 @@ pytest -svv forge/test/mlir/test_ops.py::test_add
...
+    (1, 2, 3),
+    (2, 3, 5),
+    (3, 5, 8),
+])
+def test_addition(arg1, arg2, expected):
+    assert arg1 + arg2 == expected
+```
+
+### Explanation
+- The first argument is a comma-separated string of parameter names.
```
nvukobratTT, line 66: We don't need to explain how pytest internal works. For that, we can mention in single sentence what is the purpose of it and why we use it, and for other details we can link to their original documentation.

Main purpose of documenting this on our side is to outline how we use pytest, and what are best practices on Forge-FE for N reasons

docs/src/test.md
```diff
@@ -46,6 +46,96 @@ pytest -svv forge/test/mlir/test_ops.py::test_add
...
+- The second argument is a list of tuples, each containing values for the parameters.
+- Pytest runs the test function once for each tuple, substituting the parameter values.
+
+## Marking Specific Parameters with `pytest.param`
```
nvukobratTT, line 70: For example, this section can focus on explaining when, how and why we use xfail. Similar to skip

docs/src/test.md
```diff
@@ -46,6 +46,96 @@ pytest -svv forge/test/mlir/test_ops.py::test_add
...
+        ((1, 2, 3), (4, 5, 6)), marks=pytest.mark.xfail(reason="reason"))
+])
+```
+
+#### Explanation
+- `pytest.param` allows you to add specific markers or metadata to individual parameter sets.
+- In this example, the first parameter combination is marked as `xfail` with a reason provided, indicating it is expected to fail.
+- This is useful when only some parameter sets are failing or not working correctly.
+
+## Skipping Tests with `@pytest.mark.skip`
```
nvukobratTT, line 88: As mentioned, this one is good! However, a detailed explanation is probably redundant. We can show an example, and refer to other details in the main pytset documentation.

docs/src/test.md
```diff
@@ -46,6 +46,96 @@ pytest -svv forge/test/mlir/test_ops.py::test_add
...
+    assert some_function() == "expected result"
+```
+
+### Explanation
+- The `reason` parameter explains why the test is being skipped.
+- Skipped tests will appear in the pytest output with the reason.
+- Skipping tests is particularly useful when a test is causing crashes (e.g., segmentation faults) or breaking the CI pipeline.
+
+
+## Marking Tests as Expected to Fail with `@pytest.mark.xfail`
```
nvukobratTT, line 106: For xfail, it'll be good to mention what not to do as well. E.g. not using as part of test, and why :))

docs/src/test.md
```diff
@@ -46,6 +46,93 @@ pytest -svv forge/test/mlir/test_ops.py::test_add
 
 > - The `-svv` flag is optional and used to display more information about the test run.
 
+## Pytest practices
```
nvukobratTT, line 49: Looks good! 

What do you think about moving it to as sub-section of test.md file? E.g. creating a new file, and nesting it under Testing section. 
image

E.g.
```
...
4. Testing
    4.1. Pytest
...
```

nvukobratTT (approved):

PR #957, opened 2024-12-24
Add support for generating nightly test in model analysis

### LINKED ISSUE #761: Optimize the model analysis script and generate models ops test

chandrasekaranpradeep:
Fixes #761

### DIFF
--- a/forge/forge/module.py
+++ b/forge/forge/module.py
@@ -722,7 +722,14 @@ class ForgeModule(Module):
         else:
             parameter._set_auto_name(name)
 
-    def add_constant(self, name: str, prepend_name: bool = False, shape: Tuple[int] = None):
+    def add_constant(
+        self,
+        name: str,
+        prepend_name: bool = False,
+        shape: Tuple[int] = None,
+        dtype: torch.dtype = torch.float32,
+        use_random_value: bool = False,
+    ):
         """
         Adds a new constant.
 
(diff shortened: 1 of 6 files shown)


chandrasekaranpradeep:
The changes are uplifted into main in another PR so closing the PR

PR #960, opened 2024-12-24
Documentation for verify()

vkovinicTT:
Added documentation for the `verify()` function and `VerifyConfig` class. This includes an overview of verification steps, examples of usage, and details on configurable checkers such as `AutomaticValueChecker`, `AllCloseValueChecker` and `FullValueChecker`. Additionally, provided information for enabling/disabling verification features through the `VerifyConfig`.


docs/src/SUMMARY.md
```diff
@@ -7,6 +7,7 @@
 # Project setup
 - [Building](./build.md)
 - [Testing](./test.md)
+- [Verification](./verification.md)
```
pilkicTT, line 10: imo this should be in the dev notes section, something like `Verification in tests`

docs/src/verification.md
```diff
@@ -0,0 +1,164 @@
...
+- Handles forward passes for both framework and compiled models
+- Compares results using a combination of comparison methods
+- Supports customization through the `VerifyConfig` class.
+
+**Example of usage**
+```python
+def test_add():
+
+	class Add(nn.Module):
+		def __init__(self):
```
pilkicTT, line 24: fix indent (4 spaces)

docs/src/verification.md
```diff
@@ -0,0 +1,164 @@
...
+- `AutomaticValueChecker` **(default)**
+- `AllCloseValueChecker`
+- `FullValueChecker`
+
+For more information about **Checkers** you can look at `forge/forge/verify/value_checkers.py`.
+
+<br/>
+<br/>
+
+<img src="imgs/checkers/checkers.svg" alt="Checkers" style="width: 1200%;">
```
pilkicTT, line 107: let's see what others say, but to me the picture doesn't add much value and makes the doc harder to edit/update

docs/src/verification.md
```diff
@@ -0,0 +1,164 @@
...
+	co_out = compiled_model(*inputs)
+```
+
+## Verify Config Overview
+
+If `VerifyConfig` isn't passed as a param, default one will be used. Currently through `VerifyConfig` you can disable/enable:
+| Feature                           | Enabled (default) |
+|-----------------------------------|:-----------------:|
+| Verification as a method          | `True`            |
+| Output size check                 | `True`            |
```
pilkicTT, line 67: what is the output size?

pilkicTT (changes requested):
thanks for adding the docs!

ps. please update the PR title

docs/src/verification.md
vkovinicTT: for some reason github renders it to 8 spaces, even though it's 4 spaces for me locally.

docs/src/verification.md
vkovinicTT: @nvukobratTT what do you think?

docs/src/verification.md
vkovinicTT: I've changed table a bit, hopefully it's clearer now.

docs/src/verification.md
pilkicTT: you should use spaces, not tabs

docs/src/verification.md
pilkicTT: what is the output length? 😄 is it the number of output tensors?

vkovinicTT: hahah yes (I will put number of output tensors instead)

docs/src/verification.md
nvukobratTT: I'm okay with using images and they can be really useful as part of documentaion. However, it's important that those can be edit/updated over time. 

That said, I would opt out for using ExcallyDraw to create this and similar diagrams, and next to the image also store exported ED file which we can load and edit in the future.

docs/src/verification.md
```diff
@@ -0,0 +1,166 @@
+# Verification in tests
+
+## General Overview
+
+When comparing our `compiled_model` with the `framework_model` (e.g., `PyTorch`), we aim to verify whether the output from the `compiled_model` is sufficiently similar to the output from the `framework_model` (where required degree of similarity is configurable).
```
nvukobratTT, line 5: For the introduction section, let's use common language instead of code-specific names (e.g. instead of `compiled_model`, let's use compiled model). 

Using code-specific names is beneficial below examples. In other case, I would more prefer to have common language :))

nvukobratTT, line 5: Let's also mention what the framework model is. E.g. model that is run on host/cpu.

docs/src/verification.md
```diff
@@ -0,0 +1,166 @@
...
+
+So generally we want to perform the following steps:
+1. Create a framework model
+2. Run a forward pass through the framework model
+3. Compile the framework model using `Forge`
+4. Run a forward pass through the compiled model
+5. Compare the outputs.
```
nvukobratTT, line 12: Nit picking, but let's put `.` on every point, or remove it :))

docs/src/verification.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Verify Config Overview
+
+If `VerifyConfig` isn't passed as a param, default one will be used. Currently through `VerifyConfig` you can disable/enable:
+
+| Feature                           | Name                | Enabled (default) |
+|-----------------------------------|---------------------|:-----------------:|
+| Verification as a method          | `enabled`           | `True`            |
+| Number of output tensors check    | `verify_size`       | `True`            |
+| Output type check                 | `verify_dtype`      | `True`            |
```
nvukobratTT, line 69: Let's be more precise. E.g. output data format/type check

docs/src/verification.md
```diff
@@ -0,0 +1,166 @@
...
+| Output shape check                | `verify_shape`      | `True`            |
+
+
+For more information about `VerifyConfig` you can check `forge/forge/verify/config.py`.
+
+**Example of usage**
+```python
```
nvukobratTT, line 76: Maybe we can keep following examples a bit shorter. E.g. now that we have full example above, we can do something like this:
```py
    ...

    fw_out = framework_model(*inputs)
    co_out = compiled_model(*inputs)

    verify(inputs, framework_model, compiled_model, VerifyConfig(verify_dtype=False))
```

nvukobratTT (approved):
LGTM!

PR #966, opened 2024-12-25
[CI] Reduce logger verbosity for test runs #921

### LINKED ISSUE #921: Reduce logging in nightly run
Nightly runs have verbose logging that contains little useful information.
Github jobs have a problem displaying these logs

```
This step has been truncated due to its large size. Download the full logs from the  menu once the workflow run has completed. 
```

The current log file is 160 MB ( 9MB compressed ) and is hard to read even after downloading.

Log example 

https://productionresultssa5.blob.core.windows.net/actions-results/d35ac2d0-3a16-4bcf-8b9b-e20a42bcbba9/workflow-job-run-aa1574ee-da86-5472-34f7-65348102b899/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-12-17T12%3A01%3A51Z&sig=L4G60xTIoV%2FrZlbKIvcIAq9UKTSVvs9JDJnnbgdKE5Y%3D&ske=2024-12-17T23%3A14%3A10Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-12-17T11%3A14%3A10Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-11-04&sp=r&spr=https&sr=b&st=2024-12-17T11%3A51%3A46Z&sv=2024-11-04

nvukobratTT:
Fix #921

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -143,8 +143,13 @@ jobs:
 
     - name: Run Test
       env:
+        # Setup HuggingFace
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
         HF_HUB_DISABLE_PROGRESS_BARS: 1
+
+        # Configure logging levels for the test
+        LOGGER_LEVEL: INFO
+        LOGURU_LEVEL: INFO
       shell: bash
       run: |
         source env/activate


.github/workflows/build-and-test.yml
```diff
@@ -143,8 +143,13 @@ jobs:
 
     - name: Run Test
       env:
+        # Setup HuggingFace
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
         HF_HUB_DISABLE_PROGRESS_BARS: 1
+
+        # Configure logging levels for the test
+        LOGGER_LEVEL: INFO
+        LOGURU_LEVEL: INFO
```
pilkicTT, line 152: Isn't this the default logging level?

pilkicTT (changes requested):
I think we should change the default `-svv` in `pytest.ini` to reduce log size.

However, that also has negative consequences on debugging failures (since the generated TTIR and TTNNIR won't be logged).

nvukobratTT:
> I think we should change the default `-svv` in `pytest.ini` to reduce log size.
> 
> However, that also has negative consequences on debugging failures (since the generated TTIR and TTNNIR won't be logged).

Valid point. However, having all these details as part of log make it hard to interpret on daily basis (e.g. tracking and bisecting nightly run).

In order to preserve debugability, and still keep TT/TTNN IRs, maybe we should store them for each run (already existing feature) and just upload them as additional set of artifacts. @pilkicTT @vmilosevic what do you think guys?

PR #975, opened 2024-12-27
Fix model analysis weekly pipeline failures

chandrasekaranpradeep:
The PR will fix below issues found in the Model Analysis Weekly pipeline - [Link](https://github.com/tenstorrent/tt-forge-fe/actions/runs/12482523195)
   1. Unique Op tests are not generated for models like Llama, Gemma and mistral because of the Huggingface access token issue so adding the HF_Token as env in the model analysis weekly yml 
   2. Update the exception rule by checking the failed test logs of the [unknown compiler component or N/A ](https://github.com/tenstorrent/tt-forge-fe/blob/main/model_analysis_docs/ModelsInfo.md)
   3. Add the module_name for the stereo model and update the module_name for the mistral and clip model in the compile function

### DIFF
--- a/.github/workflows/model-analysis-weekly.yml
+++ b/.github/workflows/model-analysis-weekly.yml
@@ -74,6 +74,9 @@ jobs:
           cmake --build ${{ steps.strings.outputs.build-output-dir }}
 
       - name: Run Model Analysis Script
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+          HF_HUB_DISABLE_PROGRESS_BARS: 1
         shell: bash
         run: |
           source env/activate
--- a/forge/test/models/pytorch/audio/stereo/test_stereo.py
+++ b/forge/test/models/pytorch/audio/stereo/test_stereo.py
@@ -30,6 +30,8 @@ def test_stereo(variant):
     input_ids, attn_mask, decoder_input_ids = load_inputs(framework_model, processor)
     inputs = [input_ids, attn_mask, decoder_input_ids]
 
-    compiled_model = forge.compile(framework_model, sample_inputs=inputs)
+    compiled_model = forge.compile(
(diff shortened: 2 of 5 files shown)


nvukobratTT (approved):
Looks good! 

Once this one is merged, let's kick one more run right away to complete over the weekends :))

chandrasekaranpradeep:
> Once this one is merged, let's kick one more run right away to complete over the weekends :))

FYI, there is nightly scheduled on serbia time: saturday at 12:00 AM
Would you like to trigger it after this PR merge and cancel the nightly scheduled?

nvukobratTT:
> > Once this one is merged, let's kick one more run right away to complete over the weekends :))
> 
> FYI, there is nightly scheduled on serbia time: saturday at 12:00 AM Would you like to trigger it after this PR merge and cancel the nightly scheduled?

If that si the case, than there is no need to run one manually. Let's wait for regular one to run over the weekend :))

chandrasekaranpradeep:
> PCC issues for E2E run are treated as FFE errors. I think in most of the cases right now this is a consequence of TTNN. So let's move it into that group. Once we fix our Golden checks, we can have a customer error separating the two.

Moved the pcc issues exception rule from Forge Compiler Component to TT-Metal Compiler Component.

> When we have unknown errors, let's subtract that percentage from all 3 component passing score results. This is mostly to emphasize that when passing rate is 99%, that it's not actually that close if we have 20% in N/A field. However, this isn't a blocker for this PR.

The unknown compiler component percentage is subtracted from the other compiler components pass percentage.
[Compiler Component Pass/fail updater](https://github.com/tenstorrent/tt-forge-fe/blob/a0f710eec90287b50207115092e8491a35fb04b2/scripts/model_analysis.py#L398C1-L419C73)
[Percentage Calculation for the compiler components](https://github.com/tenstorrent/tt-forge-fe/blob/a0f710eec90287b50207115092e8491a35fb04b2/scripts/model_analysis.py#L992C1-L1007C117)

PR #976, opened 2024-12-27
Different types of indexing in pytorch

vkovinicTT:
Exploring different types of indexing in pytorch and looking at what is supported through `tvm`/`forge-fe`/`mlir`/`ttnn`. More info about results of testing can be found [here](https://docs.google.com/document/d/1SoMZWKsplNIRXx01HDeWQWJ2NXhyQ0VHYZ9o7Pxr2h8/edit?usp=sharing).


forge/test/mlir/indexing/test_advanced_indexing.py
```diff
@@ -0,0 +1,267 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
```
nvukobratTT, line 1: Can we have all of these as part of the `test_ops.py` file? 

My reasoning: We want to eventually move all op tests to be part of the testing framework (operators directory). Therefore, these tests will be deprecated one by one sooner or later. Consequently, I wouldn't impose additional structure, and keep tests added by engineers in multiple places, to make the organizational side a bit easier and keep these tests in one place for search and removal one by one. 

Ofc, I would like to hear other opinions as well.

forge/test/mlir/indexing/test_advanced_indexing.py
```diff
@@ -0,0 +1,267 @@
...
+            (torch.arange(10, dtype=torch.float32).reshape(2, 5), torch.tensor([0, 2, 8, -1])),
+            marks=pytest.mark.xfail(reason="IndexError: index 2 is out of bounds for dimension 0 with size 2"),
+        ),
+        pytest.param(
+            (torch.arange(27, dtype=torch.float32).reshape(3, 3, 3), torch.tensor([0, 13, -1])),
+            marks=pytest.mark.xfail(reason="IndexError: index 13 is out of bounds for dimension 0 with size 3"),
+        ),
+    ],
+)
+def test_indexing_with_torch_take(tensor_and_indices):
```
nvukobratTT, line 25: Regarding test naming, let's keep it simple without redundant decorations. E.g. `test_take`, `test_index_add`, etc.

There is no need to have indexing, torch, and similar decorations as they will be inferred from the path or file name.

forge/test/mlir/indexing/test_advanced_indexing.py
```diff
@@ -0,0 +1,267 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
```
nvukobratTT, line 3: Additionally, can we rename a `test_ops.py` into the  `test_ops_pt.py`?

nvukobratTT (commented):
One general comment (not ra equest for a change). If you're marking the whole test as xfail, or just leave it as passing, you can use pytest parameters in a way to automatically test out all combinations, instead of you needing to outline them. E.g.
```py
@pytest.mark.parametrize("shape", [(1, 32, 64, 64), (32, 64, 64), (64, 64)])
@pytest.mark.parametrize("dim", [-1, -2])
@pytest.mark.parametrize("begin", [0, 16])
@pytest.mark.parametrize("length", [4, 16])
@pytest.mark.parametrize("stride", [16, 32])

def test_select(shape, dim, begin, length, stride):
    ...
```

This way, pytest will run all combinations from given parameters

forge/test/mlir/indexing/test_advanced_indexing.py
nvukobratTT: Yes, it will be. 

However, in order to have a easy to follow structure, we have 2 main options:
1. Move all into a single file.
2. Restructure the current test_ops.py file, and split it into multiple files based on the structure we have in the  `operations` directory.

Keeping something in between will have cons from both approaches, which is in my opinion worse than going with 1. point. Ofc, the best approach is to go with the 2. option and go the extra mile, but it will require a bit more time from your side to push changes unrelated to indexing only. 

Let me know your thoughts :))

vkovinicTT: If 2. option is the best, I will go with that one, ofc if it's okay with you.

nvukobratTT: Ofc! 💪

vkovinicTT:
I've also reworked some tests in `test_ops.py` before split, since they were `XPASS`ing, so now everything should be fine.

nvukobratTT (approved):
Looks good! 

Before merging into main, let's get feedback from @vbrkicTT and @vobojevicTT as well. :))

nvukobratTT:
@vbrkicTT , @vobojevicTT seems like PR was set on auto-merge after approval. Nothing major, feel free to leave comments if needed, and we can do changes in followup PR if needed :))

Issue #663, opened 2024-11-09
[Model Analysis] Document analysis workflow


nvukobratTT:
Create new doc section that outlines main purpose of this logic, as well as how to use it in order to generate desired results.

chandrasekaranpradeep:
While documenting the generate models ops test, please add the below details:

1) The ForgeModules present in the generated  models ops test doesn't use actual parameter/constant tensor values from the model parameter/buffers by using the process_framework_parameter function. Instead, it will generate random tensor based upon the constant/parameter tensor shapes and dtypes.

2) There aren't 100% 1:1 mapping to models. E.g. when we run full model, some optimization pass can change ops shapes, insert new ones, fuse some ops, etc.

PR #998, opened 2025-01-02
Add support for generating models ops test

chandrasekaranpradeep:
1. **Restructure the model analysis script:**

Fixes [#1001](https://github.com/tenstorrent/tt-forge-fe/issues/1001)

Instead of having multiple class(i.e MarkDownWriter, MatchingExceptionRule, ModelVariantInfo, etc.) and functions in single python file, created python package named model_analysis and declared the function and class in separate file. 

Eg: MarkDownWriter class is used for creating and writing markdown files which is included in markdown.py file and common_failure_matching_rules_list has separate python module named expcetion_rules.py file
     
     
2. **Created a script for generating models ops test from the unique ops configuration extracted all the models present in the `forge/test/models` directory.**

Fixes [#874](https://github.com/tenstorrent/tt-forge-fe/issues/761)

 Workflow:
    1. Collect all the test that doesn't contain skip_model_analysis marker in the directory path specified the user 
     eg: `forge/test/models`.
    2. Run all the collected tests to extract the unique ops configuration and export the model unique ops configuration as excel and metadata json file 
    Note: It will doesn't generate unique op test like model analysis pipeline
    3. After extracting unique op configuration for the all test, then try to extract the unique ops configuration across all the test (i.e model variants) 
    4. By using extracted unique ops configuration across all the model variants, create models ops test with forge module in the directory path specified by the user.
    5. Black formating and spdx headers are also automatically done in the generated nightly/push test
    
**Note:** 
    The ForgeModules present in the generated models ops test doesn't use actual parameter/constant tensor values from the model parameter/buffers by using the process_framework_parameter function. Instead, it will generate random tensor based upon the constant/parameter tensor shapes and dtypes which is done inside the test function.
Screenshot 2025-01-07 at 6 32 35 PM

    
3. **Breakdown the Model Analysis Weekly workflow:**

Fixes [#1002](https://github.com/tenstorrent/tt-forge-fe/issues/1002)
     1. model-analysis.yml -> Common yml file used for running the model analysis pipeline for markdown generation and models ops test generation.
     2. model-analysis-weekly.yml -> Workflow used for triggering the Model Analysis Workflow(i.e model-analysis.yml) for running the model analysis and markdown files generation.
     3. model-analysis-config.sh -> shell script containing the script and PR configuration/environmental variables for the markdown generation and model test generation.
              
The generated models ops test PR for albert model - https://github.com/tenstorrent/tt-forge-fe/pull/1014


forge/forge/module.py
```diff
@@ -733,6 +740,15 @@ def add_constant(self, name: str, prepend_name: bool = False, shape: Tuple[int]
 
         prepend_name: Bool
             Whether to prepend module name to constant name
+
+        shape: Tuple[int]
```
ashokkumarkannan1, line 744: Same here

forge/forge/parameter.py
```diff
@@ -49,6 +50,9 @@ def __init__(
 
         dev_data_format: DataFormat, optional
             If set, forces the data type on device. If not provided, the closest type to given value will be used.
+
+        use_random_value: bool, optional
```
ashokkumarkannan1, line 54: remove optional, Just `use_random_value: bool` is good

forge/forge/python_codegen.py
```diff
@@ -1024,14 +1036,18 @@ def write_model_parameter_function(self, param_file_name, named_params_file_name
     def write_pytest_function(
         self,
         forge_module_names: List[str],
-        framework: str,
         pytest_input_shapes_and_dtypes_list: List[List[Tuple]],
+        markers: Optional[List[str]] = [],
+        module_metadata: Optional[Dict[str, Any]] = {},
+        pytest_metadata_list: Optional[List[Dict[str, Any]]] = [],
+        use_ids_function: Optional[bool] = False,
```
ashokkumarkannan1, line 1043: Pleases use Optional only when `None` is one of the possible values

forge/forge/python_codegen.py
```diff
@@ -1041,28 +1057,67 @@ def write_pytest_function(
 
         Args:
             forge_module_names (List[str]): List of names of the modules to be tested, each corresponding to a forge module.
-            framework (str): The name of the framework under which the model is to be tested (e.g., "pytorch").
             pytest_input_shapes_and_dtypes_list (List[List[Tuple]]): A list of input shapes and corresponding data types for each module. Each tuple contains the shape and dtype to be tested.
+            markers (Optional[List[str]]): A list of pytest markers that will be added above the test function. Defaults to an empty list.
+            module_metadata (Optional[Dict[str, str]]): A dictionary containing metadata about the test function. Each key-value pair represents a metadata property name and its corresponding value, which will be recorded using the `record_property` pytest fixtures. Defaults to an empty dictionary.
+            pytest_metadata_list (Optional[List[Dict[str, str]]]): A list of dictionaries containing metadata for each pytest parameter.
```
ashokkumarkannan1, line 1063: same here

forge/forge/python_codegen.py
```diff
@@ -1041,28 +1057,67 @@ def write_pytest_function(
...
+            use_ids_function: If set, the forge module name and shapes and dtyes will used as id for the pytest parameter.
```
ashokkumarkannan1, line 1064: Type is missing?

forge/forge/tensor.py
```diff
@@ -274,6 +274,23 @@ def create_from_torch(
         """
         return TensorFromPytorch(torch_tensor, dev_data_format, constant)
 
+    @classmethod
+    def create_torch_tensor(
+        cls,
+        shape: Union[List, Tuple, torch.Size],
```
ashokkumarkannan1, line 280: What is the type of elements inside list or tuple?

forge/forge/tensor.py
```diff
@@ -274,6 +274,23 @@ def create_from_torch(
...
+        dtype: Optional[torch.dtype] = None,
+        integer_tensor_high_value: int = 1000,
```
ashokkumarkannan1, line 282: can we name it as `max_int`?

forge/forge/tensor.py
```diff
@@ -274,6 +274,23 @@ def create_from_torch(
...
+    ) -> torch.Tensor:
+
+        if dtype in [torch.float16, torch.bfloat16, torch.float32, torch.float64]:
+            torch_tensor = torch.rand(shape, dtype=dtype)
+        elif dtype in [torch.int8, torch.int, torch.int16, torch.int32, torch.int64]:
```
ashokkumarkannan1, line 287: torch.int and torch.int32 is same. Let's have only one of them,

forge/forge/tvm_unique_op_generation.py
```diff
@@ -372,6 +372,7 @@ def create_unique_operations(
         ops: Dict[int, Operation],
         named_parameters: Dict[str, torch.Tensor],
         node_name_to_node_type: Optional[Dict[str, NodeType]] = None,
+        use_constant_value: Optional[bool] = True,
```
ashokkumarkannan1, line 375: Do we need Optional here?

forge/forge/tvm_unique_op_generation.py
```diff
@@ -423,6 +426,38 @@ def create_unique_operations(
...
+                                operand_types, operand_shapes, operand_names
+                            )
+                        ]
+                    )
+                    operation_info["Operand_Types"] = str(
+                        [NodeType.to_json(operand_type) for operand_type in operand_types]
+                    )
+                    operation_info["Operand_Dtypes"] = str(operand_dtypes)
+                    operation_info["Args"] = str(args)
+                    operation_info["Testfile"] = ""
```
ashokkumarkannan1, line 457: What is this Testfile? Can you pls some comments here?

ashokkumarkannan1 (changes requested):

forge/forge/module.py
```diff
@@ -722,7 +722,14 @@ def add_parameter(self, name: str, parameter: Parameter, prepend_name: bool = Fa
         else:
             parameter._set_auto_name(name)
 
-    def add_constant(self, name: str, prepend_name: bool = False, shape: Tuple[int] = None):
+    def add_constant(
+        self,
+        name: str,
+        prepend_name: bool = False,
+        shape: Tuple[int] = None,
```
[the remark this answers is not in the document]
chandrasekaranpradeep: It will always receive tuple of ints 
ref: https://github.com/tenstorrent/tt-forge-fe/blob/288aafafbc9dc039fb75a659c2e9bc0b398929a1/forge/forge/python_codegen.py#L217C1-L217C36

forge/forge/module.py
chandrasekaranpradeep: It will always receive tuple of ints
ref: https://github.com/tenstorrent/tt-forge-fe/blob/288aafafbc9dc039fb75a659c2e9bc0b398929a1/forge/forge/python_codegen.py#L217C1-L217C36
No type change required

forge/forge/parameter.py
chandrasekaranpradeep: Reference: https://github.com/huggingface/transformers/blob/42865860ec6dc135972d9555753cb7ee17f51fb4/src/transformers/models/llama/modeling_llama.py#L906C1-L915C44

chandrasekaranpradeep: We can use optional with bool
Hugging face reference: https://github.com/huggingface/transformers/blob/42865860ec6dc135972d9555753cb7ee17f51fb4/src/transformers/models/llama/modeling_llama.py#L328C1-L329C43

forge/forge/python_codegen.py
chandrasekaranpradeep: We can use optional with bool
Hugging face reference: https://github.com/huggingface/transformers/blob/42865860ec6dc135972d9555753cb7ee17f51fb4/src/transformers/models/llama/modeling_llama.py#L328C1-L329C43

forge/forge/python_codegen.py
chandrasekaranpradeep: Will add it

forge/forge/python_codegen.py
chandrasekaranpradeep: Will update it

forge/forge/tensor.py
chandrasekaranpradeep: will update it

forge/forge/tensor.py
chandrasekaranpradeep: User can create torch.int32 tensor if the pass torch_dtype parameter as torch.int or torch.int32. Thats why there is 2 option being added

forge/forge/tvm_unique_op_generation.py
chandrasekaranpradeep: Addressed above

forge/forge/tvm_unique_op_generation.py
chandrasekaranpradeep: Testfile contains path to the unique operation test for each unique operation config

forge/forge/python_codegen.py
ashokk-cerebras: Please refer the official doc 
https://docs.python.org/3/library/typing.html#typing.Optional

forge/forge/module.py
ashokkumarkannan1: Please refer the official doc https://docs.python.org/3/library/typing.html#typing.Optional

forge/forge/python_codegen.py
ashokkumarkannan1: Please refer the official doc https://docs.python.org/3/library/typing.html#typing.Optional

forge/forge/module.py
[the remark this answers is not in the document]
ashokkumarkannan1: Please refer https://docs.python.org/3/library/typing.html#annotating-tuples

forge/forge/module.py
ashokkumarkannan1: https://docs.python.org/3/library/typing.html#annotating-tuples

forge/forge/module.py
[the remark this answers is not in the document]
chandrasekaranpradeep: It doesn't throws any error even if there is more than one element in tuple
Screenshot 2025-01-03 at 12 24 54 PM

nvukobratTT:
Great job! Few general comment based on PR description. All of these should be part of the future changes, so feel free to create separate issues to tackle with them @chandrasekaranpradeep :))

> 1. Collect all the pytest that has model_analysis pytest marker in the directory path specified the user
eg: forge/test/models/pytorch.

Going forward, idea is to try to run all models (for now Pytorch). That said, let's update logic to mark only models that we don't wan't to have as a part of model analysis logic. E.g. collect all pytest from specified directory, while ignoring ones that have skim_model_analysis marker (or something similar) :))

> The ForgeModules present in the generated nightly/push test doesn't use actual parameter/constant tensor values from the model parameter/buffers by using the process_framework_parameter function. Instead, it will generate random tensor based upon the constant/parameter tensor shapes and dtypes which is done by setting the use_random_value argument to True in the Parameter class in forge/forge/parameter.py file and add_constant method in the ForgeModule in forge/forge/module.py file.

Make sure to outline these details when you start with documenting these workflows. This is one of the major notes that can influence our results more or less. 

Also, let's mention that these aren't 100% 1:1 mapping to models. E.g. when we run full model, some optimization pass can change ops shapes, insert new ones, fuse some ops, etc.

> The nightly/push test generated by the generate-nightly-test.yml for the albert model
As we already synced offline, let's update "nightly" naming to better fit workflow changes :))

.github/workflows/model-analysis.yml
```diff
@@ -0,0 +1,161 @@
...
+      - name: Set reusable strings
+        id: strings
+        env:
+          NIGHTLY_GENERATION: ${{ inputs.nightly_generation }}
+        shell: bash
+        run: |
+          echo "work-dir=$(pwd)" >> "$GITHUB_OUTPUT"
+          echo "build-output-dir=$(pwd)/build" >> "$GITHUB_OUTPUT"
+          if [ "$NIGHTLY_GENERATION" = "true" ]; then
+            echo "branch-name=generate_nightly_tests" >> "$GITHUB_OUTPUT"
```
nvukobratTT, line 56: Let's extract these configurations to be more visible. E.g. env of this job, or env vars as part of run section. 

My main point is to make it easy to spot and configure on the fly. Plus, to have it clearly outline as we'll refactor these in the future to align across all workflow files.

.github/workflows/model-analysis.yml
```diff
@@ -0,0 +1,161 @@
...
+          -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+          cmake --build ${{ steps.strings.outputs.build-output-dir }}
+
+      - name: Run Model Analysis Script
+        if: ${{ !inputs.nightly_generation }}
+        shell: bash
+        run: |
+          source env/activate
+          apt-get update
+          apt install -y libgl1 libglx-mesa0
```
nvukobratTT, line 110: This is probably good candidate for our docker images if some models/libs are dependant on those. Ofc, let's just open a reminder issues, this one is not a blocker.

.github/workflows/model-analysis.yml
```diff
@@ -0,0 +1,161 @@
...
+          set -o pipefail # Ensures that the exit code reflects the first command that fails
+          python scripts/model_analysis/run_analysis_and_generate_md_files.py \
+            --test_directory_or_file_path forge/test/models/pytorch \
```
nvukobratTT, line 113: Similar env comment as for one above. Additionally, in this case, maybe we can put top-level env for this file. E.g. some of the configurations are duplicated across multiple jobs. Maybe we should think about unifying them as a single configuration string, and only change on single place in this file when needed

forge/forge/module.py
```diff
@@ -722,7 +722,14 @@ def add_parameter(self, name: str, parameter: Parameter, prepend_name: bool = Fa
...
+        dtype: torch.dtype = torch.float32,
+        use_random_value: bool = False,
```
nvukobratTT, line 731: Can we generate these on the fly within the test instead of extending the `add_constant` function logic? 

I think this logic will only be used for testing similar to this one. That said, I think it'll be better just to generate these and assign them in the testing logic as is. 

Please, if you see a more general use case to keep it, share it with us here. Also, if my requests complicates the logic quite a bit, let's discuss it here and determine what is the best path going forward.

forge/forge/tvm_unique_op_generation.py
```diff
@@ -512,25 +546,304 @@ def get_param_const(name):
...
+    """
+
+    # Iterate over the unique operations dictonary after sorting it by operation name.
+    for forge_op_function_name in sorted(unique_operations):
 
         # Extract operation name from forge op function name
         op_name = forge_op_function_name.split(".")[-1].lower()
 
+        # Dictionary containing module metadata which will used in the test function record_property fixture.
+        module_metadata = {
```
nvukobratTT, line 832: For now, let's not use record_property fixture in generated op tests. Let's keep the logic, but not generate for now :))

.github/workflows/model-analysis.yml
```diff
@@ -0,0 +1,161 @@
...
+jobs:
+
+  docker-build:
+    uses: ./.github/workflows/build-image.yml
+    secrets: inherit
+
+  model-analysis:
+    needs: docker-build
+    runs-on: runner
+    timeout-minutes: 10080 # Set job execution time to 7 days(default: 6 hours)
```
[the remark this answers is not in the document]
chandrasekaranpradeep: Sure Nikola

chandrasekaranpradeep:
> Going forward, idea is to try to run all models (for now Pytorch). That said, let's update logic to mark only models that we don't wan't to have as a part of model analysis logic. E.g. collect all pytest from specified directory, while ignoring ones that have skim_model_analysis marker (or something similar) :))

I have added **model_analysis** marker in all the test inside the `forge/test/models/pytorch` path and I am collecting the test from this `forge/test/models/pytorch` path. Even if change path of collecting test from `forge/test/models/pytorch` path to `forge/test/models` path, it will only collect the test that has model_analysis marker, if we don't want to run the model analysis script for particular model test, we can remove the **model_analysis** marker and it won't be included in the collected test. There is no need for new marker for skipping the test for model analysis.

chandrasekaranpradeep:
> Make sure to outline these details when you start with documenting these workflows. This is one of the major notes that can influence our results more or less.
> 
> Also, let's mention that these aren't 100% 1:1 mapping to models. E.g. when we run full model, some optimization pass can change ops shapes, insert new ones, fuse some ops, etc.

Will add these details as points in model analysis documentation

.github/workflows/generate-nightly-tests.yml
```diff
@@ -0,0 +1,11 @@
+name: Generate Nightly Tests
+
+on:
+  workflow_dispatch:
```
vmilosevic, line 4: Why do you need this workflow, what does it add?
model-analysis.yml already has workflow_dispatch, you can just start that one with 
nightly_generation: true and have the same effect

.github/workflows/model-analysis.yml
```diff
@@ -0,0 +1,161 @@
...
+            --markdown_directory_path ./model_analysis_docs \
+            --unique_ops_output_directory_path ./models_unique_ops_output \
+            2>&1 | tee script.log
+
+      - name: Generate nightly test
+        if: ${{ inputs.nightly_generation }}
+        shell: bash
+        run: |
+          source env/activate
+          apt-get update
```
vmilosevic, line 124: You already have apt install in previous step, no need to do it again in the same job.
All job steps execute on the same runner, so you only need to install once.

.github/workflows/generate-nightly-tests.yml
chandrasekaranpradeep: Got it, will remove this generate-nightly-tests.yml

.github/workflows/model-analysis.yml
```diff
@@ -0,0 +1,161 @@
+name: Model Analysis
+
+on:
+  workflow_dispatch:
+    inputs:
+      nightly_generation:
+        description: 'Generate nightly test from the unique ops configuration across all the models'
```
vmilosevic, line 7: Can you give better description for this flag?
If I understand correctly it toggles between "generate_nightly" and "run_analysis_and_generate_md_file" modes

.github/workflows/model-analysis.yml
```diff
@@ -0,0 +1,161 @@
...
+          path: script.log
+
+      - name: Upload Models Unique Ops test Failure Logs
+        uses: actions/upload-artifact@v4
+        if: success() || failure()
+        with:
+          name: unique-ops-logs
+          path: ./models_unique_ops_output
+
+      - name: Create Pull Request
```
vmilosevic, line 148: Do we want to create a PR in both "generate_nightly" and "run_analysis_and_generate_md_file" modes, or just if "generate_nightly" = true ?

.github/workflows/model-analysis-weekly.yml
```diff
@@ -6,116 +6,8 @@ on:
...
-          base: main
-          commit-message: "Update model analysis docs"
-          title: "Update model analysis docs"
-          body: "This PR will update model analysis docs"
-          labels: automatic_model_analysis
-          delete-branch: true
-          token: ${{ env.GITHUB_TOKEN }}
-          add-paths: |
-              model_analysis_docs/
+    with:
```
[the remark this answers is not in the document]
chandrasekaranpradeep: For workflow input nightly_generation I didn't got False or True when I triggered the workflow with push like below
```
  push:
    branches: [ "pchandrasekaran/nightly_test_generation" ]
```
So when we call the workflow with schedule the same behaviour of nightly_generation will replicate right? so that why I created another workflow file for schedule and pass the nightly_generation to False

nvukobratTT:
> > Going forward, idea is to try to run all models (for now Pytorch). That said, let's update logic to mark only models that we don't wan't to have as a part of model analysis logic. E.g. collect all pytest from specified directory, while ignoring ones that have skim_model_analysis marker (or something similar) :))
> 
> I have added **model_analysis** marker in all the test inside the `forge/test/models/pytorch` path and I am collecting the test from this `forge/test/models/pytorch` path. Even if change path of collecting test from `forge/test/models/pytorch` path to `forge/test/models` path, it will only collect the test that has model_analysis marker, if we don't want to run the model analysis script for particular model test, we can remove the **model_analysis** marker and it won't be included in the collected test. There is no need for new marker for skipping the test for model analysis.

My point here was reversed approach, not keeping both. 

E.g. instead of marking which models to include in analysis, let's mark which one not to include. The benefits of my proposal are:
1. Every new model we add, doesn't need to have an additional marker. It'll automatically be included in the analysis logic.
2. It's easier to search and manage which models we're not running with the new marker. Why? Overtime we want to include all of them, and deprecate those markers. Therefore, having a easier overview where and why we're using it will be beneficial

Please, let me know if I'm missing point here :))

chandrasekaranpradeep:
> My point here was reversed approach, not keeping both.
> 
> E.g. instead of marking which models to include in analysis, let's mark which one not to include. The benefits of my proposal are:
> 
> Every new model we add, doesn't need to have an additional marker. It'll automatically be included in the analysis logic.
> It's easier to search and manage which models we're not running with the new marker. Why? Overtime we want to include all of them, and deprecate those markers. Therefore, having a easier overview where and why we're using it will be beneficial
> Please, let me know if I'm missing point here :))

Understood. I will add the skip_model_analysis marker for the test that we don't want to run model analysis and collect all the test that doesn't have skip_model_analysis marker

chandrasekaranpradeep:
> Make sure to outline these details when you start with documenting these workflows. This is one of the major notes that can influence our results more or less.
> 
> Also, let's mention that these aren't 100% 1:1 mapping to models. E.g. when we run full model, some optimization pass can change ops shapes, insert new ones, fuse some ops, etc.

Mentioned the details in the issue of documentation - https://github.com/tenstorrent/tt-forge-fe/issues/663#issuecomment-2572253890

scripts/model_analysis/run_analysis_and_generate_md_files.py
```diff
@@ -0,0 +1,554 @@
...
+        Args (List[str]): List of Operation Arguments if any
+        components (dict): A dictionary indicating the support status for each compiler component.
+        failure_reason (str): The reason for failure, if any, during testing.
+    """
+
+    def __init__(
+        self,
+        Op: str,
+        Operands: List[str],
+        Args: List[str],
```
ashokkumarkannan1, line 43: Any reason to have this in PascalCase?

scripts/model_analysis/run_analysis_and_generate_md_files.py
```diff
@@ -0,0 +1,554 @@
...
+        self.Args = Args
+        self.components = {}
+        for compiler_component in CompilerComponent:
+            self.components[str(compiler_component.name)] = False
+        self.failure_reason = ""
+
+    @classmethod
+    def create(cls, op_name, operand_names, operand_types, operand_shapes, operand_dtypes, args):
+
+        operands = UniqueOpTestInfo.create_operands(operand_names, operand_types, operand_shapes, operand_dtypes)
```
ashokkumarkannan1, line 56: we can use `cls.create_operands`

ashokkumarkannan1 (changes requested):

vmilosevic (approved):

scripts/model_analysis/run_analysis_and_generate_md_files.py
chandrasekaranpradeep: > Any reason to have this in PascalCase?

To match the same name in the Excel file, I am retaining it

ashokkumarkannan1 (approved):
LGTM!

PR #1004, opened 2025-01-03
set return_dict to false in nlp models

kamalrajkannan78:
## Summary 

This PR sets return_dict to false in Albert, Bert, DPR models to avoid the below issue(check logs for full trace) on verify function &  replace compare_with_golden with verify 

```
    for fw, co in zip(fw_out, co_out):
            if verify_cfg.verify_dtype:
>               if fw.dtype != co.dtype:
E               AttributeError: 'dict' object has no attribute 'dtype'
```

Before Fix:

- [before_fix.zip](https://github.com/user-attachments/files/18301228/before_fix.zip)

After Fix:

- [after_fix.zip](https://github.com/user-attachments/files/18301237/after_fix.zip)


forge/test/models/pytorch/text/albert/test_albert.py
```diff
@@ -5,24 +5,31 @@
...
 
 @pytest.mark.nightly
 @pytest.mark.model_analysis
-@pytest.mark.xfail(reason="TT_FATAL(weights.get_dtype() == DataType::BFLOAT16) in embedding op")
-@pytest.mark.parametrize("variant", variants, ids=variants)
-@pytest.mark.parametrize("size", sizes, ids=sizes)
+@pytest.mark.parametrize(
+    "size, variant",
+    [
+        pytest.param("base", "v1", marks=pytest.mark.xfail(reason="Tensor mismatch. PCC = 0.6316098467731136")),
```
nvukobratTT, line 16: Minor comment: For models that fail with PCC issues, let's just disable the PCC check at the end, with a new flag in the tracking document.

 That way, models can still compile end-to-end, and we can track if any regressions appear.  Regarding PCC, we will figure it out as a next step.

For more details, please check the verification docs :))

forge/forge/verify/config.py
```diff
@@ -264,6 +264,7 @@ class VerifyConfig:
     verify_size: bool = True  # Check output size
     verify_dtype: bool = True  # Check output dtype
     verify_shape: bool = True  # Check output shape
+    verify_values: bool = True  # Check output values
```
nvukobratTT, line 267: @vkovinicTT can you confirm if this expansion makes sense :))

This seems like a valid case where we want to check everything, but ignore the allclose/pcc checks due to issues with TTNN. Please, comment more if this should be handled differently.

forge/test/models/pytorch/text/dpr/test_dpr.py
```diff
@@ -89,28 +82,31 @@ def test_dpr_question_encoder_pytorch(variant, test_device):
...
     )
-    co_out = compiled_model(*inputs)
-    fw_out = model(*inputs)
 
-    co_out = [co.to("cpu") for co in co_out]
-    fw_out = [fw_out] if isinstance(fw_out, torch.Tensor) else fw_out
+    if variant == "facebook/dpr-question_encoder-multiset-base":
+        verify_values = False
+    elif variant == "facebook/dpr-question_encoder-single-nq-base":
+        verify_values = True
```
nvukobratTT, line 89: Do we need this `elif` block? By default, value verification should be enabled.

forge/test/models/pytorch/text/dpr/test_dpr.py
```diff
@@ -89,28 +82,31 @@ def test_dpr_question_encoder_pytorch(variant, test_device):
...
+
+    verify(inputs, model, compiled_model, verify_cfg=VerifyConfig(verify_values=verify_values))
 
-    assert all([compare_with_golden(golden=fo, calculated=co) for fo, co in zip(fw_out, co_out)])
+    if verify_values:
```
nvukobratTT, line 93: What is the purpose of this block?

kamalrajkannan78: - This PR adds postprocessing steps for the variants which are passing with pcc > 0.99 
- In this test dpr-question_encoder-single-nq-base variant passed with valid pcc and  dpr-question_encoder-multiset-base variant got pcc = 0.9871025782806984
- so I used this condition **if verify_values:** to print the outputs only for variant passing with valid pcc

forge/forge/verify/config.py
vkovinicTT: Yeah this looks fine by me.

forge/test/models/pytorch/text/dpr/test_dpr.py
nvukobratTT: What is the expected output of the DPR model? 

I see that it consists of 3 stages:

- Question encoder: encode questions as vectors
- Context encoder: encode contexts as vectors
- Reader: extract the answer of the questions inside retrieved contexts, along with a relevance score (high if the inferred span actually answers the question).

In sum, for e2e demo we need to encode context, question, and use reader to extract context based on required question. 

If this is correct, I'm suggesting to have 3 tests that compile each module, and than one additional that will represent e2e functional model :))

What do you think? 

P.S.
This addition can be done as separate issues, while for this PR we can remove this embedding print, and merge as is :))

kamalrajkannan78: thanks for the clarification @nvukobratTT

nvukobratTT (approved):

PR #1017, opened 2025-01-08
Use newer version of "Show Test Report" action

vmilosevic:
Use newer version of "Show Test Report" action
- switch to "mikepenz/action-junit-report@v5"
- add comments to PR with test results
- show failed test summary


nvukobratTT (approved):
Looks nice! :))

vmilosevic:
> Looks nice! :))

It will add new test results comments with every CI run you do, but you can also delete those comments manually it they are too much.
There is an option to update the comments after new runs, but unfortunately it doesn't work in our case. They assume you will have only one test result and we have one for each group of tests.

PR #1029, opened 2025-01-10
Add test grouping to workflow inputs

vmilosevic:
Add option to define test grouping as workflow input.
For example, this will run tests in 4 group
```
docker-build:
    uses: ./.github/workflows/build-and-test.yml
    secrets: inherit
    with:
      test_mark: 'nightly'
      test_group_cnt: 4
      test_group_ids: '[1,2,3,4]'
```

### DIFF
--- a/.github/workflows/on-nightly.yml
+++ b/.github/workflows/on-nightly.yml
@@ -11,3 +11,5 @@ jobs:
     secrets: inherit
     with:
       test_mark: 'nightly'
+      test_group_cnt: 4
+      test_group_ids: '[1,2,3,4]'
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -17,6 +17,16 @@ on:
         description: 'Git SHA of commit in tenstorrent/tt-mlir or branch name'
         required: false
         type: string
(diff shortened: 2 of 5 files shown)


nvukobratTT (approved):

.github/workflows/on-nightly.yml
```diff
@@ -11,3 +11,5 @@ jobs:
     secrets: inherit
     with:
       test_mark: 'nightly'
+      test_group_cnt: 4
+      test_group_ids: '[1,2,3,4]'
```
nvukobratTT, line 15: Do we need to specify IDs? Is it possible for this to be automatically picked?

vmilosevic: No, unfortunately the [functions you can use in gihub context ](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#functions) are very limited (contains, startsWith, fromJSON, toJSON ..).  You can't even get the length of an array.

A way this could be done is to add one more job that can do more complicated things in bash and output the list but I thought this complicates workflow unnecessarily and gone with simpler solution

PR #1036, opened 2025-01-13
Log failing reason exception in pytest report xml

vbrkicTT:
Log exception details when failing reason doesn't match actual exception.


forge/test/operators/pytorch/conftest.py
```diff
@@ -23,17 +25,59 @@ def pytest_runtest_makereport(item: _pytest.python.Function, call: _pytest.runne
...
+                        sections=report.sections,
+                        nodeid=report.nodeid,
+                        location=report.location,
+                        keywords=report.keywords,
+                    )
+                    outcome.force_result(new_report)
+            else:
+                logger.debug(f"Test '{item.name}' failed with exception: {type(exception_value)} '{exception_value}'")
+
+        log_test_vector_properties(item, report, xfail_reason)
```
vbrkicTT, line 58: Errors here should be catched

vbrkicTT:
The PR has been split into #1087 and #1085

dgolubovicTT (approved):

PR #1053, opened 2025-01-16
Optimizer tests and states fix

pglusacTT:
This PR will add:
- Optimizer unit tests
- Fix for states update in optimizers
- Workaround for [metal issue](https://github.com/tenstorrent/tt-metal/issues/16352) hit in Adam 

I propose that we have the following testing structure:
- Optimizer unit tests (the ones in this PR) that will isolate the optimizer (copy golden gradients, compare only step)
- Integration tests that run everything besides the optimizers on host (without gradients being copied). This kind of test was added [here](https://github.com/tenstorrent/tt-forge-fe/pull/928)
- E2E tests once the loss functions problems are resolved

I think that this is thorough enough, @vladimirjovanovicTT please let me know what you think

### DIFF
--- a/forge/forge/optimizers.py
+++ b/forge/forge/optimizers.py
@@ -237,12 +237,14 @@ class Adam(Optimizer):
             self.set_parameters_to_optimize(parameters)
 
     def get_cpu_param_dict(self, dtype: torch.dtype, shape: Tuple[int]) -> Dict:
+        # HACK: shapes are set to the shape of the parameter because of the following issue
+        # https://github.com/tenstorrent/tt-metal/issues/16352
         if self.bias_correction:
             return {
                 "torch_mean": torch.full(shape, 0.0, dtype=dtype),
                 "torch_variance": torch.full(shape, 0.0, dtype=dtype),
-                "torch_beta1_pow": torch.full((1,), 1.0, dtype=dtype),
-                "torch_beta2_pow": torch.full((1,), 1.0, dtype=dtype),
+                "torch_beta1_pow": torch.full(shape, 1.0, dtype=dtype),
+                "torch_beta2_pow": torch.full(shape, 1.0, dtype=dtype),
             }
         else:
             return {
@@ -257,15 +259,15 @@ class Adam(Optimizer):
         torch_lr = torch.full((1,), self.learning_rate, dtype=dtype)
         if self.bias_correction:
             return {
-                "lr": Tensor.create_from_torch(torch_lr),
+                "lr": Tensor.create_from_torch(torch.full(shape, self.learning_rate, dtype=dtype)),
                 "mean": Tensor.create_from_torch(torch.full(shape, 0.0, dtype=dtype)),
                 "variance": Tensor.create_from_torch(torch.full(shape, 0.0, dtype=dtype)),
-                "beta1_pow": Tensor.create_from_torch(torch.full((1,), 1.0, dtype=dtype)),
-                "beta2_pow": Tensor.create_from_torch(torch.full((1,), 1.0, dtype=dtype)),
+                "beta1_pow": Tensor.create_from_torch(torch.full(shape, 1.0, dtype=dtype)),
+                "beta2_pow": Tensor.create_from_torch(torch.full(shape, 1.0, dtype=dtype)),
             }
         else:
             return {
-                "lr": Tensor.create_from_torch(torch_lr),
+                "lr": Tensor.create_from_torch(torch.full(shape, self.learning_rate, dtype=dtype)),
                 "mean": Tensor.create_from_torch(torch.full(shape, 0.0, dtype=dtype)),
                 "variance": Tensor.create_from_torch(torch.full(shape, 0.0, dtype=dtype)),
             }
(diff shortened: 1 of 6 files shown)


forge/forge/optimizers.py
```diff
@@ -237,12 +237,14 @@ def __init__(
             self.set_parameters_to_optimize(parameters)
 
     def get_cpu_param_dict(self, dtype: torch.dtype, shape: Tuple[int]) -> Dict:
+        # HACK: shapes are set to the shape of the parameter because of the following issue
```
nvukobratTT, line 240: Minor comment, but instead of plain "HACK", it looks to me as if this is more a "Workaround" category. "Hacks" are something that isn't likely land on main :))

forge/forge/optimizers.py
```diff
@@ -352,15 +356,15 @@ def generate_op_trace(self, ac, parameter, gradient):
 
         if self.bias_correction:
             # bias_correction1 = 1 - beta1 ** step
-            beta1_one = ac.constant(1.0)
-            beta1_pow = ac.input("beta1_pow", (1,), disable_consteval=True)  # stores beta1 ** step
+            beta1_one = ac.tensor(torch.full(parameter_shape, 1.0))
```
nvukobratTT, line 359: As this workaround spans across multiple places, can we crate a tracking issue to revert part of the changes once metal issues are resolved?

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -5,6 +5,7 @@
 import pytest
 import time
 
+from test.mlir.utils import *
```
nvukobratTT, line 8: Let's follow common guidelines for organizing python imports. E.g.
1. Builins
2. Third party libs
3. Our packages/libs

forge/test/mlir/test_optimizers.py
```diff
@@ -0,0 +1,159 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
```
nvukobratTT, line 5: Similar imports comment. Let's sort them out

forge/test/mlir/test_optimizers.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # Step
+        tt_optimizer.step()
+        golden_optimizer.step()
+
+        # Compare all the parameters
+        for i, (tt_param, golden_param) in enumerate(
+            zip(tt_model.framework_module.module.parameters(), golden_model.parameters())
+        ):
+            assert compare_with_golden(
```
nvukobratTT, line 46: Can we use our common verify function? Feel free to check model tests for more refferences

forge/test/mlir/test_optimizers.py
```diff
@@ -0,0 +1,159 @@
...
+        (784, 784, 10),
+        (33, 27, 127),
+        (128, 10, 20),
+    ],
+)
+@pytest.mark.push
+def test_sgd(shape):
+    torch.manual_seed(0)
+    num_epochs = 10
+    learning_rate = 1
```
nvukobratTT, line 66: I assume we use larger LR for quicker convergence (outline data error earlier). Can you add a comment if that is the case? Just to avoid confusion as this is non-standard LR.

nvukobratTT (approved):

forge/test/mlir/test_optimizers.py
pglusacTT: Verify compares the outputs of the models, and here we are comparing weights of the models.
I understand that those two are highly correlated (if optimizers work the outputs will be the same), but if our forward breaks for some reason that does not mean that the optimizer itself is broken, and it would make this a test of two components rather than being a unit test of a single component.

Does this make sense? Please let me know if I am missing something or if I am misunderstanding verify

forge/forge/optimizers.py
pglusacTT: I agree that HACK is maybe too strong of a word here, but we should have a convention that is recognized by most IDEs and setups like TODO, FIXME and HACK keywords are

Do you have a proposal? I don't think WORKAROUND is recognized by most IDEs, but we could add a settings.json file that has that set up for vs code since most of us are using it

[Here](https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree) is the list of such common keywords in vscode

nvukobratTT: From my perspective, we can use a common `TODO` for this use case. 

In sum, having a description of a specific change, with a follow-up line with a specific call to action (`TODO`) is good enough for now. Why? Mostly, because introducing new keywords will work for individual setups, while most engineers will not see any benefits of it, just confusion as in this case. 

But I agree, this is a good pattern we should look into and propose a solution. My question is are we there yet, or this is more of pre-mature optimization? If we're there, let's share this Q with a wider audience and get more feedback :))

forge/test/mlir/test_optimizers.py
nvukobratTT: Hmm, but the check doesn't compare forward tensor, but rather parameters directly? My question is why we don't use the general `verify` method here? 

To avoid confusion, the `verify` method isn't intended to be used only for e2e model verification. But rather, to be used for any tensor comparison (outputs, constants, parameters, etc.). Also, it has an extended set of verification tools that will check shapes and data types as well (which is neat for parameter checking as well). You can see more details in this file (`forge/forge/verify/value_checkers.py`). 

Please, let me know if I'm missing some points here :))

forge/test/mlir/test_optimizers.py
```diff
@@ -0,0 +1,162 @@
...
+    [
+        # input, hidden, output
+        (784, 784, 10),
+        (33, 27, 127),
+        (128, 10, 20),
+    ],
+)
+@pytest.mark.push
+def test_sgd(shape):
+    torch.manual_seed(0)
```
pilkicTT, line 66: FYI, `torch.manual_seed()` is not needed, i think i have introduced this in the training tests by mistake... 😄

pilkicTT (approved):

nvukobratTT:
@pglusacTT is this one ready for review? If not, can we move it to draft?

PR #1058, opened 2025-01-17
Fix DecomposeMultiDimSqueeze callback failure in densenet121_hf_xray variant. 

### LINKED ISSUE #642: TypeError in DecomposeMultiDimSqueeze TVM callback in densenet xray variant
## Description

-   TypeError: int() argument must be a string, a bytes-like object or a real number, not 'Any' in DecomposeMultiDimSqueeze TVM callback in densenet xray variant

## Reproduce

```
git checkout kkannan/model_issues
pytest forge/test/model_demos/high_prio/cnn/pytorch/test_densenet.py -k "xray" -svv
```

## Observed Behaviour
```
>   input_shape = [int(dim) for dim in pre.args[0].checked_type.shape]
E   TypeError: int() argument must be a string, a bytes-like object or a real number, not 'Any'
```

## Logs

- [densenet_xray.log](https://github.com/user-attachments/files/17658982/nov7_densenet_xray.log)

meenakshiramanathan1:
This PR tracks [this](https://github.com/tenstorrent/tt-tvm/pull/52/commits) tvm commit which fixes #642.


forge/test/models/pytorch/vision/densenet/test_densenet.py
```diff
@@ -43,10 +43,14 @@ def test_densenet_121_pytorch(record_forge_property, variant):
     inputs = [img_tensor]
 
     # Forge compile framework model
+    compiler_cfg = forge.config._get_global_compiler_config()
+    if variant == "densenet121_hf_xray":
+        compiler_cfg.compile_depth = forge.CompileDepth.INIT_COMPILE
```
nvukobratTT, line 48: As we're going to start using superset, let's not use reduced compile depth for a moment :))

Using compile depth, test will get marked as passed on superset, which is an invalid result. Therefore, let's pause usage of compile depth for time being until we come up with a better plan.

nvukobratTT (approved):

PR #1066, opened 2025-01-20
Make maxpool2d channel last using transpose.

### LINKED ISSUE #1068: Maxpool2d should be channel last
Currently, we lower forge maxpool2d to `mlir::tt::ttir::MaxPool2dOp` which maps to ttnn.maxpool2d. Since ttnn.maxpool2d requires channel last input we should make sure that is the case before lowering to ttir. 
This will ensure that `mlir::tt::ttir::MaxPool2dOp` doesn't have the need for a channel last flag (discussed [here](https://github.com/tenstorrent/tt-mlir/pull/1664)).
At some point we will lower all pooling ops to single ttir Pooling op which will later be transformed into specific pooling ops (effectively making a buffer between forge fe and mlir). Currently, forge directly lowers to specific `mlir::tt::ttir::MaxPool2dOp`, and therefore needs to respect channel last requirement

dgolubovicTT:
Closes #1068
- Make maxpool2d channel last using transpose.
- Edit maxpool2d tests to reflect the current state.
- Remove depricated decomposes.

### DIFF
--- a/forge/forge/op/eval/forge/pooling.py
+++ b/forge/forge/op/eval/forge/pooling.py
@@ -112,6 +112,47 @@ class MaxPool2d(PyOp):
 
         return out_shape, []
 
+    def decompose(self, dc, inputs):
+        # TTNN can only perform a channel last pooling with its maxpool2d op.
+        # The TTNN  maxpool2d requires the input to be in the shape: (N, H, W, C).
+        # If the forge maxpool2d op is not channel last, we must permute the input (N, C, H, W) tensor to (N, H, W, C)
+        # and then transpose it back to (N, C_out, H_out, W_out) afterward.
+        #     - This is done with two transposes
+        #     - (N, C, H, W) --> transpose(-3, -2): (N, H, C, W) --> transpose(-2, -1): (N, H, W, C)
+        # Afterward:
+        #     - (N, H_out, W_out, C_out) --> transpose(-2, -1): (N, H_out, C_out, W_out) --> transpose(-3, -2): (N, C_out, H_out, W_out)
+        activations = inputs[0]
+
+        is_channel_last = self.channel_last
+
+        if not is_channel_last:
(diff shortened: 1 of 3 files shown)


forge/forge/op/eval/forge/pooling.py
```diff
@@ -111,6 +111,47 @@ def shape(self, input_shapes):
...
+                    self.padding_left,
+                    self.padding_right,
+                    self.padding_top,
+                    self.padding_bottom,
+                    True, # channel_last
+                ),
+                new_inputs,
+            )
+
+            result = dc.op(TransposeTM.create(dim0=-2, dim1=-1), [result])
```
nvukobratTT, line 151: Can we expand the comments for this one? E.g. why transposing results back.

forge/forge/op/eval/forge/pooling.py
```diff
@@ -718,285 +759,6 @@ def decompose(type, attr, dc, inputs):
 
         dc.fuse(result)
 
-    elif type == "max_pool2d":
```
nvukobratTT, line 721: ❤️

nvukobratTT (commented):
LGTM! Few minor comments

Also, can we have a following issue with this PR?

For reviewers without context around this topic, it'll be helpful to have an issue description :))

dgolubovicTT:
> LGTM! Few minor comments
> 
> Also, can we have a following issue with this PR?
> 
> For reviewers without context around this topic, it'll be helpful to have an issue description :))

Added an issue with explanation why we do this...

forge/forge/op/eval/forge/pooling.py
dgolubovicTT: Since decompose should result in new set of ops that is equivalent to the decomposed one, we need to transpose result back to channel first if it was channel first on input. I can put that in, sure

nvukobratTT (approved):

PR #1070, opened 2025-01-20
Test different input sequence lengths for Llama

pmarkovicTT:
Add test to make sure Llama compiles and run fwd pass with different input sequence lengths as we will have inputs of various lengths during training.

Close https://github.com/tenstorrent/tt-forge-fe/issues/1071

### DIFF
--- a/forge/test/mlir/llama/test_llama_inference.py
+++ b/forge/test/mlir/llama/test_llama_inference.py
@@ -6,6 +6,7 @@ import torch
 import pytest
 
 import forge
+from forge.verify.verify import verify
 from test.mlir.llama.utils.utils import load_model
 
 
@@ -152,4 +153,5 @@ def test_llama_input_sequence_lengths(model_path, seq_len):
 
     # Compile the model and run fwd pass
     compiled_model = forge.compile(framework_model, input_ids)
-    logits = compiled_model(input_ids)
+
+    verify([input_ids], framework_model, compiled_model)
--- a/forge/test/mlir/llama/test_llama_inference.py
+++ b/forge/test/mlir/llama/test_llama_inference.py
@@ -137,6 +137,7 @@ def test_llama_inference_cache_cpu(model_path):
     ],
 )
 @pytest.mark.parametrize("seq_len", [1, 2, 4, 7, 8, 16, 28, 32, 63, 64, 99, 117, 128, 256, 341, 512, 1024, 1790, 2048])
+@pytest.mark.skip(reason="No need to run in CI as it takes a long time to run.")
 def test_llama_input_sequence_lengths(model_path, seq_len):
     # Load Model and Tokenizer
     framework_model, tokenizer = load_model(model_path, seq_len=seq_len)
@@ -147,7 +148,7 @@ def test_llama_input_sequence_lengths(model_path, seq_len):
     tokenizer.model_max_length = seq_len
 
     prompt = "Q: What is the largest animal?\nA:"
-    input_ids = tokenizer(prompt, return_tensors="pt").input_ids
+    input_ids = tokenizer(prompt, padding="max_length", truncation=True, return_tensors="pt").input_ids
 
     # Compile the model and run fwd pass
     compiled_model = forge.compile(framework_model, input_ids)
--- a/forge/test/mlir/llama/test_llama_inference.py
+++ b/forge/test/mlir/llama/test_llama_inference.py
@@ -130,26 +130,22 @@ def test_llama_inference_cache_cpu(model_path):
     print(generated_text)
 
 
-@pytest.mark.parametrize(
-    "model_path",
-    [
-        "openlm-research/open_llama_3b",
-        pytest.param("meta-llama/Llama-3.2-1B", marks=pytest.mark.xfail(reason="Unsupported Op: repeat_interleave")),
-    ],
-)
-@pytest.mark.parametrize("seq_len", [1, 2, 4, 7, 8, 16, 28, 32, 63, 64, 99, 117, 128, 256, 341, 512, 1024, 1790, 2048])
-@pytest.mark.skip(reason="No need to run in CI as it takes a long time to run.")
+@pytest.mark.parametrize("model_path", ["openlm-research/open_llama_3b", "meta-llama/Llama-3.2-1B"])
+@pytest.mark.parametrize("seq_len", [2048, 512, 128])
 def test_llama_input_sequence_lengths(model_path, seq_len):
+    if model_path == "openlm-research/open_llama_3b" and seq_len == 2048:
+        pytest.skip("ValueError: Data mismatch for openlm-research/open_llama_3b - sequence length of 2048")
     # Load Model and Tokenizer
-    framework_model, tokenizer = load_model(model_path, seq_len=seq_len)
+    framework_model, tokenizer = load_model(model_path, num_hidden_layers=1)
 
     # Adjust tokenizer for max sequence length padding
-    tokenizer.pad_token = "<pad>"
+    tokenizer.pad_token = tokenizer.eos_token
     tokenizer.padding_side = "right"
     tokenizer.model_max_length = seq_len
 
     prompt = "Q: What is the largest animal?\nA:"
     input_ids = tokenizer(prompt, padding="max_length", truncation=True, return_tensors="pt").input_ids
+    input_ids = input_ids.to(torch.int32)
 
     # Compile the model and run fwd pass
     compiled_model = forge.compile(framework_model, input_ids)
--- a/forge/test/mlir/llama/test_llama_inference.py
+++ b/forge/test/mlir/llama/test_llama_inference.py
@@ -145,7 +145,6 @@ def test_llama_input_sequence_lengths(model_path, seq_len):
 
     prompt = "Q: What is the largest animal?\nA:"
     input_ids = tokenizer(prompt, padding="max_length", truncation=True, return_tensors="pt").input_ids
-    input_ids = input_ids.to(torch.int32)
 
     # Compile the model and run fwd pass
     compiled_model = forge.compile(framework_model, input_ids)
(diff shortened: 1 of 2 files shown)


forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -125,3 +125,28 @@ def test_llama_inference_cache_cpu(model_path):
...
+    tokenizer.pad_token = "<pad>"
+    tokenizer.padding_side = "right"
+    tokenizer.model_max_length = seq_len
+
+    prompt = "Q: What is the largest animal?\nA:"
+    input_ids = tokenizer(prompt, return_tensors="pt").input_ids
+
+    # Compile the model and run fwd pass
+    compiled_model = forge.compile(framework_model, input_ids)
+    logits = compiled_model(input_ids)
```
nvukobratTT, line 153: Do we want to verify output as well? 

To confirm that PCC is in the correct state as well for different input sizes.

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -125,3 +125,29 @@ def test_llama_inference_cache_cpu(model_path):
...
+
+@pytest.mark.parametrize(
+    "model_path",
+    [
+        "openlm-research/open_llama_3b",
+        pytest.param("meta-llama/Llama-3.2-1B", marks=pytest.mark.xfail(reason="Unsupported Op: repeat_interleave")),
+    ],
+)
+@pytest.mark.parametrize("seq_len", [1, 2, 4, 7, 8, 16, 28, 32, 63, 64, 99, 117, 128, 256, 341, 512, 1024, 1790, 2048])
+@pytest.mark.skip(reason="No need to run in CI as it takes a long time to run.")
```
nvukobratTT, line 138: My recommendation is to choose which of these will be part of the training focus, instead of skipping it entirely. 

E.g. if we're going to focus on training 2048 seq len model, let's fully compile and run as part of push CI that variant alone.

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -125,3 +125,29 @@ def test_llama_inference_cache_cpu(model_path):
...
+    # Adjust tokenizer for max sequence length padding
+    tokenizer.pad_token = "<pad>"
+    tokenizer.padding_side = "right"
+    tokenizer.model_max_length = seq_len
+
+    prompt = "Q: What is the largest animal?\nA:"
+    input_ids = tokenizer(prompt, padding="max_length", truncation=True, return_tensors="pt").input_ids
+
+    # Compile the model and run fwd pass
+    compiled_model = forge.compile(framework_model, input_ids)
```
nvukobratTT, line 152: Do we want to test out bwd compile/run as well? 

One general question, is there a clean way to test a backward part of a graph in isolation? For example, our compile should return compiled context that contains information about each compiled component (e.g. fwd, bwd, loss, etc.). 

Therefore, is there a clean way to just call the bwd part of the graph with random inputs, without a need to run the forward part, and initialize the loss and optimizer part of the training workflow? 

_Note: this is not a requirement for this PR, just a general question that can be useful here as well. I.e. can we have granular tests that target specific functionality, rather than the whole workflow (only the bwd part of the model). I see this as especially useful for bwd generallity push in the future. cc @vladimirjovanovicTT_

forge/test/mlir/llama/test_llama_inference.py
vladimirjovanovicTT: That's right - understanding which sequence length is relevant for Llama finetuning is one of the training team's tasks. 
Once we establish which set of seq lengths is needed, we will continue with PCC tests and run as part of CI.

forge/test/mlir/llama/test_llama_inference.py
vladimirjovanovicTT: I think this is a must-have functionality as part of our training generality/BFS effort.
Let's discuss the implementation details offline.

forge/test/mlir/llama/test_llama_inference.py
pmarkovicTT: Yep, added verify

forge/test/mlir/llama/test_llama_inference.py
pmarkovicTT: Agree as well, will update seq_len parameters with required ones for training once we choose them (we will run some experiments separately).

pmarkovicTT: This is updated to use only dim sizes we care about. Additionally, I setup only one hidden layer to be used for test to speed it up (while I also ran full model test locally to make sure it passes).

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -124,3 +125,30 @@ def test_llama_inference_cache_cpu(model_path):
     # Generated text
     generated_text = tokenizer.decode(generated_tokens[0], skip_special_tokens=True)
     print(generated_text)
+
+
+@pytest.mark.parametrize(
+    "model_path",
+    [
+        "openlm-research/open_llama_3b",
+        pytest.param("meta-llama/Llama-3.2-1B", marks=pytest.mark.xfail(reason="Unsupported Op: repeat_interleave")),
```
nvukobratTT, line 134: Support for `repeat_interleave` is added, so feel free to test out 3.2. 1B on latest main :))

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -124,3 +125,30 @@ def test_llama_inference_cache_cpu(model_path):
...
+    framework_model, tokenizer = load_model(model_path, num_hidden_layers=1)
+
+    # Adjust tokenizer for max sequence length padding
+    tokenizer.pad_token = tokenizer.eos_token
+    tokenizer.padding_side = "right"
+    tokenizer.model_max_length = seq_len
+
+    prompt = "Q: What is the largest animal?\nA:"
+    input_ids = tokenizer(prompt, padding="max_length", truncation=True, return_tensors="pt").input_ids
+    input_ids = input_ids.to(torch.int32)
```
nvukobratTT, line 149: Why is this required? What is the default type for input IDs? 

Do we expect that embedding input will always be int-based? If yes, maybe we should have a pass that will encompass this.

forge/test/mlir/llama/utils/utils.py
```diff
@@ -16,6 +16,7 @@ def load_model(model_path="openlm-research/open_llama_3b", **kwargs):
     config.use_cache = kwargs.get("use_cache", False)
     config.output_attentions = kwargs.get("output_attentions", False)
     config.output_hidden_states = kwargs.get("output_hidden_states", False)
+    config.num_hidden_layers = kwargs.get("num_hidden_layers", 26)
```
nvukobratTT, line 19: Was this intentional? 

Any specific reasons for updating original number of hidden layers?

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -124,3 +125,30 @@ def test_llama_inference_cache_cpu(model_path):
...
+
+
+@pytest.mark.parametrize(
+    "model_path",
+    [
+        "openlm-research/open_llama_3b",
+        pytest.param("meta-llama/Llama-3.2-1B", marks=pytest.mark.xfail(reason="Unsupported Op: repeat_interleave")),
+    ],
+)
+@pytest.mark.parametrize("seq_len", [128, 512, 2048])
```
nvukobratTT, line 137: Any thoughts on testing on lower precisions? E.g. bfloat16? 

In full precision, Open Llama will require 12GB, while 3.2. 4GB. That said, we should either:
- Test out lower precision DF (Open Llama will barely fit n150 for inference, definitely not for training)
- Focus only on Llama 3.2 for training. In this case as well, we'll need to run in half-precision for training in order to fit on n150 (depending on which optimizer we use during fine-tuning; full training will probably be a stretch)

forge/test/mlir/llama/utils/utils.py
pmarkovicTT: Yep, that's per our discussion in the last sync. Running llama with all layers takes quite some time and since this is not e2e/demo test, I thought it makes sense to speed it up by using a single layer.

forge/test/mlir/llama/test_llama_inference.py
pmarkovicTT: Yep you are completely right. That's something we plan to do with llama backward/training tests, and eventually incorporate this test there as well. We are currently investigating memory footprint of llama models on GPU to find optimal setup for our devices. Our plan is to add tests based on findings.

nvukobratTT:
@pmarkovicTT  is this one ready for review and potential merge? If not, can we move it to draft?

forge/test/mlir/llama/test_llama_inference.py
```diff
@@ -124,3 +125,26 @@ def test_llama_inference_cache_cpu(model_path):
     # Generated text
     generated_text = tokenizer.decode(generated_tokens[0], skip_special_tokens=True)
     print(generated_text)
+
+
+@pytest.mark.parametrize("model_path", ["openlm-research/open_llama_3b", "meta-llama/Llama-3.2-1B"])
+@pytest.mark.parametrize("seq_len", [2048, 512, 128])
+def test_llama_input_sequence_lengths(model_path, seq_len):
```
nvukobratTT, line 132: Isn't this test run on push or nightly? Is this intentional?

forge/test/mlir/llama/test_llama_inference.py
nvukobratTT: I think this one is resolved, I ping on issues to confirm should we close it

nvukobratTT (approved):
Not to block anymore, left are minor details, so I'm approving this PR :)) 

Regarding CI, I would suggest adding 1B variant just to make sure that it's passing. 3B we can skip. 

And make sure to check do you need input IDs DF :))

PR #1078, opened 2025-01-21
Restructuring Benchmark Models Data

vcanicTT:
Restructuring data we get and generate during the benchmark runs. The database where we preserve the data must be aligned with the data we generate.

### DIFF
--- a/forge/test/benchmark/benchmark/models/mnist_linear.py
+++ b/forge/test/benchmark/benchmark/models/mnist_linear.py
@@ -12,6 +12,8 @@ from torch import nn
 import forge
 from forge.verify.compare import compare_with_golden_pcc
 
+# Common constants
+GIT_REPO_NAME = "tenstorrent/tt-forge-fe"
 
 # Batch size configurations
 MNIST_BATCH_SIZE_EXP_RANGE = 7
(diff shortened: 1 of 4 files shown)


vmilosevic (approved):

vcanicTT:
@pilkicTT Please, take a look! Especially, the part with the loop count.

forge/test/benchmark/benchmark/models/mnist_linear.py
```diff
@@ -97,7 +101,8 @@ def test_mnist_linear(
 
     compiled_model = forge.compile(framework_model, sample_inputs=inputs)
     start = time.time()
-    co_out = compiled_model(*inputs)
+    for _ in range(loop_count):
```
pilkicTT, line 104: It might be good to run inference one time before starting the timer. For program cache to warm up (once we enable it) and to also exclude device open time from the measurement.

forge/test/benchmark/scripts/github_token.txt
```diff
@@ -0,0 +1 @@
+1234567891011
```
pilkicTT, line 1: Why is this added?

vcanicTT: Removed.

forge/test/benchmark/benchmark/models/mnist_linear.py
vcanicTT: @pilkicTT I put the first run outside of the loop.

pilkicTT (approved):

PR #1097, opened 2025-01-23
Run perf tests nightly

vmilosevic:
Move perf tests from on PR workflow to nightly


vcanicTT (approved):
@vmilosevic I merged benchmark changes in tt-forge-fe, so I think this change can go.

PR #1093, opened 2025-01-23
Add post processing steps for Albert, DPR, bert(masked LM), T5, Auoencoder, Whisper

kamalrajkannan78:
## Summary

- post processing steps added for Albert, DPR, bert(masked LM), T5, Auoencoder, Whisper
- push marker is added to all above models except whisper because it is failing on optimized graph pass due to this error `AssertionError: Eltwise binary ops must have the same shape in both inputs, or one operand must be 1 wide to broadcast: [1, 3000, 1, 384] vs [1, 384, 3000, 1] ` on latest main-> [whisper_issue.log](https://github.com/user-attachments/files/18517977/jan23_whisper_check_1.log) 
## Note :
- **Albert**  : Although we are using task-specific models like AlbertForTokenClassification, these models are pre-trained on general datasets. To make them suitable for specific tasks, they need to be fine-tuned on a labeled dataset for that task. After fine-tuning, the model's classification head (i.e., the classifier) will have the appropriate weights for the task at hand.
- **DPR** : The architectures of the DPR Context Encoder and Question Encoder are identical. Both consist of the same underlying components without any additional classification layers. The DPR Reader, on the other hand, extends this architecture by including a classifier layer, which enables it to predict spans for answers. -> [DPR_archs.log](https://github.com/user-attachments/files/18518185/archs.log)

- As a result, post-processing cannot be applied to the outputs of the Context Encoder and Question Encoder, as their outputs are embeddings rather than classifiable logits. @nvukobratTT  [comment](https://github.com/tenstorrent/tt-forge-fe/pull/1004#discussion_r1908791517) - separate test for end-to-end testing  is unnecessary since both encoding and reading functionalities are already encapsulated within the Reader module itself. 

- **Autoencode**r : [Our Linear Autoencoder implementation](https://github.com/tenstorrent/tt-forge-fe/blob/9895fdee000945c3c43bcf88dbca6d76106bcebc/forge/test/models/pytorch/vision/autoencoder/test_autoencoder.py#L52) is based on the [this](https://github.com/udacity/deep-learning-v2-pytorch/blob/master/autoencoder/linear-autoencoder/Simple_Autoencoder_Solution.ipynb). However, pretrained weights are not provided, so training the model is necessary to obtain meaningful outputs.

PR #1103, opened 2025-01-23
Code cleanup in pytorch models

kamalrajkannan78:
## Summary 

This PR cleans up 

- Unused imports, classes, functions
- Tests that involved testing parts of the model
- Buda specific wrappers
- Tests that have KVcache , past cache implementation
- Tests that that relied on the forge_pipeline
- Redundant tests


forge/test/models/pytorch/audio/whisper/test_whisper.py
```diff
@@ -0,0 +1,90 @@
...
+variants = [
+    "openai/whisper-tiny",
+    "openai/whisper-base",
+    "openai/whisper-small",
+    "openai/whisper-medium",
+    "openai/whisper-large",
+]
+
+
+def generate_model_whisper_congen_hf_pytorch(variant):
```
nvukobratTT, line 27: typo `congen`

forge/test/models/pytorch/audio/whisper/test_whisper.py
```diff
@@ -0,0 +1,90 @@
...
+    decoder_input_ids = torch.tensor([[1, 1]]) * model_config.decoder_start_token_id
+    decoder_input_ids = decoder_input_ids.to(torch.int32)
+    encoder_outputs = framework_model.model.model.encoder(input_features)[0].detach()
+    encoder_outputs = encoder_outputs.to(torch.float32)
+
+    return framework_model, [decoder_input_ids, encoder_outputs]
+
+
+@pytest.mark.nightly
+@pytest.mark.parametrize("variant", variants, ids=variants)
```
nvukobratTT, line 73: Should we include a model analysis mark as well?

forge/test/models/pytorch/audio/whisper/test_whisper.py
```diff
@@ -0,0 +1,90 @@
...
+    "openai/whisper-tiny",
+    "openai/whisper-base",
+    "openai/whisper-small",
+    "openai/whisper-medium",
+    "openai/whisper-large",
+]
+
+
+def generate_model_whisper_congen_hf_pytorch(variant):
+    class Wrapper(torch.nn.Module):
```
nvukobratTT, line 28: This test out only generative mode? Do we need to have a separate test for prefill as well? I would like to hear opinions from @dgolubovicTT on this one as well. 

E.g. do we want to have separate prefill and decode tests, or maybe have a single one with two compiles? Maybe something in between? Let me know your thoughts

forge/test/models/pytorch/multimodal/clip/utils/clip_model.py
```diff
@@ -7,20 +7,10 @@
...
-        super().__init__()
-        self.clip_model = model
-
-    def forward(self, pixel_values):
-
-        vision_outputs = self.clip_model.vision_model(pixel_values, return_dict=False)
-        return vision_outputs
 
 
 class CLIPTextWrapper(torch.nn.Module):
```
nvukobratTT, line 16: As a note, we're running an encoder on the CPU. Do we need to have a test to cover these on the device as well?

forge/test/models/pytorch/audio/whisper/test_whisper.py
```diff
@@ -0,0 +1,90 @@
...
+    inputs = processor(sample_audio, return_tensors="pt")
+    input_features = inputs.input_features
+
+    # Get decoder inputs
+    decoder_input_ids = torch.tensor([[1, 1]]) * model_config.decoder_start_token_id
+    decoder_input_ids = decoder_input_ids.to(torch.int32)
+    encoder_outputs = framework_model.model.model.encoder(input_features)[0].detach()
+    encoder_outputs = encoder_outputs.to(torch.float32)
+
+    return framework_model, [decoder_input_ids, encoder_outputs]
```
vkovinicTT, line 69: In forward function wrapper has argument `encoder_hidden_states`, I assume it is the same as the `encoder_outputs`. If they are, can we just rename one of them to match the other, so it's easier to understand what's going on :)

forge/test/models/pytorch/audio/whisper/test_whisper.py
```diff
@@ -0,0 +1,90 @@
...
+
+        def forward(self, decoder_input_ids, encoder_hidden_states):
+            dec_out = self.model.model.decoder(
+                decoder_input_ids,
+                self.decoder_attention_mask,
+                encoder_hidden_states,
+            )
+            lin_out = self.model.proj_out(dec_out[0])
+
+            return lin_out
```
vkovinicTT, line 43: Can we move this function to `utils` folder and reuse it both here and in `test_whicper_large_v3.py` file, since I think they are the same?

forge/test/models/pytorch/multimodal/clip/utils/clip_model.py
kamalrajkannan78: - please have a look on whisper e2e test script [here ](https://github.com/tenstorrent/tt-forge-fe/blob/470aaec5228ce97749435db31dfd581957a63b16/forge/test/models/pytorch/audio/whisper/test_whisper_0.py#L43C1-L123C88)on this [PR](https://github.com/tenstorrent/tt-forge-fe/pull/1093). except prefill , all the above comments are addressed in this [script](https://github.com/tenstorrent/tt-forge-fe/blob/470aaec5228ce97749435db31dfd581957a63b16/forge/test/models/pytorch/audio/whisper/test_whisper_0.py#L43C1-L123C88).

vkovinicTT (changes requested):

forge/test/models/pytorch/audio/whisper/test_whisper.py
kamalrajkannan78: - This test( testing decoder alone) will be replaced by this [script](https://github.com/tenstorrent/tt-forge-fe/blob/470aaec5228ce97749435db31dfd581957a63b16/forge/test/models/pytorch/audio/whisper/test_whisper_0.py#L43C1-L123C88) (e2e testing )  from this [PR](https://github.com/tenstorrent/tt-forge-fe/pull/1093) soon. 
- As I don't want to dupliacate same changes on both PRs I just included the changes related to cleanup alone on this PR.

forge/test/models/pytorch/audio/whisper/test_whisper.py
kamalrajkannan78: - As a result of this [change](https://github.com/tenstorrent/tt-forge-fe/pull/1103#discussion_r1928474358) , whisper & whisper3 will not have same wrapper. currently we are running only the decoder part of whisper3 using this wrapper. In future ,if this test pass completely I will remove this decoder specific wrapper too and will add changes for e2e test. At that time, If whisper3 uses [same wrapper](https://github.com/tenstorrent/tt-forge-fe/blob/1f983cbf487275c866d36ee416b0486644787bf5/forge/test/models/pytorch/audio/whisper/test_whisper_0.py#L75C1-L83C33) as whisper I will move it to utils folder. Is it okay for you?

vkovinicTT: Yep, thanks for clarification :)

vkovinicTT (approved):

mstojkovicTT (approved):
LGTM

forge/test/models/pytorch/audio/whisper/test_whisper.py
dgolubovicTT: A general guideline for model bring-up should be to divide the model into logical components based on its architecture and bring up each part incrementally. This approach allows for measurable progress, since it enables tracking which components are functional (e.g., "the encoder works, but there are issues with the decoder"). I would use smallest available version of the model for this.

I would like to have encoder and decoder tests. 

As far as prefill and decode are concerned, encoder participates in prefill so encoder test will be actual prefill test. In decode stage, decoder generates token by token based on encoder outputs and previously generated tokens (KV cache). Therefore, decoder test is actually decode phase.

In my opinion, we need one prefil, one decode, and one full model test. Prefil and Decode tests will enable us to develop both in parallel.

nvukobratTT: I agree with this. Having a prefill, decode, and full model test is a plus. 

A detail to consider is how we will do model analysis with this setup. For example, should we run both prefill and decode as part of nightly + model analysis while keeping the full model run as a reference? Can these two tests be enough to cover the full model run? 

The issue with a full model run at the start is that we'll have to compile, prefill, and decode separately, which can be a bit of an issue for the model analysis. But maybe we can work around this in a general way.

Looping in @chandrasekaranpradeep who is currently pushing model analysis effort.

chandrasekaranpradeep: To run it in model analysis pipeline, we should have separate test function for prefill and decode

**Reason:**
 In current model analysis, when we are compiling the model test, it will only run upto the ForgeModule generation and after that it will exit test and so the code below the compile will not be executed because the purpose of model analysis is to extract the unique ops configuration from TVM and then generate unique op test so there is no need to run the initial graph passes and other forge passes. In that case we don't get the compiled model that the reason we are exiting the test. 
 In sum we should have separate test function for prefill and decode and single test function should contain one compilation and if there are multiple compilation in single test function it will not be executed

kamalrajkannan78:
> Should we merge this PR @kamalrajkannan78 ?

I am waiting for your approval @nvukobratTT .

nvukobratTT (approved):

PR #1107, opened 2025-01-24
[cleanup] minor refactoring and type hinting fixes

pilkicTT:
This cleanup is centered around `compiled_graph_state.py` and the code it uses. I've taken slight detours to cleanup a bit more (e.g. `compiler_cfg.py`, `module.py`) but generally tried to contain changes as much as possible.

Removed `typing.py` as it was causing issues with circular dependencies
- now the types related to tensors are defined in `tensor.py`, and module related stuff is in `module.py`.

In `graph_lib/python_bindings.cpp`, some of the definitions are reordered so that `mypy` can generate better stubs (for some reason the order is important so that proper type hints are generated).


mstojkovicTT (approved):

forge/forge/compiled_graph_state.py
```diff
@@ -232,13 +193,12 @@ def __init__(
         self.framework_module = framework_module
         self.intermediates = []
         if self.bwd_compiled_graph_state is not None:
-            assert self.bwd_compiled_graph_state is not None
-            self.gradient_inputs = [None] * len(self.bwd_compiled_graph_state.ordered_input_gradient_names)
+            self.gradient_inputs = [torch.tensor(0)] * len(self.bwd_compiled_graph_state.ordered_input_gradient_names)
```
nvukobratTT, line 196: I'm more in favor of this change, but is there any special reason for using `torch.tensor(0)` vs `None` in this case?

forge/forge/compiled_graph_state.py
```diff
@@ -262,25 +222,27 @@ def __call__(self, *inputs: AnyTensor) -> List[torch.Tensor]:
...
+        self.inputs = [*to_pt_tensors(inputs)]
+
         inputs_and_parameters = [
-            *inputs,
+            *self.inputs,
             *self.fwd_compiled_graph_state.get_ordered_constant_tensors(),
             *self.fwd_compiled_graph_state.get_ordered_parameter_tensors(),
         ]
 
-        if any([not isinstance(t, torch.Tensor) for t in inputs_and_parameters]):
```
nvukobratTT, line 272: ❤️

forge/forge/tensor.py
```diff
@@ -507,19 +507,14 @@ def detach(self) -> Tensor:
...
-        # generate zeros
-        pt = torch.zeros(size=self.shape.get_pytorch_shape(), dtype=forge_dataformat_to_pytorch_dtype(self.data_format))
-        pt.requires_grad = self.requires_grad
-        return pt
-
     def to_framework(self, framework: str) -> "Tensor":
         return super().to_framework(framework)
 
 
+FrameworkTensor: TypeAlias = torch.Tensor | tf.Tensor
```
nvukobratTT, line 514: Should we keep this on top of `tensor.py` files?

forge/forge/tensor.py
```diff
@@ -1133,59 +1128,41 @@ def to_tf_tensors(
...
-        Tuple[Union[torch.Tensor, Tensor, tf.Tensor], ...], Dict[str, Union[torch.Tensor, Tensor, tf.Tensor]]
-    ],
-    convert_format: bool = False,
-) -> Tuple[torch.Tensor, ...]:
-    """
-    Take a tuple of either pytorch or forge tensors, and return pytorch tensors. Generate zero-tensors
-    if no value exists.
-    """
-    pytorch_tensors = []
+def create_pt_zeros(tensor: Tensor) -> torch.Tensor:
```
nvukobratTT, line 1131: What is the purpose of this function? 

I see it's used for parameter initialization for TF and Jax. Is that its only use case? 

If yes, maybe we can change its name + maybe rely on numpy tensor instead of PyTorch if it's generating it for other frameworks. Also, not requirement for this PR, just a mental note for a future if this is the case.

forge/forge/tensor.py
pilkicTT: I wanted to keep the two type definitions together.

forge/forge/tensor.py
pilkicTT: The use cases don't make sense to me, so I didn't want to waste time any further... To be honest, I can't remember the exact reason I've touched this method, other than it didn't make sense to have a method that asserts on the state of `self` to be used just for creating a brand new tensor... 😄

forge/forge/compiled_graph_state.py
pilkicTT: I've changed it because of the type error, `List[torch.Tensor]` cannot have None in itself. I'll maybe change this to `List[Optional[torch.Tensor]]`, because now that I see it I don't like it... 😄

nvukobratTT (approved):

Issue #1116, opened 2025-01-27
Need attribute name remapping and type conversion system for TTForge to MLIR lowering


mstojkovicTT:
## Problem
Currently, when lowering from Forge-FE to MLIR, we have several challenges with attribute handling:

1. Attribute name mismatches:
   - Different ops use different naming conventions (e.g., "kernel_size" vs "kernel_shape")
   - No systematic way to map between Forge-FE and MLIR attribute names
   - Hard to maintain consistency across codebase

2. Type conversion issues:
   - Some attributes need to be unsigned (uint32_t) in MLIR but Forge-FE only supports int
   - No clear way to specify intended MLIR types
   - Type conversions are scattered and inconsistent

3. Maintainability concerns:
   - Adding new ops requires modifying conversion code
   - Hard to track what attributes get renamed
   - Risk of inconsistency between similar ops

## Examples
Current issues with repeat_interleave:
'repeats' needs to be uint32_t but Forge-FE only has int

PR #1117, opened 2025-01-27
Add Attribute Name and Type Mapping System

mstojkovicTT:
### Ticket
fixes: #1116 

### What's changed
This PR introduces a new attribute mapping system that provides a clean way to handle attribute name remapping and type conversion when lowering from Forge-FE to MLIR.

Key Changes:
- Added new AttributeMapper class that manages attribute name mappings and type conversions
- Updated MLIRGenerator to use AttributeMapper for attribute handling
- Added default mappings for operations like repeat_interleave

The new system allows:
- Remapping attribute names (e.g., "kernel_size" -> "kernel_shape")
- Specifying target types for MLIR conversion (e.g., int -> uint32_t)
- Easy addition of new operation mappings
- Better separation of concerns between mapping and conversion

Example default mappings include:
- repeat_interleave: "repeats" as uint32

### DIFF
--- a/forge/csrc/passes/lower_to_mlir.cpp
+++ b/forge/csrc/passes/lower_to_mlir.cpp
@@ -50,6 +50,66 @@ using namespace tt;
  * @brief Implementation of TT-MLIR emission from the Forge module (set of graphs).
  */
 
+// Type-safe conversion enum
+enum class TargetType
+{
+    NoConversion,
+    AsUInt32,  // Will be stored as int but treated as uint32 in MLIR
+};
+
+struct AttributeRemap
+{
+    std::string new_name;          // New name for the attribute (empty string means keep original)
+    TargetType target_type_value;  // Target type conversion
+
+    // Constructor for easier initialization
+    AttributeRemap(const std::string &name = "", TargetType type = TargetType::NoConversion) :
+        new_name(name), target_type_value(type)
+    {
+    }
+};
+
+class AttributeMapper
+{
+   public:
+    AttributeMapper() { initialize_default_mappings(); }
+
+    // Add mapping for a specific op's attribute
+    void add_op_mapping(const std::string &op_name, const std::string &attr_name, const AttributeRemap &remap)
+    {
+        mappings_[op_name][attr_name] = remap;
+    }
+
+    // Get the mapped name and target type for an attribute
+    std::pair<std::string, TargetType> get_mapped_name_and_type(
+        const std::string &op_name, const std::string &attr_name) const
+    {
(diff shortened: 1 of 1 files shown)


forge/csrc/passes/lower_to_mlir.cpp
```diff
@@ -357,8 +432,17 @@ class MLIRGenerator
         // Evaluate operation operands: inputs and outputs per DPS
         llvm::SmallVector<mlir::Value> operands = get_mlir_operands(graph, op_node);
 
-        // Evaluate opeartion attributes
-        llvm::SmallVector<mlir::NamedAttribute> attributes;
+        // Process attributes using the mapper
```
pilkicTT, line 435: I think these types of comments don't add any value (I know they are already spread throughout this file 😄). If we are going to comment what the loop does, its main objective is to create mlir attributes for this particular op - the fact that it is using the mapper is not important. So, something like this:
```
// Map forge to MLIR attributes for this operation.
```

forge/csrc/passes/lower_to_mlir.cpp
```diff
@@ -147,6 +207,9 @@ class MLIRGenerator
     /// Map of lowering handlers for ttforge operations to MLIR.
     std::map<std::string, HandlerType> lowering_handler_map;
 
+    /// Attribute mapper for handling attribute conversions
+    AttributeMapper attr_mapper_;
```
pilkicTT, line 211: This can be a `static` field.

forge/csrc/passes/lower_to_mlir.cpp
```diff
@@ -162,8 +225,20 @@ class MLIRGenerator
...
-    mlir::Attribute convert_to_mlir_attribute(const tt::ForgeOpAttr &value)
+    mlir::Attribute convert_to_mlir_attribute(const tt::ForgeOpAttr &value, TargetType target_type)
     {
+        if (target_type != TargetType::NoConversion)
+        {
+            // Convert the attribute to the target type
+            switch (target_type)
+            {
+                case TargetType::AsUInt32:
+                    return builder_.getUI32IntegerAttr(static_cast<int32_t>(std::get<int>(value)));
```
pilkicTT, line 236: We should assert here that the original `int` value is actually greater or equal than 0, also cast directly to `uint32_t`.

```
TT_ASSERT(value >= 0, ...)
return builder_.getUI32IntegerAttr(static_cast<uint32_t>(std::get<int>(value)));
```

pilkicTT (changes requested):

Issue #938, opened 2024-12-19
[Llama 3.2 1B] Unsupported Ops in Self-Attention Test  (repeat_interleave)


mstojkovicTT:
## Steps to Reproduce:

To reproduce the error, run the following command:
`pytest -svv forge/test/mlir/llama/tests/test_llama_self_attn.py::test_llama_self_attn["meta-llama/Llama-3.2-1B"]`

## Expected Behavior:
The test should pass without any errors, verifying the correct behavior of the self-attention mechanism.

## Actual Behavior:
The test fails during the `RUN_MLIR_COMPILER` stage with the following error:
```
Unsupported Ops at: RUN_MLIR_COMPILER stage
repeat_interleave
                 Input_shape: [{1, 8, 1, 12, 64}, ]
                                         Attributes: repeat_interleave(1,0,){dim: 0, repeats: 1}
                                         Attributes: repeat_interleave(4,2,){dim: 2, repeats: 4}
```
## Aditional info
The repeat_interleave operation is currently unsupported. However, [ttnn.repeat_interleave](https://docs.tenstorrent.com/ttnn/latest/ttnn/ttnn/repeat_interleave.html#ttnn.repeat_interleave) support is available in Metal and awaiting uplift.

PR #1118, opened 2025-01-27
Add repeat_interleave Op

mstojkovicTT:
### Ticket
fixes: #938 

### Problem description
Uplifting repeat_interleave op

### What's changed
Added repeat_interleave op to handler map, as well as 
test for llama32 specific ops

### DIFF
--- a/forge/test/mlir/llama/tests/test_specific_ops_llama32.py
+++ b/forge/test/mlir/llama/tests/test_specific_ops_llama32.py
@@ -171,7 +171,6 @@ def test_matmul(shapes):
         ((1, 8, 1, 12, 64), 2, 4),
     ],
 )
-@pytest.mark.xfail(reason="repeats' failed to satisfy constraint: 32-bit unsigned integer attribute")
 @pytest.mark.push
 def test_repeat_interleave(shape, dim, repeats):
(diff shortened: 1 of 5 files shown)


forge/test/mlir/llama/tests/test_specific_ops_llama32.py
```diff
@@ -164,6 +164,35 @@ def forward(self, x, y):
     verify(inputs, framework_model, compiled_model, VerifyConfig(value_checker=AutomaticValueChecker(pcc=0.95)))
 
 
+@pytest.mark.parametrize(
+    "shape, dim, repeats",
+    [
+        ((1, 8, 1, 12, 64), 0, 1),
```
nvukobratTT, line 170: Can we test out different ranks, and few different shapes as well?

mstojkovicTT: We should definitely do that. I was just writing the initial test for Llama-specific ops. Should I add the `repeat_interleave` test to `forge/test/operators/tm` then?

nvukobratTT: Synced offline, adding new test to `forge/test/operators/tm` path

PR #1120, opened 2025-01-27
Update model analysis documentation

vmilosevic:
This PR will update model analysis documentation.


chandrasekaranpradeep:
@nvukobratTT FYI, I have removed the Models/ directory present in the model_analysis_docs because in the enhancement of model analysis script we have renamed all the sub directory present inside model_analysis_docs to smaller case(i.e models/, stats/) but Models/ directory was created by previous pipeline and it was not removed in current pipeline because the current pipeline only check for models directory, not Models directory

PR #1136, opened 2025-01-30
Add test for deepseek_vl

meenakshiramanathan1:
[Model](https://huggingface.co/deepseek-ai/deepseek-vl-1.3b-base) doesn't have huggingface direct version, hence model implementation is from corresponding github [repository](https://github.com/deepseek-ai/DeepSeek-VL).
Model requires timm version>=0.9.16 which was conflicting with segmentation_models_pytorch package so both versions have been upgraded to resolve conflicts.
The custom generate function generates text by iteratively predicting tokens while maintaining a fixed sequence length using padding.  
The model input shape is kept static by preallocating padded_inputs_embeds, which is updated each step  
with new token embeddings corresponding to next_token_id until the sequence is complete.

Model is currently failing in verify with pcc drop of 0.96.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+forge/test/models/pytorch/multimodal/deepseek_vl/image/training_pipelines.jpg filter=lfs diff=lfs merge=lfs -text
--- a/env/core_requirements.txt
+++ b/env/core_requirements.txt
@@ -51,3 +51,5 @@ pytorch_forecasting==1.0.0
 patool
 openpyxl==3.1.5
 GitPython==3.1.44
+dotmap==1.3.30
+einops==0.8.1
--- a/env/linux_requirements.txt
+++ b/env/linux_requirements.txt
@@ -17,11 +17,11 @@ sacrebleu==2.1.0
 sacremoses==0.0.53
 seaborn
 scikit-image==0.20.0 # For DenseNet 121 HF XRay model
-segmentation_models_pytorch==0.3.2
+segmentation_models_pytorch==0.4.0
 sentencepiece==0.2.0
 subword-nmt==0.3.8
 tensorflow-hub==0.12.0
-timm==0.6.12
+timm==0.9.16
 yolov5==7.0.9
 # The CPU versions of torch and torch visions are used due to their size being
 # several GB smaller which made a large impact on the performance of CI
--- a/forge/test/models/pytorch/multimodal/deepseek_math/test_deepseek_math.py
+++ b/forge/test/models/pytorch/multimodal/deepseek_math/test_deepseek_math.py
@@ -5,10 +5,10 @@ import pytest
 
 import forge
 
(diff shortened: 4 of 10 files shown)


nvukobratTT (commented):
Don't we have HuggingFace versions for which we don't need to copy model implementations?

meenakshiramanathan1:
> Don't we have HuggingFace versions for which we don't need to copy model implementations?

[Model](https://huggingface.co/deepseek-ai/deepseek-vl-1.3b-base) doesn't have huggingface direct version, hence model implementation is from corresponding github [repository](https://github.com/deepseek-ai/DeepSeek-VL).

nvukobratTT (commented):
As this is a lot of new code that is mostly copied from GitHub, let me see can we get new repository for this purpose, and then link it to ours.

E.g. have a repo that will contain all third-party GitHub repositories as submodules, from which we can load this and similar models to run compile against them. That way, in FFE we'll only store model tests + wrappers if some modification is needed. 

What do you think @meenakshiramanathan1 ?

meenakshiramanathan1:
> As this is a lot of new code that is mostly copied from GitHub, let me see can we get new repository for this purpose, and then link it to ours.
> 
> E.g. have a repo that will contain all third-party GitHub repositories as submodules, from which we can load this and similar models to run compile against them. That way, in FFE we'll only store model tests + wrappers if some modification is needed.
> 
> What do you think @meenakshiramanathan1 ?

Sure @nvukobratTT , I will keep this in draft for now.

ashokkumarkannan1:
> As this is a lot of new code that is mostly copied from GitHub, let me see can we get new repository for this purpose, and then link it to ours.
> 
> E.g. have a repo that will contain all third-party GitHub repositories as submodules, from which we can load this and similar models to run compile against them. That way, in FFE we'll only store model tests + wrappers if some modification is needed.
> 
> What do you think @meenakshiramanathan1 ?

That sound good @nvukobratTT . We can do that. As we are going to focus on P1 models and models supported in tt-torch which is not yet supported in tt-forge-fe. What will be the priority of moving all the GitHub copied code to a new repo? Because there are some other models also fall in this case as you know. Could you please clarify this?

nvukobratTT:
> > As this is a lot of new code that is mostly copied from GitHub, let me see can we get new repository for this purpose, and then link it to ours.
> > E.g. have a repo that will contain all third-party GitHub repositories as submodules, from which we can load this and similar models to run compile against them. That way, in FFE we'll only store model tests + wrappers if some modification is needed.
> > What do you think @meenakshiramanathan1 ?
> 
> That sound good @nvukobratTT . We can do that. As we are going to focus on P1 models and models supported in tt-torch which is not yet supported in tt-forge-fe. What will be the priority of moving all the GitHub copied code to a new repo? Because there are some other models also fall in this case as you know. Could you please clarify this?

Synced offline

Issue #1142, opened 2025-01-30
Change backward op output data format setup


pmarkovicTT:
The current logic for setting up output data format for backward ops selects data format of the first input operand.
Reference: `void OpNode::set_output_df_from_operands(const Graph *graph)` in the `node_types.cpp`.

This is problematic since different ops can require output to match data format of some other input operand. For instance, the embedding backward op should match data format of embedding weights (the second operand).

PR #1143, opened 2025-01-30
Change embedding backward op output data format setup

pmarkovicTT:
### Ticket
Close https://github.com/tenstorrent/tt-forge-fe/issues/1142

### Problem description
The current logic for setting up output data format for backward ops selects data format of the first input operand.
Reference: `void OpNode::set_output_df_from_operands(const Graph *graph)` in the `node_types.cpp`.

This is problematic since different ops can require output to match data format of some other input operand. For instance, the embedding backward op should match data format of embedding weights (the second operand).

### What's changed
Changed the way embedding backward op is created (explicitly setting up output data format through autogradContext).

### Checklist
- [x] Add set_output_df method to autogradContext
- [x] Updated backward method for embedding op (`embedding.py`)
- [x] Test new logic (test_llama_backward) - note here, couldn't find explicit tests for adding new functions so test with existing one.

### DIFF
--- a/forge/forge/op/eval/forge/embedding.py
+++ b/forge/forge/op/eval/forge/embedding.py
@@ -48,5 +48,6 @@ def decompose(type, attr, dc, inputs):
 def backward(type, attr, ac, operand, inputs, output, grad):
     assert type == "embedding"
     embedding_bw_context = ac.op("embedding_bw", [inputs[0], inputs[1], grad])
+    # Output data format should match second operand data format (embedding weights)
     ac.set_output_df(embedding_bw_context, inputs[1].output_df)
     return embedding_bw_context
--- a/forge/forge/op/eval/forge/embedding.py
+++ b/forge/forge/op/eval/forge/embedding.py
@@ -47,4 +47,6 @@ def decompose(type, attr, dc, inputs):
 
 def backward(type, attr, ac, operand, inputs, output, grad):
     assert type == "embedding"
-    return ac.op("embedding_bw", [inputs[0], inputs[1], grad])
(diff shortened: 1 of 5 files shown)


pilkicTT (changes requested):
I think i have an idea on how to do this more cleanly, let's sync tomorrow.

forge/forge/op/eval/forge/embedding.py
```diff
@@ -47,4 +47,6 @@ def decompose(type, attr, dc, inputs):
 
 def backward(type, attr, ac, operand, inputs, output, grad):
     assert type == "embedding"
-    return ac.op("embedding_bw", [inputs[0], inputs[1], grad])
+    embedding_bw_context = ac.op("embedding_bw", [inputs[0], inputs[1], grad])
+    ac.set_output_df(embedding_bw_context, inputs[1].output_df)
```
pilkicTT, line 51: Maybe add a comment here, explaining that we are explicitly setting data format based on the second operand (embedding weights)...

pilkicTT (approved):

forge/forge/op/eval/forge/embedding.py
pilkicTT: Basically, just so that we know what is `inputs[1]` - not having to debug surrounding code to figure that out. Unfortunately, our op definitions are painful to read if you haven't memorized which operand idx is which... :smile:

pmarkovicTT: Yep, totally makes sense. I forgot to copy comment from previous version.

pilkicTT:
One more thing, please change the title so it reflects that this is only related to `embedding_bw` op. 😄

PR #1150, opened 2025-01-31
Copy of the inputs for the evaluation during compilation

vkovinicTT:
### Summary

This PR ensures that a copy of the input tensors is created when running forward during compilation (e.g. when trying to verify outputs using forward pass of the framework model). This prevents unintended modifications to the original input tensors if in-place operations are performed during the forward pass.

In order for it to work, first this [PR](https://github.com/tenstorrent/tt-tvm/pull/60) in the `tt-tvm` needs to be merged.

### Why is this needed?
Some models perform in-place operations on input tensors during the forward pass, which can lead to unintended changes in the original inputs. By making a copy of the inputs, we ensure correctness and avoid potential issues when running forward during compilation.

Example test:
```python
    class Inplace(torch.nn.Module):
        def __init__(self):
            super().__init__()

        def forward(self, x):
            y = x + 1 
            x += 2 # in-place operation on the input that causes the inputs to change during forge.compile(...)

            return x + y
        
    input = torch.zeros(shape, requires_grad=False)
    framework_input = input.detach().clone()
    tt_inputs = [input]

    framework_model = Inplace()
    y = framework_model(framework_input)
        
    compiled_model = forge.compile(framework_model, sample_inputs=tt_inputs, module_name="inplace")
    tty = compiled_model(*tt_inputs)[0]

    compare_with_golden(golden=y, calculated=tty) # this would fail
```
❗❗❗ IMPORTANT NOTES ❗❗❗

In the training mode if the inputs require grad, pytorch will try to do the forward and will throw runtime error:
`a leaf Variable that requires grad is being used in an in-place operation.`, while the compiled model silently computes the gradients and **this need to be addressed**.

Introducing this change our compiler will not change it's input, so it will act as tensorflow (it doesn't change it's input tensor), but it is won't be aligned with pytorch which allows in-place changes of the input tensor.

### DIFF
--- a/forge/test/mlir/operators/tm/test_in_place.py
+++ b/forge/test/mlir/operators/tm/test_in_place.py
@@ -1,0 +1,111 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import pytest
+import torch
+import tensorflow as tf
+
+import forge
+from forge.verify.compare import compare_with_golden
+
+
+@pytest.mark.parametrize(
+    "shape",
+    [
+        (3, 3),
+    ],
+)
+@pytest.mark.push
+def test_in_place_torch(shape):
+    class Inplace(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+
+        def forward(self, x):
+            y = x + 1
+            x += 2
+
+            return x + y
+
+    input = torch.zeros(shape, requires_grad=False)
+    framework_input = input.detach().clone()
+    tt_inputs = [input]
+
+    framework_model = Inplace()
+    y = framework_model(framework_input)
+
+    compiled_model = forge.compile(framework_model, sample_inputs=tt_inputs, module_name="inplace")
+    tty = compiled_model(*tt_inputs)[0]
+
+    compare_with_golden(golden=y, calculated=tty)
+
+
+@pytest.mark.parametrize(
+    "shape",
+    [
+        (3, 3),
+    ],
+)
+@pytest.mark.push
+def test_in_place_tf(shape):
+    class Inplace(tf.keras.Model):
+        def __init__(self):
+            super().__init__()
+
+        def call(self, x):
+            y = x + 1
+            x += 2
+            return x + y
+
+    input = tf.zeros(shape)
+    framework_input = tf.identity(input)
+    tt_inputs = [input]
+
+    framework_model = Inplace()
+    y = framework_model(framework_input)
+
+    compiled_model = forge.compile(framework_model, sample_inputs=tt_inputs, module_name="inplace")
+    tty = compiled_model(*tt_inputs)[0]
+
+    # convert tensor from tf to torch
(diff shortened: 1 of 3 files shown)


forge/test/mlir/operators/tm/test_in_place.py
```diff
@@ -0,0 +1,116 @@
...
+            super().__init__()
+
+        def forward(self, x):
+            y = x + 1
+            x += 2
+
+            return x + y
+
+    input = torch.zeros(shape, requires_grad=False)
+    framework_input = input.detach().clone()
```
nvukobratTT, line 31: Why we're doing detach and clone here?

forge/test/mlir/operators/tm/test_in_place.py
```diff
@@ -0,0 +1,116 @@
...
+    tt_inputs = [input]
+
+    framework_model = Inplace()
+    y = framework_model(framework_input)
+
+    compiled_model = forge.compile(framework_model, sample_inputs=tt_inputs, module_name="inplace")
+    tty = compiled_model(*tt_inputs)[0]
+
+    compare_with_golden(golden=y, calculated=tty)
```
nvukobratTT, line 40: Should we use the standard verify function?

forge/test/mlir/operators/tm/test_in_place.py
```diff
@@ -0,0 +1,116 @@
...
+    print(framework_input)
```
nvukobratTT, line 41: Do we need these prints?

forge/test/mlir/operators/tm/test_in_place.py
```diff
@@ -0,0 +1,116 @@
...
+    framework_model = Inplace()
+    y = framework_model(framework_input)
+
+    compiled_model = forge.compile(framework_model, sample_inputs=tt_inputs, module_name="inplace")
+    tty = compiled_model(*tt_inputs)[0]
+
+    # convert tensor from tf to torch
+    y = torch.tensor(y.numpy())
+
+    compare_with_golden(golden=y, calculated=tty)
```
nvukobratTT, line 75: Similar comments as for PT example

forge/test/mlir/operators/tm/test_in_place.py
vkovinicTT: Because if run inference on the `torch` module and then want to run it on `compiled module`, inputs for the compiled module would be different due to the changes made by torch (as default torch is not functional, meaning it will change input)

forge/test/mlir/operators/tm/test_in_place.py
vkovinicTT: For the nature of the in-place problem we are facing, I have separate input for `compiled module` and `torch module` so I can't pass it to the `verify` function.

forge/test/mlir/operators/tm/test_in_place.py
vkovinicTT: Not really.

vkovinicTT:
I would still wait with this PR until we resolve which policy we want to follow (functional or per framework policy).

Issue #200, opened 2024-08-30
Create tt-forge WHL Installer


vmilosevic:
## Description:

With the increasing number of tests, we need to scale test execution across multiple runners efficiently. To achieve this, we want to build tt-forge on one machine, package its outputs, and distribute them to test runners. This will allow running tests without needing to rebuild the project on every machine. 
We can expect that the test runner machine already has the toolchain installed, so we just need to install tt-forge binaries. 

## Tasks:

- Make tt-forge installable via cmake install
    - Ensure the CMake installation process collects and copies all necessary tt-forge binaries and dependencies to a specified installation directory.
    - Provide an example of how to use cmake install to achieve this.
- Package installed artifacts into a WHL package
    - Collect the installed tt-forge artifacts and package them into a Python wheel (.whl) for easy distribution and installation.
- Use whl installer to split workflow into two stages

##  Workflow changes

Split pipeline into two jobs
- Build Job
   - Run build and create a WHL installer
   - Upload the  whl installer to artifacts
- Test jobs (matrix with two parallel jobs)
   - Check out tt-forge tests from the repository
   - Download and install whl installer
   - Run the tests (without rebuilding the project)
	
This is a prerequisite for #96

staylorTT:
What is the latest on this? @vmilosevic

pilkicTT:
In-progress, I am working on it...

PR #1152, opened 2025-01-31
[build] tt-forge-fe wheel

pilkicTT:
This change enables creating a `tt-forge-fe` wheel. To create a wheel, run either of the following commands (inside our `venv`):
- `python setup.py bdist_wheel`
- `python -m build wheel` - this one requires build to be installed, `pip install build`

Logic for creating the wheel is contained in `setup.py`. It simply runs cmake to build and install the project. The `setuptools` then picks up and creates the wheel.

`tt-mlir` is now an `ExternalProject` in cmake.

Installing `tt-mlir` via `cmake --install build --component SharedLib` installs all needed dependencies for `tt-mlir` (including `tt-metal`).

`TT_METAL_HOME` environment variable is no longer needed, small piece of logic in `forge/forge/__init__.py` tries to deduce it, relative to its (`__init__.py`) location. If it can't be deduced, an error message will be shown.

However, if the `TT_METAL_HOME` variable is set, the logic will respect it as long as it is a valid path, so overriding it in custom scenarios is possible.

Changes to the CI workflows will follow this change.

Issue #200


third_party/tt-mlir
pilkicTT, line 1: I will revert the submodule change, when the new `tt-mlir` lands in main...

nvukobratTT (approved):
LGTM! 

One suggestion from my side is to add a doc section on how to build and install wheels. We have a placeholder on the [Getting started page](https://docs.tenstorrent.com/tt-forge-fe/getting-started.html), and I think we should include more details on a regular [Build page](https://docs.tenstorrent.com/tt-forge-fe/build.html).

pilkicTT:
> LGTM!
> 
> One suggestion from my side is to add a doc section on how to build and install wheels. We have a placeholder on the [Getting started page](https://docs.tenstorrent.com/tt-forge-fe/getting-started.html), and I think we should include more details on a regular [Build page](https://docs.tenstorrent.com/tt-forge-fe/build.html).

Yeah, I refrained from updating docs, since we still don't have this in CI. Once it's there and we know it's working as expected, I'll update the docs.

PR #1154, opened 2025-01-31
Switch to new dylib structure (split compiler + runtime dylibs) for ttmlir

vwellsTT:
### Problem description
Making some changes to tt-mlir to avoid static linking.  This entails splitting TTMLIR.so into 2 parts, this is trivial PR to change so we link both these parts (compiler + runtime) so instead of master so

### What's changed
Very little, TTMLIR -> TTMLIRCompiler & TTMLIRRuntime in a cmake file

### Checklist
- [X] Clean CI run
https://github.com/tenstorrent/tt-forge-fe/actions/runs/13080436376

### DIFF
--- a/forge/csrc/CMakeLists.txt
+++ b/forge/csrc/CMakeLists.txt
@@ -87,7 +87,8 @@ target_link_libraries(ttforge_csrc PRIVATE
     LLVM
     MLIR
 
-    TTMLIR
+    TTMLIRCompiler
+    TTMLIRRuntime
 
     xml2
     curses
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,10 @@ class CMakeBuild(build_ext):
         self.spawn(["cmake", "--build", str(build_dir)])
         self.spawn(["cmake", "--install", str(build_dir)])
 
+        # NOTE: Workaround for regression in tt-mlir,
+        # Remove once tt-mlir#2129 is merged.
+        self.copy_tree(cwd / "lib", str(install_dir / "lib"))
+
 
 with open("README.md", "r") as f:
     long_description = f.read()


pilkicTT:
Hey @vwellsTT, just to clarify... this PR, or concretely this commit should go with the `tt-mlir` uplift change (not as a separate PR). This PR will never pass checks until `tt-mlir` version is bumped and `tt-mlir` version won't be bumped (checks will fail)  unless this PR is in, so we are in a deadlock.

So we can cherry-pick this commit in the uplift PR, once the auto uplift picks up your change in `tt-mlir`. Sounds good?

pilkicTT:
> Yeah, that makes sense. I wanted PR just to sort of get "verbal approval", before I merge `tt-mlir` change, since after that uplifting won't work without some kind of cherry-pick etc. But manual run of On-pr did pass, I linked in description.
> 
> I'm not familiar with how uplifting works for this repo, seems different that tt-torch and tt-xla? But I assume you don't have objections to me merge the `tt-mlir` PR, given manually trigger test works here, right?

Sure, feel free to merge it! I think the uplift should work in the same way as in those repos, i.e. there is an automatic workflow which creates a PR for uplift each day.

vwellsTT:
> Sure, feel free to merge it! I think the uplift should work in the same way as in those repos, i.e. there is an automatic workflow which creates a PR for uplift each day.

Ah, I see.  The other 2 repos have a hash manually set I can update in a PR, whereas I don't see such a mechanism here is what I mean.  But yeah that makes sense

pilkicTT:
> Ah, I see. The other 2 repos have a hash manually set I can update in a PR, whereas I don't see such a mechanism here is what I mean. But yeah that makes sense

Sorry @vwellsTT, i got so used to the automatic uplift PRs, that i completely forgot that you can actually do it manually. 🤦 The difference between us and the other FEs, is they are probably not using `git submodule` but are using CMake to take care of the `tt-mlir` repo.  Not sure if you are familiar with git submodules but here's how you can uplift manually:
```bash
cd third_party/tt-mlir
git checkout <your commit on tt-mlir>
cd ../../
git add third_party/tt-mlir
git commit -m "uplift"
```

If you go this way, just make sure before merging that `tt-mlir` submodule is pointing on a commit on main (once you merge the changes on `tt-mlir`).

So to summarize, you can either manually uplift `tt-mlir` with this PR, or we can cherry pick this PR commit on the auto-uplift PR. Either way, I'll approve this PR. Again, sorry for confusion...

pilkicTT (approved):

vwellsTT:
Uh oh, somehow docker build is failing here? Hmm.  Maybe cherry-pick to uplift PR is easier lol

vwellsTT:
Hmm, _ttnn.so should be part of `tt-mlir` SharedLib component.  Not sure I understand exactly how CI checks for this repo handles the installation, I did have to mess with cmake install process for `tt-mlir` for similar issue

vmilosevic (approved):

PR #1160, opened 2025-02-03
[CI] reorganize workflows to use python wheel

### LINKED ISSUE #200: Create tt-forge WHL Installer

pilkicTT:
Before this change, we had `build-and-test` workflow which coupled building and testing. Which means if we would want to split tests into more groups or add more testing (both on `n150` and `n300`) each of those jobs would waste time on building everything from scratch.

With this change we now have one `forge-build` job which creates both `tt-tvm` and `tt-forge-fe` wheels, which will then be reused by all of the testing jobs. Both `nightly` and `on PR` runs are udpated.

Additionally, we will run all tests on both `n150` and `n300` cards to have better test coverage. In the past, we've had blind spots where some of the tests were passing on `n150` and failing on `n300`.

In the future, the plan is to add workflows for `debug` build, test coverage runs, etc.

Closes #200

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,121 @@
+name: Build
+
+on:
+  workflow_dispatch:
+    inputs:
+      mlir_override:
+        description: 'Git SHA of commit in tenstorrent/tt-mlir or branch name'
+        required: false
+        type: string
+  workflow_call:
+    inputs:
+      mlir_override:
+        description: 'Git SHA of commit in tenstorrent/tt-mlir or branch name'
+        required: false
+        type: string
+      docker-image:
+        description: 'Docker image to use for build'
+        required: true
+        type: string
+
+permissions:
+  packages: write
+  checks: write
+  pull-requests: write # only required if `comment: true` was enabled
+
+jobs:
+
+  forge-build:
+
+    strategy:
+      fail-fast: false
+      matrix:
+        build:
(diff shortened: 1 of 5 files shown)


vmilosevic (approved):

pilkicTT:
Merging now to avoid conflicts and issues with branch protection (since we needed to change required jobs).

PR #1174, opened 2025-02-04
[cleanup] removing global compiler configuration

### LINKED ISSUE #147: [compile] remove global compiler configuration
In the previous stack, the compiler config was part of the **global state***. We should make it so that the compiler configuration is optionally passed into `pyforge.compile(...., config=user_defined_config)`.

**more generally, we should remove any global state**

pilkicTT:
This is a part of ongoing effort to remove any global state left over from pybuda.

There should be no need of having global configuration of compiler. If user wants to override some of the exposed settings, it should be done locally, by creating a new (default) config and then passing it into the compile function. All tests which need to modify the config should do the same.

No feature should rely on reading from a "volatile global configuration". Instead, each feature must either be designed to receive the user-specified compiler configuration directly or defer config-dependent decisions for later (until the compilation stage where the appropriate configuration is available).

In majority of the cases i would assume that direct access to the user passed config is easily available.

Closes #147


forge/test/conftest.py
```diff
@@ -186,7 +185,7 @@ def initialize_global_compiler_configuration_based_on_pytest_args(pytestconfig):
     Set the global compiler config options for the test session
     which will generate op tests for the given model.
     """
-    compiler_cfg = _get_global_compiler_config()
+    compiler_cfg = CompilerConfig()
```
pilkicTT, line 188: @chandrasekaranpradeep i think this will break the model analysis?

chandrasekaranpradeep: Yeah, It will break the model analysis feature.
FYI, Working on removing the dependency of global compiler config in model analysis - [PR](https://github.com/tenstorrent/tt-forge-fe/pull/1178)

chandrasekaranpradeep:
> @vbrkicTT @chandrasekaranpradeep In this change, i am removing the global compiler config. Are you aware if this can break your features (test sweeps/model analysis)? I haven't got around to testing them, yet...
> 
> Also, it would be great if you added some simple sanity tests for both test sweeps and model analysis so that we can run it on push CI, increasing the likelihood that they won't be broken.

Sure I will create a sanity test for model analysis and add it in push pipeline

forge/test/operators/utils/utils.py
```diff
@@ -90,16 +89,12 @@ class CompilerUtils:
...
         """Set compiler configuration for input source"""
-        compiler_cfg = _get_global_compiler_config()
-        compiler_cfg.input_queues_on_host = input_source_flag.input_queues_on_host
-        if input_source_flag.set_default_dram_parameters:
-            compiler_cfg.default_dram_parameters = input_source_flag.default_dram_parameters
+        raise NotImplementedError("set_input_source is not implemented")
 
     @staticmethod
     def set_math_fidelity(math_fidelity: MathFidelity):
         """Set compiler configuration for math fidelity"""
```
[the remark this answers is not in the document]
vbrkicTT: Yes, in sweep tests we are currently using setting of input_queues_on_host and default_dram_parameters.
Since it's not supported in Forge, it's safe to just skip this step.

@nvukobratTT @dgolubovicTT We may rethink if we need both input sources FROM_HOST vs FROM_DRAM_QUEUE. They use the same model it's just this config which is different.

forge/forge/config.py
```diff
@@ -117,43 +57,27 @@ class CompilerConfig:
...
     # Defines compilation depth. Used to limit scope of some unit tests
-    compile_depth: int = field(default=CompileDepth.FULL, metadata=as_json(CompileDepth))
-
+    compile_depth: CompileDepth = field(default=CompileDepth.FULL, metadata=as_json(CompileDepth))
     # Create cpu device for unsupported forge ops
     enable_tvm_cpu_fallback: bool = False
     # Types of ops to fall back to CPU for
     cpu_fallback_ops: Set[str] = field(default_factory=lambda: set(["embedding"]))
-    # Extend CPU fallback for TM ops
-    enable_tm_cpu_fallback: bool = False
```
nvukobratTT, line 149: I'm more in favor of keeping the configurations still defined in TVM or FFE. 

If we have them here, it'll be easier to follow which compiler features need to be depricate, and where. Out of existing configs, only one that match this is `enable_tm_cpu_fallback`. But let me know if you saw few others with the similar situation.

pilkicTT: Good catch, these two are used in tvm. Probably dead code, but i'll revert this part of removal, so that it can be cleaned up completely in the future.

nvukobratTT (approved):

vbrkicTT (approved):

PR #1201, opened 2025-02-10
Add test for deepseek_math prefill

meenakshiramanathan1:
Deepseek math test and deepseek math prefill test is now failing with below runtime error:
```
RuntimeError: TT_THROW @ /proj_sw/user_dev/mramanathan/bgdlab12_feb10_forge/tt-forge-fe/third_party/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/impl/allocator/bank_manager.cpp:132: tt::exception
E       info:
E       Out of Memory: Not enough space to allocate 180355072 B DRAM buffer across 12 banks, where each bank needs to store 15032320 B
```

### DIFF
--- a/forge/test/models/pytorch/multimodal/deepseek/test_deepseek_math.py
+++ b/forge/test/models/pytorch/multimodal/deepseek/test_deepseek_math.py
@@ -21,7 +21,7 @@ def test_deepseek_inference_no_cache_cpu(variant):
     framework_model = DeepSeekWrapper(model)
 
     generated_text = generation(
-        max_new_tokens=100, compiled_model=framework_model, input_ids=input_ids, tokenizer=tokenizer
+        max_new_tokens=200, compiled_model=framework_model, input_ids=input_ids, tokenizer=tokenizer
     )
     print(generated_text)
 
--- a/forge/test/models/pytorch/multimodal/deepseek/test_deepseek_math_prefill.py
+++ b/forge/test/models/pytorch/multimodal/deepseek/test_deepseek_math_prefill.py
@@ -1,0 +1,96 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import torch
+
+import forge
+from forge.verify.compare import compare_with_golden
+
+from test.models.pytorch.multimodal.deepseek.utils.model import (
+    DeepSeekWrapper_decoder,
+    download_model_and_tokenizer,
+)
+
+
+def prefill_on_cpu(model, input_ids):
+    with torch.no_grad():
+        transformer_outputs = model.model(
+            input_ids=input_ids,  # Pass the entire updated sequence
+        )
+        hidden_states = transformer_outputs.last_hidden_state
+    return hidden_states
+
+
+def decode_on_cpu(model, tokenizer, input_ids, hidden_states, max_new_tokens):
+    output_logits = []
+    for i in range(max_new_tokens):
+        # Use only the hidden state of the last token in the sequence
+        next_token_logits = model.lm_head(hidden_states[:, -1, :])
+        output_logits.append(next_token_logits)
+
+        # Get the next token ID
+        next_token_id = torch.argmax(next_token_logits, dim=-1)
+
+        # Stop if the EOS token is generated
+        if next_token_id == tokenizer.eos_token_id:
+            break
+
+        # Update input_ids with the new token
+        input_ids = torch.cat([input_ids, next_token_id.unsqueeze(0)], dim=-1)
+
+        # Run the model again to get the new hidden state
+        # Here we do effectively prefil again because we don't use KV cache.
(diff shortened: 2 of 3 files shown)


forge/test/models/pytorch/multimodal/deepseek/utils/model.py
```diff
@@ -49,4 +49,14 @@ def __init__(self, model):
...
+        return self.model(input_tensor, max_new_tokens=200).logits
+
+
+class DeepSeekWrapper_decoder(torch.nn.Module):
+    def __init__(self, model):
+        super().__init__()
+        self.model = model
+
+    def forward(self, input_tensor):
+        output = self.model(input_tensor, max_new_tokens=200)
```
vkovinicTT, line 61: Also here.

forge/test/models/pytorch/multimodal/deepseek/test_deepseek_math_prefill.py
```diff
@@ -0,0 +1,103 @@
...
+        model, tokenizer, input_ids, hidden_states_framework, max_new_tokens
+    )
+
+    # Compare the logits of the generated tokens with the golden values from CPU.
+    assert all(
+        [
+            compare_with_golden(golden=out_logits_fw, calculated=out_logits_tt)
+            for out_logits_fw, out_logits_tt in zip(output_logits_framework, output_logits_compiled)
+        ]
+    )
```
vkovinicTT, line 99: Since all of the decode is done on CPU, is there any particular reason why do we compare with golden all the subsequent logits. Correct me if I am wrong, but I think that we can only check the first `hidden_state` between framework and compiled model which you did in line 83. All the subsequent errors will accumulate if the first `hidden_state` is not correct. I'd like to hear your thoughts on this :))

forge/test/models/pytorch/multimodal/deepseek/utils/model.py
```diff
@@ -49,4 +49,14 @@ def __init__(self, model):
         self.model = model
 
     def forward(self, input_tensor):
-        return self.model(input_tensor, max_new_tokens=100).logits
+        return self.model(input_tensor, max_new_tokens=200).logits
```
vkovinicTT, line 52: Please pass the `max_new_tokens` through the constructor of the model.

forge/test/models/pytorch/multimodal/deepseek/test_deepseek_math.py
```diff
@@ -21,7 +21,7 @@ def test_deepseek_inference_no_cache_cpu(variant):
     framework_model = DeepSeekWrapper(model)
 
     generated_text = generation(
-        max_new_tokens=100, compiled_model=framework_model, input_ids=input_ids, tokenizer=tokenizer
+        max_new_tokens=200, compiled_model=framework_model, input_ids=input_ids, tokenizer=tokenizer
```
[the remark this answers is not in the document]
meenakshiramanathan1: Number of token has been increased for full generation of output since model dtype has now been changed to float32 initially it was bfloat16.

forge/test/models/pytorch/multimodal/deepseek/test_deepseek_math_prefill.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    This function tests the inference of the deepseek_math model split into two parts:
+    - The first part is the prefilling of the model on the device.
+    - The second part is the decoding of the model on the CPU without KV cache.
+    """
+
+    model_name = f"deepseek-ai/{variant}"
+    model, tokenizer, input_ids = download_model_and_tokenizer(model_name)
+
+    input_ids = input_ids.to(torch.int32)
```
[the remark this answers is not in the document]
meenakshiramanathan1: Without this cast, it was failing in pcc comparison in line 83 , similar cast was created for llama model as mentioned [here](https://github.com/tenstorrent/tt-forge-fe/blob/f42cad73ee09be7957453c74c35e8791002566ea/forge/test/mlir/llama/tests/test_llama_prefil.py#L68), this behaviour was observed when this test was created in a previous commit however this is not observed in latest main as model is failing in runtime and not going till compare_with_golden stage yet.

forge/test/models/pytorch/multimodal/deepseek/test_deepseek_math_prefill.py
vkovinicTT: Can we just compare logits from the first forward pass then :)

vkovinicTT (approved):

forge/test/models/pytorch/multimodal/deepseek/test_deepseek_math_prefill.py
vkovinicTT: Sorry I meant `assert compare_with_golden(golden= output_logits_framework[0], calculated= output_logits_compiled[0]), 'some message'`

ashokkumarkannan1 (approved):
LGTM!

vkovinicTT:
LGTM

PR #1227, opened 2025-02-14
Pilkic/fix docker build

pilkicTT:
In cases where the `tt-mlir` commit hash is specified in workflows, we don't respect it in the docker build job. So we end up with wrong docker image in the rest of the workflow.

Fix: check if the `tt-mlir` commit hash is overriden and use it in the docker build job.


.github/workflows/build-image.yml
```diff
@@ -25,6 +25,21 @@ jobs:
             submodules: recursive
             fetch-depth: 0 # Fetch all history and tags
 
+      - name: Update submodule if mlir_override is set
+        if: ${{ inputs.mlir_override }}
+        run: |
+          cd third_party/tt-mlir
+          git fetch origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/*
```
pilkicTT, line 32: @vmilosevic I am not sure how the build job doesn't need these additional flags... When testing this fix, I couldn't reference commits which were not on `origin/main` (the fetch command is only fetching from `origin/main`).

.github/workflows/build-image.yml
```diff
@@ -25,6 +25,21 @@ jobs:
             submodules: recursive
             fetch-depth: 0 # Fetch all history and tags
 
+      - name: Update submodule if mlir_override is set
+        if: ${{ inputs.mlir_override }}
```
vmilosevic, line 29: We also need to add mlir_override to workflow inputs, and pass it down from on-pr workflow

.github/workflows/build-image.yml
vmilosevic: I believe git fetch fetches all remote tracking branches (origin/*), and it should be already up to date after actions/checkout recursive anyways.
Maybe this should better go after "Cleanup submodules" step?

.github/workflows/build-image.yml
pilkicTT: Yeah, it somehow works without it, but I'll add it anyway...

.github/workflows/build-image.yml
```diff
@@ -2,7 +2,17 @@ name: Build Docker Image
 
 on:
   workflow_dispatch:
+    inputs:
+      mlir_override:
```
vmilosevic, line 6: Change on-pr.yaml to pass in mlir_override.
We call on-pr.yaml workflow from tt-mlir uplift workflow, and it needs to pass override to build-image.yaml

vmilosevic (approved):

PR #1239, opened 2025-02-18
Operator repository for Forge and Torch

vbrkicTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist

### DIFF
--- a/forge/forge/op_repo/forge_operators.py
+++ b/forge/forge/op_repo/forge_operators.py
@@ -1,0 +1,108 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+
+# Forge repostiory operators
+
+
+from .datatypes import OperatorDefinition, OperatorRepository
+from .datatypes import OperatorParamNumber
+
+
+# TODO describe operand and shapes
+_OPERATORS = [
+    # Unary operators
+    OperatorDefinition("exp", "forge.op.Exp", 1),
+    OperatorDefinition("reciprocal", "forge.op.Reciprocal", 1),
+    OperatorDefinition("buffer", "forge.op.Buffer", 1),
+    OperatorDefinition("sqrt", "forge.op.Sqrt", 1),
+    OperatorDefinition("relu", "forge.op.Relu", 1),
+    OperatorDefinition(
+        "leaky_relu",
+        "forge.op.LeakyRelu",
+        1,
+        forward_params=[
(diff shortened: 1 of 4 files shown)


forge/forge/op_repo/forge_operators.py
```diff
@@ -0,0 +1,108 @@
...
+    OperatorDefinition("power", "forge.op.Power", 2),
+    OperatorDefinition("greater", "forge.op.Greater", 2),
+    OperatorDefinition("greater_equal", "forge.op.GreaterEqual", 2),
+    OperatorDefinition("less", "forge.op.Less", 2),
+    OperatorDefinition("less_equal", "forge.op.LessEqual", 2),
+    OperatorDefinition("equal", "forge.op.Equal", 2),
+    OperatorDefinition("not_equal", "forge.op.NotEqual", 2),
+    OperatorDefinition("logical_and", "forge.op.LogicalAnd", 2),
+    # Nary operators
+    OperatorDefinition("where", "forge.op.Where", 3),
```
nvukobratTT, line 76: Does this mean that this nary operator has only 3 inputs? 

Similar question for others from the nary group, do we have a limited range we would like to cover for testing? I.e. do we want to limit some of the ops to 3, 10 or more inputs?

forge/forge/op_repo/forge_operators.py
```diff
@@ -0,0 +1,108 @@
...
+    ),
+    OperatorDefinition(
+        "stack",
+        "forge.op.Stack",
+        (2, 4),
+        forward_params=[
+            OperatorParamNumber("axis", int, 1, 10),
+        ],
+    ),
+    OperatorDefinition("matmul", "forge.op.Matmul", 2),
```
nvukobratTT, line 103: Is this nary group?

forge/forge/op_repo/forge_operators.py
vbrkicTT: Where operator can have only 3 inputs. However below we have interleave, concatenate and stack which have variable number of operands defined via a tuple (min, max)

forge/forge/op_repo/forge_operators.py
vbrkicTT: No matmul usually have its own group together with sparse matmul.

Atm in operator repository we don't specify groups, it's just comments to make navigation through file little bit easier.

But maybe in future we could introduce explicit grouping if needed. Atm for RGG there is no need for grouping/separation.

nvukobratTT (approved):

PR #1249, opened 2025-02-20
Store execution phase and stage inside tags property as dict

chandrasekaranpradeep:
Instead of storing execution phase and execution stage as two different property inside the xml report, created property named "tags" to store execution phase and execution stage as dictionary


ashokkumarkannan1 (approved):
LGTM!!! Thanks for the PR @chandrasekaranpradeep

forge/test/conftest.py
```diff
@@ -441,12 +441,15 @@ def pytest_runtest_makereport(item, call):
     report = outcome.get_result()
 
     if call.when == "call":  # Only execute after the test has run
-        # Store additional properties(i.e execution_phase and execution_stage) same as `record_property`
+        # Store additional properties(i.e tags which contains execution_phase and execution_stage as dictionaries) same as `record_property`
         execution_phase, execution_stage = fetch_execution_phase_and_stage()
+        tags = {}
         if execution_phase is not None:
```
ashokkumarkannan1, line 447: Just curious, Is there a chance of having execution_phase or execution_stage as None? I which case we skip this

chandrasekaranpradeep: There are some condition that has been there in `fetch_execution_phase_and_stage` function in `forge/forge/execution_tracker.py`  such as if the current_test we are running  not found in the json file or if the json file is deleted by some other code in between or not created, in those cases we will return None

PR #1252, opened 2025-02-20
Layernorm operator test plan

vobojevicTT:


forge/test/operators/pytorch/tm/test_layernorm.py
```diff
@@ -0,0 +1,222 @@
...
+
+    test_plan: TestPlan = None
+
+    operator = ["layer_norm"]
+
+    @classmethod
+    def generate_kwargs(cls, test_vector: TestVector):
+        yield {
+            # only normalization over the last dimension is currently supported
+            "normalized_shape": (test_vector.input_shape[-1],),
```
dgolubovicTT, line 130: Is this a metal constraint? So we can only do layer norm o last dimension?

vobojevicTT: Yes, we can do layer norm only over the last dimension. In tvm part is this check: https://github.com/tenstorrent/tt-tvm/blob/7a1ec5df5f6eaca8befbe8c00d2a023716c5c270/python/tvm/relay/frontend/pytorch.py#L1655

dgolubovicTT (approved):

PR #1234, opened 2025-02-17
Generate and update models ops tests and reimplement the models ops test failure updation script

vmilosevic:
### Summary:
1. Generated models ops tests by extracting the unique ops configurations across all the pytorch models present inside the forge/test/models directory path.
2. Skipped avgpool1d, avgpool3d, conv2d, pad and reshape ops test cases due to segmentation fault in CI and transpose ops test cases due to Insufficient host DRAM (requires 30 GB of memory)
3. Reimplemented the model ops test failure updation script:
           The `scripts/model_analysis/models_ops_test_failure_update.py` script automates the extraction of detailed test **failure information** from pytest logs by identifying test cases along with their statuses (**FAILED, XFAIL, or SKIP**) and corresponding **error messages**, then compiles this data into a comprehensive **Excel report** with columns for **test cases**, **marker statuses**, and **reasons**. This report enables users to manually refine error messages by updating the "Reason" column, and once revised, the script can be rerun using the `--use_report` option with the `report file path` to automatically update the model ops test files with marker and more descriptive failure messages
4. The adv_index ops test are failing with indexerror out of bound error(i.e `IndexError: index 563 is out of bounds for dimension 0 with size 448`) in evaluation. The adv_index ops takes two inputs one is reference tensor and other is indicies tensor based upon the indicies tensor values will pick the reference tensor in respective dim but the indicies values exceeds the range of reference tensor. To resolve these issue, added pytest param max_int which will be used by the create_from_shape method in forge/forge/tensor.py to generate the indices tensor values with the range of the reference tensor

### Process:
1. Triggered the nightly models ops pipeline for collecting the pytest logs
**Pipeline:** https://github.com/tenstorrent/tt-forge-fe/actions/runs/13536357567
**Collected logs:**
[test-log-n150-1](https://github.com/tenstorrent/tt-forge-fe/actions/runs/13536357567/artifacts/2653622726)
[test-log-n150-2](https://github.com/tenstorrent/tt-forge-fe/actions/runs/13536357567/artifacts/2653622477)
[test-log-n150-3](https://github.com/tenstorrent/tt-forge-fe/actions/runs/13536357567/artifacts/2653619334)
[test-log-n150-4](https://github.com/tenstorrent/tt-forge-fe/actions/runs/13536357567/artifacts/2653707424)

2. Using the collected pytest logs, generated excel report  by using `scripts/model_analysis/models_ops_test_failure_update.py` script
**Command:** 
[...]

PR #1319, opened 2025-02-25
Update record_forge_property fixture to store property inside the ForgePropertyStore and update it after test execution

chandrasekaranpradeep:


ashokkumarkannan1 (approved):
Great one. LGTM! Thanks for the PR @chandrasekaranpradeep

chandrasekaranpradeep:
> @chandrasekaranpradeep For CI failures, pls rerun the failed jobs alone.

Sure Thanks for the info

nvukobratTT (approved):

forge/test/conftest.py
```diff
@@ -88,6 +89,116 @@ def set_environment_variable(request):
...
+    A class to manage properties using dot notation for nested keys.
+
+    Attributes:
+        forge_properties (dict): A dictionary that stores properties.
+    """
+
+    def __init__(self):
+        self.forge_properties = {}
+
+    def add(self, key: str, value: Any):
```
ashokkumarkannan1, line 103: @chandrasekaranpradeep  can we also add `__call__` method here. so that we don't need to change anything in the model files.

```python
def __call__(self, key, value):
    return self.add(key, value)
```

This will also make sure we don't get any surprising conflicts from this PR - https://github.com/tenstorrent/tt-forge-fe/pull/1234 . This have record_forge_property in this format `record_forge_property("op_name", "Abs")`. With this `__call__` method, we can have the same format, which we have now. What do you say?

cc: @nvukobratTT

chandrasekaranpradeep: Thats good idea to have `__call__` method for `ForgePropertyStore` class but model files will change because we are storing model_name inside the tags using dot notation

ashokkumarkannan1: Thanks @chandrasekaranpradeep

PR #1323, opened 2025-02-25
[test] add e2e on device training test

pilkicTT:
Adding MNIST training test which runs everything on device, i.e. fwd/bwd of the model and loss function, as well as the optimizer.

Additionally, modified the `test_lora` test to use optimizer on device.

### DIFF
--- a/forge/test/mlir/mnist/training/test_training.py
+++ b/forge/test/mlir/mnist/training/test_training.py
@@ -19,8 +19,6 @@ from test.mlir.utils import *
 
 @pytest.mark.push
 def test_mnist_training():
-    torch.manual_seed(0)
-
     # Model and data type.
     # For bfloat16, the following line should be added to the test_forge_vs_torch function:
     # In file forge/forge/op/eval/forge/eltwise_unary.py:418 should be replaced with: threshold_tensor = ac.tensor(torch.zeros(shape, dtype=torch.bfloat16) + threshold)
@@ -100,8 +98,6 @@ def test_mnist_training():
 
 @pytest.mark.push
 def test_mnist_training_with_grad_accumulation():
-    torch.manual_seed(0)
-
     # Config
     num_epochs = 3
     batch_size = 1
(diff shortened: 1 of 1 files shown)


forge/test/mlir/mnist/training/test_training.py
```diff
@@ -537,3 +529,72 @@ def test_optimizer_device():
...
+        total_loss = 0
+        for batch_idx, (data, target) in enumerate(train_loader):
+
+            # Create target tensor and leave on CPU.
+            target = nn.functional.one_hot(target, num_classes=10).float()
+
+            # Forward pass (prediction) on device.
+            pred = tt_model(data)[0]
+            golden_pred = framework_model(data)
+            assert compare_with_golden(golden_pred, pred, pcc=0.95)
```
nvukobratTT, line 572: Can you use standard `verify` metho?

nvukobratTT (approved):
Looks good, one minor comment.

forge/test/mlir/mnist/training/test_training.py
pilkicTT: we need the outputs from the model, to pass into loss function, which the `verify()` function does not provide... we can probably expand the `verify()` method to return the values, but I would leave that for later...

nvukobratTT: https://github.com/tenstorrent/tt-forge-fe/issues/1331

pglusacTT (approved):
Great, just one question

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -451,16 +441,20 @@ def test_lora():
...
             tt_loss.backward()
 
-            # Adjust weights (on CPU)
-            framework_optimizer.step()
+            # Adjust weights on the device.
+            # NOTE: after executing the step, this will also zero the gradients.
+            tt_optimizer.step()
 
         print(f"epoch: {epoch_idx} loss: {total_loss}")
+        assert prev_total_loss - total_loss > 1.0, "Loss should go down"
```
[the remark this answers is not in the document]
pglusacTT: If our main concern is floating point issues, I suggest doing something like `assert prev_total_loss > total_loss + 1e-5`, what do you think of that? I think we should be safe in most cases by doing this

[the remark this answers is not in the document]
pilkicTT: Done!

Issue #952, opened 2024-12-23
Invalid Runtime inputs to embedding


dgolubovicTT:
I am getting data mismatch for embedding op when input indices are int64. Turns out that input indices are invalid right before call to ttnn::embedding in runtime: `third_party/tt-mlir/runtime/lib/ttnn/operations/embedding/embedding.cpp`

Namely, input indices in torch look like this:

`16044   8239   2933  13760  16963  16379  31427   6503  31497   9683  14101 26866`

In `third_party/tt-mlir/runtime/lib/ttnn/operations/embedding/embedding.cpp` run method they look like this:

`16044  0  8239  0  2933  0  13760  0  16963  0  16379  0 , data type is UINT32`

So every other index is now 0, which causes embedding to return invalid rows from embedding matrix. 

I've verified that input tensor is still valid in `forge/csrc/runtime/runtime.cpp::run_binary`, so the issue is somewhere between.

To repro:

checkout: dgolubovic/repro-embedding-input-indices-issue
run: pytest -svv forge/test/mlir/test_ops.py::test_embedding

One additional note: If the input indices are int64:

```
    inputs = [
        torch.randint(0, vocab_size, (1, token_num)),
    ]
```
This issue occurs. However, If input indices is int32, everything works.
```
    inputs = [
        torch.randint(0, vocab_size, (1, token_num),dtype=torch.int32),
    ]
```

On the other hand, ttnn::embedding op accepts dtype=DataType::UINT32, so runtime tensor has to have that datatype. @jnie-TT do you know where do we do this datatype conversion from pytorch tensors to ttnn.Tensor?

jnie-TT:
@dgolubovicTT can you also include the dumped mlir graph here as well in the ttnn dialect?

dgolubovicTT:
Ofc. It is the same in both cases (input indices int32 and int64), which is expected.

[Embedding_test_ops_ttnn.txt](https://github.com/user-attachments/files/18233803/Embedding_test_ops_ttnn.txt)

jnie-TT:
The issue is probably coming from this function:
```
static target::DataType torch_scalar_type_to_dt(torch::ScalarType st)
{
    switch (st)
    {
        case torch::ScalarType::Byte: return target::DataType::UInt8;
        case torch::ScalarType::Char: return target::DataType::UInt8;
        case torch::ScalarType::Short: return target::DataType::UInt16;
        case torch::ScalarType::Int: return target::DataType::UInt32;
[...]

PR #1335, opened 2025-02-26
Cast torch input tensors to dtype that is supported in forge. 

dgolubovicTT:
Remove manual casting in tests.

### Ticket
Issue #952 

### Problem description
There was a bug when input tensors have dtype that is not supported in Forge (for example int64). Compiler handled inputs like they were some lower dtype, while actual runtime tensors stayed in original dtype. This is explained in detail in issue #952

### What's changed
Now we cast torch tensor to appropriate dtype based on forge support.

### DIFF
--- a/forge/test/models/pytorch/audio/whisper/test_whisper_large_v3.py
+++ b/forge/test/models/pytorch/audio/whisper/test_whisper_large_v3.py
@@ -66,7 +66,6 @@ def test_whisper_large_v3_speech_translation(record_forge_property, variant):
 
     # Get decoder inputs
     decoder_input_ids = torch.tensor([[1, 1]]) * model_config.decoder_start_token_id
-    decoder_input_ids = decoder_input_ids.to(torch.int32)
     encoder_outputs = framework_model.model.model.encoder(input_features)[0].detach()
     encoder_outputs = encoder_outputs.to(torch.float32)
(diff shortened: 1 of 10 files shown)


forge/test/models/pytorch/audio/whisper/test_whisper_large_v3.py
```diff
@@ -40,7 +40,7 @@ def forward(self, decoder_input_ids, encoder_hidden_states):
 @pytest.mark.nightly
 @pytest.mark.parametrize("variant", ["openai/whisper-large-v3-turbo"])
 def test_whisper_large_v3_speech_translation(record_forge_property, variant):
-    pytest.skip("Skipping due to the current CI/CD pipeline limitations")
+    # pytest.skip("Skipping due to the current CI/CD pipeline limitations")
```
nvukobratTT, line 43: Why we're un-skipping this one?

nvukobratTT (approved):
Looks good! Make sure to remove xfail from one of the passing tests :))

pilkicTT (approved):

forge/test/models/pytorch/audio/whisper/test_whisper_large_v3.py
dgolubovicTT: left it commented out by mistake, when I was testing this change. will uncomment.

PR #1341, opened 2025-02-27
[api] add API for exporting model as c++ source

### LINKED ISSUE #1340: [api] add user API for exporting compiled model as c++ source
`tt-mlir` supports generating C++ code via the `emitc` pipeline.

Expose API for utilizing this feature to the user.

E.g.:

```python
compiled_model = forge.compile(model, ...)

# Export to c++ file, `user_model.cpp`
file_name = "user_model.cpp"
compiled_model.export_to_cpp(file_name)
```

pilkicTT:
Adding support for exporting compiled model as c++ source, via the MLIR `emitc` pipeline.

To export the model, the user would call `export_to_cpp()` on the already compiled model (`CompiledModel` instance). This means that there are no changes in the compile flow.

The `CompiledModel` will hold the `ForgeGraphModule` generated during the compile, so that it can execute the needed `tt-mlir` pipeline for generating c++ code on demand. The other option would be to hold the generated `ttir` text (during compile) in memory, so that the pipeline can be ran from that point onwards. I've chosen the first option, since it's simpler to fit in the existing flow, and I don't see a big improvement in the other approach.

Existing code for running the MLIR compiler is refactored a bit to allow generating different outputs (flatbuffer or cpp) without duplicating the logic.

Simple test is added to confirm the C++ generation completes successfully.

Example usage:
```python
compiled_model.export_to_cpp(file_name)
```

Closes #1340
Closes https://github.com/tenstorrent/tt-mlir/issues/1944

### DIFF
--- a/forge/forge/compile.py
+++ b/forge/forge/compile.py
@@ -393,6 +393,7 @@ def forge_compile_from_context(context: CompileContext) -> CompiledModel:
     assert context.compiled_binary is not None
 
     compiled_module = CompiledModel(
+        context.forge_module,
         fwd_compiled_graph_state,
         bwd_compiled_graph_state,
         opt_compiled_graph_state,
--- a/forge/forge/compiled_graph_state.py
+++ b/forge/forge/compiled_graph_state.py
@@ -10,9 +10,11 @@ import torch
 from typing import Dict, List, Any, Optional
 
 
+from forge._C import ForgeGraphModule
 from forge._C.graph import Graph
 import forge._C.graph as pygraph
 from forge._C.runtime import run_binary, Binary
+from forge._C import run_mlir_compiler_to_cpp
 from forge.tensor import Tensor, get_post_const_eval_tensors, to_pt_tensors, cast_unsupported_torch_dtype, AnyTensor
 from forge.module import Module, PyTorchModule, AnyModule
 from forge.execution_tracker import ExecutionPhase, record_execution_phase_and_stage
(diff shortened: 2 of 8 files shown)


forge/forge/compile.py
```diff
@@ -393,6 +393,7 @@ def forge_compile_from_context(context: CompileContext) -> CompiledModel:
     assert context.compiled_binary is not None
 
     compiled_module = CompiledModel(
+        context.forge_module,
```
nvukobratTT, line 396: Can we document why we're including `context.forge_module` next to existing components within compiled module? It can be confusing at first why it's included next to other components in this fn.

nvukobratTT (approved):

forge/forge/compile.py
pilkicTT: Added comment in the `CompiledModel` class and filed an issue to clean this up after the runtime refactor.

forge/forge/compiled_graph_state.py
```diff
@@ -392,3 +402,21 @@ def step(self) -> None:
...
+            Path to the file where the model c++ code will be exported.
+        """
+
+        logger.info("Exporting model {self.framework_module.get_name()} to cpp file...")
+        cpp_code = run_mlir_compiler_to_cpp(self.forge_graph_module)
+
+        with open(export_path, "w") as f:
+            f.write(cpp_code)
+
+        logger.info(f'Exported model as cpp file to "{export_path}"')
```
svuckovicTT, line 422: As discussed offline, let's add another info log here.

Path to docs: https://docs.tenstorrent.com/tt-mlir/ttnn-standalone.html
Path to the tool: https://github.com/tenstorrent/tt-mlir/tree/main/tools/ttnn-standalone

So something like:
```
logger.info(f'To compile and run this code, one can utilize the ttnn-standalone tool within the tt-mlir project. It provides all the necessary build and run scripts. Copy the contents of the .cpp to ttnn-standalone.cpp and use `./run` to compile&run the code.')
logger.info(f'    Path to tool: https://github.com/tenstorrent/tt-mlir/tree/main/tools/ttnn-standalone)
logger.info(f'    Path to docs: https://docs.tenstorrent.com/tt-mlir/ttnn-standalone.html)
```

svuckovicTT (approved):

PR #1344, opened 2025-02-27
Update wheel build test steps

jmcgrathTT:
Updated docs with wheel and renamed old test to source build steps

### DIFF
--- a/docs/src/getting-started.md
+++ b/docs/src/getting-started.md
@@ -17,6 +17,15 @@ pip install *.whl
 
 > Note: Make sure to run the command from the directory where wheels are downloaded.
 
+#### Run Wheel test
+
+To confirm that your wheel is installed properly, you can run this command.
+
+```bash
+curl -o forge/test/mlir/test_ops_tf.py https://raw.githubusercontent.com/tenstorrent/tt-forge-fe/main/forge/test/test_api.py
+pytest -svv forge/test/
+```
+
 ### Build from Source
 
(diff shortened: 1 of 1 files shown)


docs/src/getting-started.md
```diff
@@ -17,6 +17,15 @@ pip install *.whl
 
 > Note: Make sure to run the command from the directory where wheels are downloaded.
 
+#### Run Wheel test
+
+To confirm that your wheel is installed properly, you can run this command.
+
+```bash
+curl -o forge/test/mlir/test_ops_tf.py https://raw.githubusercontent.com/tenstorrent/tt-forge-fe/main/forge/test/test_api.py
```
nvukobratTT, line 25: Instead downloading the test and running it, we can just showcase a demo test that they can try out. For example:
```py
# Create a simple model that adds two tensors
class Add(nn.Module):
    def forward(self, a, b):
        return a + b

# Create some sample inputs
inputs = [torch.rand(2, 32, 32), torch.rand(2, 32, 32)]

# Compile the model
compiled_model = forge.compile(Add(), sample_inputs=inputs)

# Run the model
output = compiled_model(*inputs)

print(output)
```

jmcgrathTT: Agree! I am going to close this issue out since it's being tracked by in our technical documentation issue now.

https://github.com/tenstorrent/tt-forge/issues/40?issue=tenstorrent%7Ctt-forge%7C29

PR #1348, opened 2025-02-28
verify now returns outputs from forward pass + tests rework

### LINKED ISSUE #1331: Expand verify to return golden and device calculated values (optional)
## Summary
For use cases where we use outputs after specific verification, it's neat to be able to return values from the `verify` method and reuse them instead of having to generate them again after verification.

Here is one example where this logic would be beneficial:
- https://github.com/tenstorrent/tt-forge-fe/pull/1323/files#diff-ae6d94ee42280fb60339a5657934b048102b824640dee0dd61dc6296995f0caaR572

Once this feature is in place, let's make a change to the corresponding place as well.

vkovinicTT:
### Ticket
Fix #1331 

### Problem description
`verify()` needed to return outputs from the forward pass, as there are some use cases where we need outputs of forward pass.

### What's changed
Changed `verify()` to return outputs from framework and compiled model and also updated tests.

### DIFF
--- a/forge/forge/verify/verify.py
+++ b/forge/forge/verify/verify.py
@@ -270,15 +270,24 @@ def verify(
     verify_cfg: VerifyConfig = VerifyConfig(),
 ):
     """
-    Verify the compiled model against the framework model
+    Performs verification of a compiled model by comparing its outputs against a reference framework model.
+
+    Runs inference on both models with the same inputs and performs various validation checks
+    based on the provided verification configuration. Checks can include output size matching,
+    dtype consistency, shape equivalence, and numeric value comparison.
+
+    Parameters:
+        inputs: List of tensor inputs
+        framework_model: Reference model
+        compiled_model: compiled model to verify
+        verify_cfg: Configuration object controlling which verification checks to perform
+
+    Returns:
+        tuple: (framework_outputs, compiled_outputs) - outputs from both models
+               Returns (None, None) if verification is disabled
     """
-    if not verify_cfg.enabled:
-        logger.warning("Verification is disabled")
-        return
 
-    # 0th step: input checks
-
-    # Check if inputs are of the correct type
+    # 0th step: Check if inputs are of the correct type
     if not inputs:
         raise ValueError("Input tensors must be provided")
     for input_tensor in inputs:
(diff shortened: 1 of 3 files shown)


vkovinicTT:
For me personally it's easier to just have:
```
pred = tt_model(data)[0]
golden_pred = framework_model(data)
assert golden_pred.dtype == dtype
assert compare_with_golden(golden_pred, pred, pcc=0.95)
```
then what I have currently set it to:
```
golden_pred, pred = verify(inputs=[data], framework_model=framework_model, compiled_model=tt_model, verify_cfg=VerifyConfig(value_checker=AutomaticValueChecker(pcc=0.95)))
pred = pred[0]
```

as it is not really clear based on the method name `verify` (it becomes clear when you read method's description though) that it returns outputs and it feels easier to understand code with the former approach even though there are more lines of code. 
What do you guys think? @nvukobratTT @pglusacTT

forge/forge/verify/verify.py
```diff
@@ -333,3 +347,6 @@ def verify(
             verify_cfg.value_checker.check(fw, co)
 
     record_execution_phase_and_stage(ExecutionStage.VERIFICATON)
+
+    # Return both the framework and compiled model outputs
+    return fw_out, co_out
```
pilkicTT, line 352: No need to return both golden and compiled model outputs. Just return the compiled model outputs...

pilkicTT (approved):
Looks good, just follow up on the return values...

forge/forge/verify/verify.py
pglusacTT: What if we need to do `.backward()` on the golden tensor we we want to test training?

pilkicTT: Yeah, you're right... @vkovinicTT ignore this, we actually need both outputs... 😄

forge/forge/verify/verify.py
```diff
@@ -270,11 +270,25 @@ def verify(
...
+        verify_cfg: Configuration object controlling which verification checks to perform
+
+    Returns:
+        tuple: (framework_outputs, compiled_outputs) - outputs from both models
+               Returns (None, None) if verification is disabled
     """
     if not verify_cfg.enabled:
         logger.warning("Verification is disabled")
-        return
+        return None, None
```
ndrakulicTT, line 291: This would break code in `forge/test/mlir/mnist/training/test_training.py` if for some reason verify is disabled. pred would be None and then loss will raise an error

vkovinicTT: You are right, I will run forward pass either way and then return outputs without any further inspection if `verify_cfg.enabled = False`.

Issue #614, opened 2024-11-04
Making sure forge runs on Blackhole


gfengTT:
This issue will be updated with the steps I take to build and run forge on blackhole, and problem that I run into.

staylorTT:
can we use the ![Image](https://github.com/user-attachments/assets/8ef0a666-ccea-4ebc-a6fa-5deca52b2396)
 
to track the sub items needed to enable this?

zzigler-tt:
@gfengTT Can you please add the following labels - priority, Blackhole and Forge. Thanks

gfengTT:
@staylorTT Sure. The issue I described requires a fix on the metal side. Should I create an issue there? Do you know who I should assign it to? Just realized I don't have push permissions on metal.

staylorTT:
When you made the changes in your previous [comment](https://github.com/tenstorrent/tt-forge-fe/issues/614#issuecomment-2457690643) did this pass locally? Is this a good sign of life that the test can pass using BH?

As far as sending the change, I believe because metal is public you could fork and open an MR with the changes? I believe @abhullar-tt
might be able to help with this

staylorTT:
Is this issue closed with the PR that was merged?

gfengTT:
The compilation issue is closed. I'm running into a new issue though. Working on summarizing the findings.

staylorTT:
Now that we have BH in CI @gfengTT is working to get tests in a state where we can start running BH CI to identify/prevent regressions

PR #1352, opened 2025-02-28
Add Blackhole CI

gfengTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-forge-fe/issues/614)

### Problem description
Add Blackhole CI tests.

### What's changed
1. Added tests on push. Need to be triggered manually, will add to automated tests once it's stable.
2. Labeled tests that fail on BH with `xfail`.

### DIFF
--- a/forge/test/mlir/mnist/training/test_training.py
+++ b/forge/test/mlir/mnist/training/test_training.py
@@ -2,6 +2,7 @@
 
 # SPDX-License-Identifier: Apache-2.0
 
+import os
 import pytest
 import time
 
@@ -354,6 +355,9 @@ def test_forge_vs_torch(forge_property_recorder):
 
 @pytest.mark.push
 def test_loss_device(forge_property_recorder):
+    if os.environ["ARCH_NAME"] == "blackhole":
+        pytest.xfail()
+
     # Config
     num_epochs = 3
     batch_size = 1
@@ -451,6 +455,9 @@ def test_loss_device(forge_property_recorder):
 
 @pytest.mark.push
 def test_lora(forge_property_recorder):
+    if os.environ["ARCH_NAME"] == "blackhole":
+        pytest.xfail()
+
     # Config
     num_epochs = 3
     batch_size = 128
@@ -537,6 +544,9 @@ def test_lora(forge_property_recorder):
 
 @pytest.mark.push
 def test_optimizer_device(forge_property_recorder):
+    if os.environ["ARCH_NAME"] == "blackhole":
+        pytest.xfail()
+
     # Config
     num_epochs = 32
     batch_size = 1024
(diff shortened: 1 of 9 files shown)


.github/workflows/on-blackhole-push.yml
```diff
@@ -0,0 +1,29 @@
+name: On push
```
jmcgrathTT, line 1: Was this file just created for push testing?

nvukobratTT, line 1: For starters, I would propose to have a nightly pipeline that runs tests of choices for BH. As BH is still unstable, it'll just slow down current development. That said, I'm suggestion to create new nightly, track it via Superset, and determine when it's in stable state to include here as well.

forge/test/mlir/mnist/training/test_training.py
```diff
@@ -332,6 +333,9 @@ def test_forge_vs_torch():
 
 @pytest.mark.push
 def test_loss_device():
+    if os.environ["ARCH_NAME"] == "blackhole":
```
nvukobratTT, line 336: I wouldn't use env vars to target blackhole device.  We need core logic that will enable us to clearly manage test expectations on different HW devices. Right now, we only have focus on WH. However, as BH is right around the corner, we should think about how to handle different device types. 

This is probably something for @gfengTT to look into. And please, before making any drastic changes, let's create a Draft PR with small PoC, and details what options we have, proposal, etc. Do you think you can do something like that? If not, we'll put it on our roadmap as well.

nvukobratTT (commented):
@jmcgrathTT thanks for pushing this change further! 

I left few comments on what approach I think is better for starter :))

forge/test/mlir/mnist/training/test_training.py
nvukobratTT: @pilkicTT can provide feedback and help out if needed :)) As I'm OOF, feel free to sync on this one once you're back from vacation @gfengTT

.github/workflows/on-blackhole-push.yml
gfengTT: Doing so should add the tests as optional, on push tests. You need to trigger them manually, like metal's BH post commit tests.

I like the idea of adding nightly tests as well. But the two test sets ideally should not overlap too much.

forge/test/mlir/mnist/training/test_training.py
gfengTT: Yes we should find a better way of telling whether the test is running on a BH device.

FYI, I tried `get_tenstorrent_device()` in forge/forge/forgeglobal.py but got the following error:
```
>       from forge.ttdevice import TTDevice
E       ModuleNotFoundError: No module named 'forge.ttdevice'
```
@pilkicTT any idea how this should be fixed?

nvukobratTT:
@gfengTT, as we're working on this PR https://github.com/tenstorrent/tt-forge-fe/pull/2102, can we close this one?

PR #1354, opened 2025-03-02
Uplift third_party/tt-mlir to origin/main 2025-03-02 

mstojkovicTT:
This PR uplifts the third_party/tt-mlir submodule to the origin/main and fixes problems that occured during uplift.


forge/test/mlir/llama/tests/test_llama_embedding.py
```diff
@@ -9,7 +9,18 @@
 from forge.verify.verify import verify
 
 
-@pytest.mark.parametrize("model_path", ["openlm-research/open_llama_3b", "meta-llama/Llama-3.2-1B"])
+@pytest.mark.parametrize(
+    "model_path",
+    [
+        pytest.param(
+            "openlm-research/open_llama_3b",
+            marks=pytest.mark.xfail(reason="data type mismatch: expected Int32, got UInt32"),
```
pilkicTT, line 17: Let's look into why we have data type mismatch. At this point, I think we shouldn't regress models with uplifts...

pilkicTT (changes requested):

forge/forge/op/eval/forge/convolution.py
```diff
@@ -262,27 +240,41 @@ def eval(self, tensors):
...
 
-        output_padding_height = 0
-        output_padding_width = 0
+        output_padding = self.output_padding
+        if output_padding != 0:
+            output_padding_height = output_padding[0]
+            output_padding_width = output_padding[1]
+        else:
+            output_padding_height = 0
+            output_padding_width = 0
```
vkovinicTT, line 265: Do we have this `self.output_padding` param. Because if we don't would the code break here?

pilkicTT (approved):

PR #1360, opened 2025-03-04
Softmax operator - switch to use test plan

### LINKED ISSUE #1433: [Feature] Improve test framework to support excluding collection
Test framework should support excluding TestCollection from TestCollection in parent - child manner.

This will be useful in case of skipping some tests from the test plan. This is a common scenario that skipping some bug caused failures to improve sweeps pipeline performance. We want to `xfail` one or few of tests to act as check flag. If bug is fixed `xfail` will become `xpass`.

vobojevicTT:
Closes #1433


forge/test/operators/pytorch/nn/test_softmax.py
```diff
@@ -0,0 +1,184 @@
...
+    failing_rules=[
+        # All dim values are not supported except for the last one:
+        TestCollection(
+            operators=TestParamsData.operators,
+            criteria=lambda test_vector: test_vector.kwargs["dim"] != len(test_vector.input_shape) - 1
+            and test_vector.kwargs["dim"] != -1,
+            skip_reason=FailingReasons.UNSUPORTED_AXIS,
+        ),
+        # All-close value checker failed (rtol=1e-2, atol=1e-2):
+        TestCollection(
```
vbrkicTT, line 160: Loading test ids from lambda raise performance issues

vobojevicTT: @vbrkicTT Thanks, switched to use another approach.

dgolubovicTT (approved):

PR #1363, opened 2025-03-04
Update tvm commit hash retrieval for out-of-tree execution when FORGE_ENABLE_TVM_CACHE=1

sott0n:
### Ticket
N/A

### Problem description
The tvm cache path retrieval assumed execution within the Forge tree, causing an error when run out-of-tree as the tvm submodule could not be found.

### What's changed
This change modifies the commit hash retrieval method to use the path from `tvm.__file__`, allowing the cache to be saved correctly even in out-of-tree scenarios. Additionally, the commit hash retrieval is now performed using `git rev-parse` for improved cleanup.

### Checklist

### DIFF
--- a/forge/forge/tvm_calls/forge_compile.py
+++ b/forge/forge/tvm_calls/forge_compile.py
@@ -1300,15 +1300,17 @@ def get_auto_path(graph_hash, compiler_cfg, is_load):
         else:
             tvm_path = os.path.dirname(tvm.__file__)
             if "site-packages" in tvm_path:
-                tvm_info = subprocess.check_output(['pip', 'show', 'tvm']).decode('utf-8')
-                for line in tvm_info.split('\n'):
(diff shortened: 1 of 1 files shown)


sott0n:
At `action-junit-report` job, it seems to fail because the token does not have permission to access the resources.

sott0n:
Hi @pilkicTT, my account doesn't have a permission to write `tenstorrent/tt-forge-fe`. Could you give me a write permission? (NOTE: i'm a memmber in Tenstorrent org)

And currently, this check ci fails for external PRs. As a workaround, it might be practical to add `annotate_only` to generate only annotations for external PRs.

> In environments that do not allow checks: write, the action can be configured to leverage the annotate_only option.

https://github.com/mikepenz/action-junit-report?tab=readme-ov-file#pr-run-permissions

PR #1381, opened 2025-03-06
Add resnet to benchmark models

vcanicTT:
We want to track the end-to-end performance of different models, so we have added ResNet HF to the benchmark models. This PR also includes a few changes to existing models, aiming to align all models to be written in the same manner and uploaded successfully.


vcanicTT:
This PR is not necessary anymore, we added resnet with llama in a different PR.

PR #1406, opened 2025-03-07
Added support for intermediate golden verification in forward pass only

### LINKED ISSUE #1405: Enable intermediate verification in forward pass
We need intermediate golden verification for debugging data mismatches effectively.
There are two scopes of graph verification:
First scope is verifying graph between compile stages. This enables us to detect in which stage of compile data mismatch occurs.
Second scope is verifying each intermediate tensor as opposed to verifying only outputs. This will enable us to locate where exactly data mismatch occurs in the graph. 

We had these features in pybuda, the goal is just to enable them in forge setting

dgolubovicTT:
Closes #1405 
- Added support for intermediate golden verification in forward pass only
- Added few simple tests in forge/test/mlir/test_golden_verification.py, more to come with backward support.

### Ticket
[Link to Github Issue
](https://github.com/tenstorrent/tt-forge-fe/issues/1405)

There are two scopes of graph verification:

1. First scope is verifying graph between compile stages. This enables us to detect in which stage of compile data mismatch occurs.
We can now set which compile stages to verify using:
`STAGES_TO_PERFORM_INTERMEDIATE_VERIFICATION` env variable. Specify the compile stages to perform verification in comma separated manner:
`STAGES_TO_PERFORM_INTERMEDIATE_VERIFICATION=OPTIMIZED_GRAPH, PRE_LOWERING_PASS`
Also, setting `FORGE_FORCE_VERIFY_ALL=1` will do verification for all compile stages.

2. Second scope is verifying each intermediate tensor as opposed to verifying only outputs. This will enable us to locate where exactly data mismatch occurs in the graph.
By setting `FORGE_OP_LEVEL_COMPARISON=1` intermediate golden outputs will be compared with intermediate graph outputs. Without setting this variable, verification will be done only on outputs (for compile stages chosen like explained above)

### DIFF
--- a/forge/csrc/graph_lib/python_bindings.cpp
+++ b/forge/csrc/graph_lib/python_bindings.cpp
@@ -43,7 +43,6 @@ eval_graph(
     Graph *graph,
     const std::vector<py::object> &inputs,
     const std::unordered_map<std::string, py::object> &parameters,
-    py::object tt_device,
     const std::unordered_map<int, py::object> &intermediate_golden_tensors,
     const std::vector<py::object> &losses,
     const std::vector<py::object> &targets,
@@ -596,7 +595,6 @@ void GraphModule(py::module &m_graph)
                     graph,
                     inputs,
                     parameters,
-                    tt_device,
                     intermediate_golden_tensors,
                     losses,
                     targets,
@@ -624,7 +622,6 @@ void GraphModule(py::module &m_graph)
         [](Graph *graph,
            const std::vector<py::object> &inputs,
            const std::unordered_map<std::string, py::object> &parameters,
-           py::object tt_device,
            float relative_atol,
            float pcc,
            const std::unordered_map<int, py::object> &intermediate_golden_tensors,
@@ -637,7 +634,6 @@ void GraphModule(py::module &m_graph)
                 graph,
                 inputs,
                 parameters,
-                tt_device,
                 intermediate_golden_tensors,
                 losses,
                 targets,
@@ -651,7 +647,6 @@ void GraphModule(py::module &m_graph)
         py::arg("graph"),
         py::arg("inputs"),
         py::arg("parameters"),
-        py::arg("tt_device"),
         py::arg("relative_atol"),
         py::arg("pcc"),
         py::arg("intermediate_golden_tensors") = std::unordered_map<int, py::object>(),
@@ -848,7 +843,6 @@ bool compare_tensor_to_golden(
     bool warning_only = false)
 {
     py::object eval_module = py::module_::import("forge.verify.compare");
-    bool is_forge = ir_level == graphlib::IRLevel::IR_FORGE;
 
     if (pcc == 0.0)
         return eval_module
@@ -856,7 +850,6 @@ bool compare_tensor_to_golden(
                 name,
                 golden,
                 calculated,
-                is_forge,
                 py::none(), /* rtol */
                 py::none(), /* atol */
                 py::none(), /* pcc */
@@ -870,7 +863,6 @@ bool compare_tensor_to_golden(
             name,
             golden,
             calculated,
-            is_forge,
             py::none(), /* rtol */
             py::none(), /* atol */
             pcc,
@@ -917,6 +909,9 @@ py::object eval_tms(py::object tensor, const std::vector<graphlib::OpType> &tms,
     return tensor;
 }
 
+/*
+Evaluates tms on input edges of current node, and returns modified inputs. Needs outputs of previous nodes node_outputs.
+*/
 std::vector<py::object> eval_operand_tms(
     graphlib::Graph *graph,
     graphlib::Node *node,
@@ -1052,11 +1047,11 @@ py::object eval_input(
     {
         return consteval_input(node, inputs, is_forge, epoch_type);
     }
-    else if (is_forge)
+
+    // Check if there is node->name() in inputs.
+    if (inputs.find(node->name()) == inputs.end())
     {
-        auto tensor_module = py::module_::import("forge.tensor");
-        auto pad_pytorch_tensor_to_forge = tensor_module.attr("pad_pytorch_tensor_to_forge");
-        return pad_pytorch_tensor_to_forge(inputs.at(node->name()), input->get_tile_broadcast_dims());
+        throw std::runtime_error("Error: Node name '" + node->name() + "' not found in inputs.");
     }
 
     return inputs.at(node->name());
@@ -1309,6 +1304,9 @@ bool compare_tensors(std::shared_ptr<void> tensor0, std::shared_ptr<void> tensor
     return compare_tensors_func(tensor0_pt, tensor1_pt).cast<bool>();
 }
 
+/*
+Returns a value (tensor) of input node if node is ConstantInputNode.
+*/
 py::object get_constant_input_value(graphlib::Node *node, bool is_forge)
 {
     TT_ASSERT(node->as<graphlib::InputNode>()->is_constant());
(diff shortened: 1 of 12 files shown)


forge/csrc/graph_lib/python_bindings.cpp
```diff
@@ -1369,53 +1368,82 @@ bool is_gradient_comparison_valid(Graph *graph, const graphlib::Edge &gradient_e
...
-        if (not input)
+        if (node->node_type() != NodeType::kInput)
             continue;
-
+        graphlib::InputNode *input = node->as<graphlib::InputNode>();
         if (input->is_constant())
         {
             graph_inputs.insert({input->name(), get_constant_input_value(input, is_forge)});
         }
-        else if (input->is_optimizer_parameter())
```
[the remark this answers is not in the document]
dgolubovicTT: This is was only for training, but I commented it out just to make forward work. I will add some form of handling optimizer parameter in backward verification PR...

forge/csrc/graph_lib/python_bindings.cpp
```diff
@@ -1369,53 +1368,82 @@ bool is_gradient_comparison_valid(Graph *graph, const graphlib::Edge &gradient_e
...
+//             if (opt_tensor.is_none())
+//                 log_fatal("optimizer_param key: {} not found for node: {}", optimizer_param_key,
+//                 optimizer_input_name);
+
+//             graph_inputs.insert({input->name(), opt_tensor.attr("value")().cast<py::object>()});
+//         }
+//     }
+
+//     return graph_inputs;
+// }
```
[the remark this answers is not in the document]
dgolubovicTT: yeah, until I add training verification too

forge/forge/config.py
```diff
@@ -82,8 +82,12 @@ class CompilerConfig:
     # Create "unsupported" forge ops in python file, allowing user to modify later
     enable_tvm_unsupported_ops: bool = False
 
-    # Should we need to compare every op with framework output at each compilation stage.
-    enable_op_level_comparision: bool = False
+    # Setting this to true will enable intermediate outputs to remain in graph
+    # If false, all unused outputs will be removed. This needs to be true for intermediate golden verification
+    # if we want to compare all intermediate tensors in graph.
+    enable_op_level_comparision: bool = (
```
pilkicTT, line 88: I would move this to the verify config.

forge/forge/compile.py
```diff
@@ -239,7 +239,7 @@ def compile_main(
         modules=modules,
         graph_name=module_name,
         compiler_cfg=compiler_cfg,
-        verify_cfg=DepricatedVerifyConfig.disabled(),
+        verify_cfg=DepricatedVerifyConfig(),
```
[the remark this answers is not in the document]
dgolubovicTT: Currently, this DepricatedVerifyConfig has everything related to verification. I want to enable verification first, then maybe merge into VerifyConfig. I didn't want to do both at the same time...

forge/forge/config.py
```diff
@@ -170,6 +174,8 @@ def apply_env_config_overrides(self):
 
         if "FORGE_CONVERT_PARAMS_TO_TVM" in os.environ:
             self.convert_framework_params_to_tvm = bool(int(os.environ["FORGE_CONVERT_PARAMS_TO_TVM"]))
+        if "FORGE_FORCE_VERIFY_ALL" in os.environ:
```
pilkicTT, line 177: it would be great if we could remove these env variables.

forge/csrc/graph_lib/python_bindings.cpp
```diff
@@ -1452,10 +1479,9 @@ eval_graph(
     std::unordered_map<std::string, py::object> updated_parameter_mapping;
     std::unordered_map<std::string, py::object> intermediate_tensors;
 
-    const bool is_forge = graph->get_ir_level() == graphlib::IRLevel::IR_FORGE;
+    std::unordered_map<std::string, py::object> graph_inputs = get_graph_input_mapping(graph, inputs, parameters);
 
-    auto optimizer = tt_device.attr("get_optimizer")();
-    std::unordered_map<std::string, py::object> graph_inputs = get_graph_input_mapping(graph, parameters, optimizer);
+    const bool is_forge = graph->get_ir_level() == graphlib::IRLevel::IR_FORGE;
```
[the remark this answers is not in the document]
dgolubovicTT: I tried to remove is_forge from functions that I have changed. I see some places where I missed it. What i don't remove in this PR, I will remove in separate one...

forge/forge/config.py
pilkicTT: We should strive to have `CompilerConfig` consist only of options which should be controllable by the end user.

forge/forge/compile.py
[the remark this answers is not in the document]
vkovinicTT: Awesome, sounds reasonable.

forge/forge/verify/config.py
```diff
@@ -86,6 +87,24 @@ class NebulaGalaxy:
     ]  # CI
 
 
+def get_stages_to_perform_intermediate_verification():
+    """
+    Get the stages to perform intermediate verification from the environment variable.
+    The environment variable should be a comma-separated list of stage names from CompileDepth enum class.
+    """
+    # Read and parse the environment variable
+    env_value = os.getenv("STAGES_TO_PERFORM_INTERMEDIATE_VERIFICATION", "")
```
pilkicTT, line 96: Let's not use env variables for setting the wanted stages. This should go through verify config API, so that we can specify specific values of the `CompileDepth` enum.

vkovinicTT (approved):

pilkicTT (changes requested):
Can you update the commit message and specify what is the intent of the change, is golden verification enabled on by default after this change? If yes, which type of verification.

forge/csrc/graph_lib/python_bindings.cpp
```diff
@@ -1369,53 +1368,82 @@ bool is_gradient_comparison_valid(Graph *graph, const graphlib::Edge &gradient_e
...
 
-static std::unordered_map<std::string, py::object> get_graph_input_mapping(
-    Graph *graph, const std::unordered_map<std::string, py::object> &parameters, py::object optimizer)
+std::unordered_map<std::string, py::object> get_graph_input_mapping(
+    Graph *graph, const std::vector<py::object> &inputs, const std::unordered_map<std::string, py::object> &parameters)
 {
-    const bool is_forge = graph->get_ir_level() == graphlib::IRLevel::IR_FORGE;
+    std::vector<std::string> ordered_input_names = graph->get_ordered_input_names();
     std::unordered_map<std::string, py::object> graph_inputs = parameters;
+    const bool is_forge = graph->get_ir_level() == graphlib::IRLevel::IR_FORGE;
```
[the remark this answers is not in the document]
dgolubovicTT: I didn't want to change `get_constant_input_value`... Again, will do the cleanup as the follow up to this PR.

forge/forge/config.py
dgolubovicTT: Makes sense. Done.

forge/forge/config.py
dgolubovicTT: I would like to have one env variable to specify verify stage, therefore `STAGES_TO_PERFORM_INTERMEDIATE_VERIFICATION`. I see some value in having `FORGE_FORCE_VERIFY_ALL` for debugging. Do you have some other way of configuring this, or you think we don't need them.

forge/forge/verify/config.py
dgolubovicTT: Just to be sure, do you suggest I transfer the logic to verify config (from depricated verify config) and specify stages to verify in constructor of verify config within the test?

forge/forge/tvm_calls/forge_compile.py
```diff
@@ -598,9 +598,7 @@ def compile_tvm_for_forge(
     mod, params = compile_for_forge(mod, target=target, params=params, graph_name=graph_name, **verify_args)
 
     if verify_cfg is not None and verify_cfg.verify_tvm_compile:
-        assert (
-            compiler_cfg.convert_framework_params_to_tvm
-        ), "Cannot verify TVM compile without converting framework params to relay"
+        compiler_cfg.convert_framework_params_to_tvm = True
```
dgolubovicTT, line 601: @pilkicTT can you give me your opinion on this: We had ASSERT for `compiler_cfg.convert_framework_params_to_tvm` value based on `verify_cfg.verify_tvm_compile`. So these variables are dependent. but first one clearly belongs to compiler config and second to verify. Frankly, I don't know why the assert... When I was setting values via env variables, I could easily set both variables when some env is set. Now I just swaped assert with `compiler_cfg.convert_framework_params_to_tvm = True`, but I don't like it...
Wanted to ping you on how we could handle this...

pilkicTT: So, from what I see, we are never using the `convert_framework_params_to_tvm` (it's off by default, and we are never setting it to True). I would assume it only makes sense in the verification scenario (`verify_tvm_compile=True`).

Hence, I would remove this setting from the `CompilerConfig` and whereever it is used, i would calculate it based on the `verify_tvm_compile` flag.

Let me know if that makes sense.

dgolubovicTT: This assert for convert_framework_params_to_tvm exists also in TVM constant prop. When `compiler_cfg.enable_tvm_constant_prop` is true, `convert_framework_params_to_tvm` has to be also. However, `enable_tvm_constant_prop` is also off by default. If de don't use const prop (const eval) in TVM, then `convert_framework_params_to_tvm` is indeed only required when `verify_tvm_compile` is true. From what I can see, we don't use tvm constant prop anymore? @nvukobratTT can you confirm/dispute this?

dgolubovicTT: Synced with @nvukobratTT offline. We intend on using constant prop in TVM, so I will leave `enable_tvm_constant_prop` and `convert_framework_params_to_tvm` as they are.

forge/test/mlir/test_golden_verification.py
```diff
@@ -0,0 +1,69 @@
...
+    "batch_size",
+    [
+        1,
+    ],
+)
+@pytest.mark.parametrize("outer_dim_x", [7])
+@pytest.mark.parametrize("outer_dim_y", [7])
+@pytest.mark.parametrize("inner_dim", [64])
+@pytest.mark.push
+def test_matmul_and_add(batch_size, outer_dim_x, outer_dim_y, inner_dim):
```
pilkicTT, line 22: do these tests actually test golden verification? i might be missing something, but i don't see `verify_all` or `stages_for_intermediate_verification` being set anywhere?

pilkicTT (approved):
I've left some comments, but approving anyways to unblock you once they are addressed.

forge/csrc/graph_lib/python_bindings.cpp
```diff
@@ -1786,6 +1793,7 @@ eval_graph(
     }
     std::vector<py::object> ret = eval_runtime_tensor_transform(graph, module_outputs, golden_transforms_outputs);
 
+    // Partial data copy outputs.
```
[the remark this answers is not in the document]
dgolubovicTT: Yeah, just wanted to say that output of output node is just what it is on its input. Just to comment what the code is doing. I agree that introducing "fifo queue" only creates confusion. Removed that part

forge/forge/verify/config.py
```diff
@@ -216,13 +214,22 @@ def disabled(cls) -> "DepricatedVerifyConfig":
...
+    # Getter and setter for verify_all
+    def set_verify_all(self, value: bool):
+        self.__verify_all = True
+        self.verify_pipeline_result_vs_framework = True
+        self.verify_forge_codegen_vs_framework = True
+        self.verify_tvm_compile = True
+        self.verify_each_forge_pass = True
+
+    def get_verify_all(self):
+        return self.__verify_all
```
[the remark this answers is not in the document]
dgolubovicTT: I wanted to ensure that when verify_all is set, that all other verification flags are set. I will do this using `@property`. Is much more cleaner, and in line with rest of the configurations.

```
   @property
    def verify_all(self):
        return self._verify_all

    @verify_all.setter
    def verify_all(self, value: bool):
        if value:  # Only update if setting True
            self._verify_all = True
            self.verify_pipeline_result_vs_framework = True
            self.verify_forge_codegen_vs_framework = True
            self.verify_tvm_compile = True
            self.verify_each_forge_pass = True
        else:
            self._verify_all = False
```

so setting verify_all will look like:
`config.verify_all = True` and all other verify flags will be set.

forge/test/mlir/test_golden_verification.py
dgolubovicTT: Good catch! I don't know why I have removed it. The idea is to have few really small tests for golden verification. I added it back...

forge/forge/tvm_calls/forge_compile.py
dgolubovicTT: Here is an issue that sums up this: https://github.com/tenstorrent/tt-forge-fe/issues/1499

PR #1408, opened 2025-03-08
Fix ONNX module importer

sott0n:
### Ticket
N/A

### Problem description
Currently, compiling the ONNX module is not supported yet.

### What's changed
This commit resolves an issue preventing the ONNX module from being
imported through Forge. And it also adds support for loading the ONNX module
directly as bytes, eliminating the need to import an ONNX file at
verification.

### Checklist

### DIFF
--- a/forge/forge/compile.py
+++ b/forge/forge/compile.py
@@ -191,7 +191,7 @@ def compile_main(
     Parameters
     ----------
     module: AnyModule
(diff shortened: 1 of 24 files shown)


nvukobratTT (approved):
Nice! Thanks for fixing up this path @sott0n!

PR #1428, opened 2025-03-11
Requires runs-on for nightly sweeps action

### LINKED ISSUE #859: Filtering of nightly tests by operator
Support filtering of nightly sweep tests by operator

vbrkicTT:
### Ticket
Fixes #859 

### Problem description
Default value for github action runs-on parameter is not respected in scheduled run.
See error in https://github.com/tenstorrent/tt-forge-fe/actions/runs/13780085711

### What's changed
Makes runs-on explicitly required
Introduce test-sweeps.yml workflow to support both schedule and manual run with parameter runs-on 

### Checklist

### DIFF
--- a/.github/workflows/on-nightly-sweeps.yml
+++ b/.github/workflows/on-nightly-sweeps.yml
@@ -1,21 +1,25 @@
-name: On nightly_sweeps
+name: On Nightly Sweeps Tests
 
 on:
   workflow_dispatch:
     inputs:
       runs-on:
         description: 'Runs on'
-        required: false
+        required: true
         type: choice
         options:
           - n150
           - n300
-        default: 'n150'
+        default: n150
       operators:
         description: 'Operators to test (comma separated)'
+        type: string
         required: false
   schedule:
     - cron: '0 4 * * *'  # Runs at 04:00 UTC every day
+
+env:
+  runs-on-default: n150
 
 jobs:
   docker-build:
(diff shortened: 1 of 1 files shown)


.github/workflows/on-nightly-sweeps.yml
```diff
@@ -1,42 +1,27 @@
...
-    uses: ./.github/workflows/build.yml
-    secrets: inherit
-    with:
-      docker-image: ${{ needs.docker-build.outputs.docker-image }}
   test:
-    needs:
-      - docker-build
-      - build
-    uses: ./.github/workflows/test.yml
+    uses: ./.github/workflows/test-sweeps.yml
```
vbrkicTT, line 23: @vmilosevic To support both schedule and manual run with manual selection of parameter runs-on for manual run a new workflow has been introduced `test-sweeps.yml`. I didn't find an elegant solution to set default value for runs-on without the intermediate workflow.

vmilosevic: Yes, unfortunately having default "inputs" in other trigger types is not well supported.
One way you could probably do it is by using other variables like

env:
   default_runs_on: n150
...
runs-on: '[{"runs-on": "${{ inputs.runs-on || env.default_runs_on }}"}]'

vmilosevic (approved):

.github/workflows/on-nightly-sweeps.yml
vbrkicTT: It didn't help. I suspect I made a mistake in naming env variable. Followup PR is #1447

PR #1429, opened 2025-03-11
dtypes verification fix

vkovinicTT:
### Problem description
When comparing outputs between PyTorch models and their Forge-compiled counterparts, dtype mismatches were causing false verification failures. This occurred because Forge tensors support only a subset of PyTorch's data types (e.g., mapping torch.int64 to torch.int32), but our verification logic was performing direct dtype equality checks without accounting for these conversions.

### What's changed
Implemented a `verify_dtypes` function that properly handles the dtype conversion between PyTorch and Forge formats. The function:
1. Converts the PyTorch dtype to its Forge representation
2. Converts the Forge representation back to PyTorch
3. Compares this "round-trip" converted dtype with the compiled model's dtype

This allows verification to pass correctly even when dtypes differ due to Forge's internal mappings, ensuring we only fail on genuine dtype incompatibilities rather than expected conversions.

### DIFF
--- a/forge/forge/verify/verify.py
+++ b/forge/forge/verify/verify.py
@@ -17,7 +17,14 @@ import torch
 import tensorflow as tf
 from forge.tensor import to_pt_tensors
 
-from ..tensor import Tensor, TensorShape, pad_pytorch_tensor_to_forge, narrow_forge_tensor_to_pytorch
+from ..tensor import (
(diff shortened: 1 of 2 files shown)


pilkicTT (approved):

dgolubovicTT (approved):
Legit!

Issue #803, opened 2024-11-29
'ttir.typecast' op Result shape must match operand shapes after broadcasting


ashokkumarkannan1:
### Description:
For the bringup of the distilbert model, faced issue with ttir.typecast

### Reproduce

```bash
git checkout akannan/dataformat_Bfp2_b_fix
pytest forge/test/model_demos/high_prio/nlp/pytorch/test_distilbert.py
```

### Log:

`loc("cast_2407"("forward":4294967295:8970)): error: 'ttir.typecast' op Result shape must match operand shapes after broadcasting`

[test_distilbert.log](https://github.com/user-attachments/files/17959451/test_distilbert.log)

skhatriTT:
@ashokkumarkannan1 what is our status on this one?

PR #1437, opened 2025-03-12
Fix: 'ttir.typecast' op Result shape must match operand shapes after broadcasting 

ashokkumarkannan1:
### Ticket
#803

### What's changed
Created pass to identify the tms in cast node edge and convert it to repeat op
Updated the test file of the distilbert model


dgolubovicTT (changes requested):
I have a few questions about this cast op. Do you know where it comes from? Usually, problems with broadcast being on input edge and causing this shape discrepancy are related to not commuting broadcast tms through node that is added on the edge.

I am curious if that is the case here. If cast op is added on the edge which has broadcast tms (and tms stay on input of cast op) this happens. Can you provide a small repro for this, even better a unit test.

ashokkumarkannan1:
> I have a few questions about this cast op. Do you know where it comes from? Usually, problems with broadcast being on input edge and causing this shape discrepancy are related to not commuting broadcast tms through node that is added on the edge.
> 
> I am curious if that is the case here. If cast op is added on the edge which has broadcast tms (and tms stay on input of cast op) this happens. Can you provide a small repro for this, even better a unit test.

Repro Steps:
```python
import pytest
import torch
import forge
from torch.nn import functional as F
from torch.nn import Module
from forge.verify.verify import verify

class MaskedAttention(Module):
    def __init__(self):
        super(MaskedAttention, self).__init__()

    def forward(self, scores, mask):
        mask_reshp = mask.unsqueeze(1)
        mask = (mask == 0).view(mask_reshp.shape).expand_as(scores)
        scores = scores.masked_fill(mask, torch.tensor(torch.finfo(scores.dtype).min, device=scores.device))
        return scores

@pytest.mark.parametrize(
    "batch_size, n_heads, q_length, k_length, mask_shape",
    [
        (2, 4, 5, 6, (2, 5, 6)),
        (1, 8, 10, 12, (1, 10, 12)),
        (3, 4, 7, 8, (3, 7, 8)),
        (4, 2, 6, 10, (4, 6, 10)),
        (2, 6, 8, 10, (2, 8, 10)),
    ],
)
def test_masked_attention(batch_size, n_heads, q_length, k_length, mask_shape):
    scores = torch.rand(batch_size, n_heads, q_length, k_length)
    mask = torch.randint(0, 2, mask_shape)

    framework_model = MaskedAttention()
    inputs = [scores, mask]

    compiled_model = forge.compile(framework_model, inputs)

    verify(inputs, framework_model, compiled_model)

```

ashokkumarkannan1:
Hi @dgolubovicTT ,
Did you get chance to look into my previous comment. https://github.com/tenstorrent/tt-forge-fe/pull/1437#issuecomment-2723830637

cc: @nvukobratTT

forge/csrc/forge_passes.cpp
```diff
@@ -212,6 +212,12 @@ graphlib::Graph *run_pre_lowering_passes(graphlib::Graph *graph, const std::opti
     // Note this is not lowering, these are still forge tms
     convert_broadcast_ops_to_tms(graph);
 
+    // Consider a Scenario where cast node with the broadcast TMS in the producer edge is not supported.
```
nvukobratTT, line 215: Can we clarify on this one? 

More precisely, is the issues that TTNN doesn't support broadcasts on producer nodes? If that si the case, this can be a better candidate for an MLIR workaround. Correct me if I'm wrong.

nvukobratTT: If that is the case, let's create as simple sanity test for cast, and let's create similar repro on TTNN, and report back both issues to be resolved on MLIR (and potentially TTNN path if applicable). 

Let me know your thoughts

ashokkumarkannan1: Hi @nvukobratTT as the typecast  is unary op broadcast tms in its edge will result in the issue during  mlir verification.

nvukobratTT:
Is this one still active PR?

dgolubovicTT:
I've read the issue and the PR again. Essentially we have a problem with broadcast before any unary op, right @ashokkumarkannan1. It is only triggered here first. 

Since broadcast op is moved to edge (remain of pybuda), turns out that cast op has different input/output shapes.

This moving of broadcast to edge is tech debt that causes this issue. It is not trigered for binary ops because ttnn is doing implicit broadcast for those ops. However, for unary ops, there is no implicit broadcast.

In my opinion, we should have compiler remove broadcast ops where implicit broadcast is done by ttnn, and leave explicit broadcast (for example repeat) where ttnn won't do implicit broadcast.
 
To be clear, I think that leaving broadcasts on edges and then replacing it with the repeat for this case is only a patch. 
In this PR, broadcast tm is replaced by repeat op in pre lowering pass. It is hard coded only for cast op:
`std::vector<std::string> oplist = {"cast"};`
My suggestion is to do it earlier (not sure where, but we can discuss/see) and solve it generally.

dgolubovicTT:
@sdjordjevicTT Told me we have ttir.broadcast that calls ttnn.repeat under. So we should be fine with mapping forge broadcast to ttir.broadcast...

ashokkumarkannan1:
> @sdjordjevicTT Told me we have ttir.broadcast that calls ttnn.repeat under. So we should be fine with mapping forge broadcast to ttir.broadcast...
@nvukobratTT  shall I proceed with approach that is suggested by @dgolubovicTT ?

nvukobratTT:
> > @sdjordjevicTT Told me we have ttir.broadcast that calls ttnn.repeat under. So we should be fine with mapping forge broadcast to ttir.broadcast...
> > @nvukobratTT  shall I proceed with approach that is suggested by @dgolubovicTT ?

@ashokkumarkannan1 I think we can close this PR. @dgolubovicTT already has a potential solution in separate effort, so let's focus on other problems in order to escape duplicated work.

nvukobratTT:
I'll move this PR in Draft, if everyone agrees please close it :))

ashokkumarkannan1:
Sure @nvukobratTT . As @dgolubovicTT  working on a fix for this issue, I will close this PR.

PR #1438, opened 2025-03-12
Add ResNet test for ONNX

sott0n:
Forge can compile ONNX models end-to-end, and this PR adds a test for ResNet50 with minor changes.

In this test, the ONNX file is generated from a PyTorch module using the `onnx.export` API and stored in a temporary path. However, this export process takes approximately 7–8 seconds, which could become a bottleneck if we increase the number of test cases, potentially impacting CI runtime.  

An alternative approach is to store a pre-generated ONNX file in this repository using Git LFS or retrieve it from an external source. Some inactive tests reference an ONNX file from `confidential_customer_models`, but this third-party repository appears to be internal to TT and is not publicly accessible.  

If you have any suggestions regarding handling the ONNX file efficiently, please share your thoughts in the comments.

### DIFF
--- a/forge/forge/module.py
+++ b/forge/forge/module.py
@@ -26,6 +26,7 @@ from .tensor import (
 )
 from .parameter import Parameter
 import onnx
+from onnx import numpy_helper
 import jax.numpy as jnp
 import numpy as np
 
@@ -440,13 +441,7 @@ class OnnxModule(Module):
     def get_parameters(self) -> List[Parameter]:
         params = []
         for param in self.module.graph.initializer:
-            d_type = onnx.mapping.TENSOR_TYPE_TO_NP_TYPE[param.data_type]
-            if param.raw_data:
-                param_data = np.frombuffer(param.raw_data, dtype=d_type).reshape(param.dims)
-            elif param.float_data:
-                param_data = param.float_data
-            else:
-                raise ValueError(f"Parameter: {param.name} don't have raw_data or float_data")
+            param_data = numpy_helper.to_array(param)
             forge_param = Parameter(torch.tensor(param_data), requires_grad=False, name=param.name)
             params.append(forge_param)
         return params
--- a/forge/forge/module.py
+++ b/forge/forge/module.py
@@ -441,7 +441,12 @@ class OnnxModule(Module):
         params = []
         for param in self.module.graph.initializer:
             d_type = onnx.mapping.TENSOR_TYPE_TO_NP_TYPE[param.data_type]
-            param_data = np.frombuffer(param.raw_data, dtype=d_type).reshape(param.dims)
+            if param.raw_data:
+                param_data = np.frombuffer(param.raw_data, dtype=d_type).reshape(param.dims)
+            elif param.float_data:
+                param_data = param.float_data
+            else:
+                raise ValueError(f"Parameter: {param.name} don't have raw_data or float_data")
             forge_param = Parameter(torch.tensor(param_data), requires_grad=False, name=param.name)
             params.append(forge_param)
         return params
--- a/forge/test/models/onnx/vision/resnet/test_resnet.py
+++ b/forge/test/models/onnx/vision/resnet/test_resnet.py
@@ -43,9 +43,9 @@ def test_resnet_onnx(variant, tmp_path, record_forge_property, opset_version):
 
     # Export model to ONNX
     torch_model = ResNetForImageClassification.from_pretrained(variant)
-    dummy_input = torch.randn(1, 3, 224, 224)
+    input_sample = torch.randn(1, 3, 224, 224)
     onnx_path = f"{tmp_path}/resnet50.onnx"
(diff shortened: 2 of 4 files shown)


forge/forge/module.py
```diff
@@ -441,7 +441,12 @@ def get_parameters(self) -> List[Parameter]:
         params = []
         for param in self.module.graph.initializer:
             d_type = onnx.mapping.TENSOR_TYPE_TO_NP_TYPE[param.data_type]
-            param_data = np.frombuffer(param.raw_data, dtype=d_type).reshape(param.dims)
+            if param.raw_data:
```
nvukobratTT, line 444: For easier future understanding, let's add comments describing what raw and float data represent

forge/test/models/onnx/vision/resnet/test_resnet.py
```diff
@@ -0,0 +1,64 @@
...
+
+# Opset 7 is the minimum version in Torch, and 17 is the maximum version in Torchscript.
+opset_versions = [7, 17]
+
+
+@pytest.mark.push
+@pytest.mark.nightly
+@pytest.mark.parametrize("variant", variants, ids=variants)
+@pytest.mark.parametrize("opset_version", opset_versions, ids=opset_versions)
+def test_resnet_onnx(variant, tmp_path, record_forge_property, opset_version):
```
nvukobratTT, line 31: Do you need tmp_path here?

forge/test/models/onnx/vision/resnet/test_resnet.py
```diff
@@ -0,0 +1,64 @@
...
+        model="resnet",
+        variant="50",
+        source=Source.HUGGINGFACE,
+        task=Task.IMAGE_CLASSIFICATION,
+    )
+    record_forge_property("tags.model_name", module_name)
+
+    # Export model to ONNX
+    torch_model = ResNetForImageClassification.from_pretrained(variant)
+    dummy_input = torch.randn(1, 3, 224, 224)
```
nvukobratTT, line 46: Let's call it sample input for consistency

forge/test/models/onnx/vision/resnet/test_resnet.py
```diff
@@ -0,0 +1,64 @@
...
+    onnx_path = f"{tmp_path}/resnet50.onnx"
```
nvukobratTT, line 47: Seems like tmp path isn't defined. We can just store them in root dir, and delete after test run

nvukobratTT:
General commit regarding storing ONNX files offline. I agree. Request for getting this storage for onnx files, and similar data is in progress. 

For now, it's okay to lose ~10ish second on each model until that solution is in place. 

And thanks once again for pushing these models e2e! 🤘❤️

forge/test/models/onnx/vision/resnet/test_resnet.py
sott0n: In pytest, tmp_path is a built-in fixture that automatically creates a temporary directory for each test. The directory follows the structure {temproot}/pytest-of-{user}/pytest-{num}/{testname}/ and is automatically removed after the test runs. Since tmp_path simplifies temporary file management, is it better than manually creating and deleting files ourselves?

Ref tmp_path fixture: https://docs.pytest.org/en/stable/how-to/tmp_path.html

forge/test/models/onnx/vision/resnet/test_resnet.py
sott0n: same question: https://github.com/tenstorrent/tt-forge-fe/pull/1438#discussion_r1992450997

forge/test/models/onnx/vision/resnet/test_resnet.py
nvukobratTT: I see. Thanks for the context! :))

forge/test/models/onnx/vision/resnet/test_resnet.py
```diff
@@ -0,0 +1,64 @@
...
+
+variants = [
+    "microsoft/resnet-50",
+]
+
+# Opset 7 is the minimum version in Torch, and 17 is the maximum version in Torchscript.
+opset_versions = [7, 17]
+
+
+@pytest.mark.push
```
pilkicTT, line 27: I think running this only in nightly should be enough?

nvukobratTT: I think it'll be good to have at least a single model example for ONNX running e2e. Rest should be only nightly :))

forge/forge/module.py
sott0n: The ONNX numpy_helper API provides a convenient way to convert TensorProto to NumPy arrays. I updated the code to use this API, as it supports both raw data types and ONNX's element types.
Sorry, my initial commit did not cover ONNX element types widely enough. I believe this can be resolved by using this API.

- https://onnx.ai/onnx/api/numpy_helper.html#onnx.numpy_helper.to_array

dgolubovicTT (approved):

nvukobratTT (approved):

PR #1440, opened 2025-03-12
Fix Huggingface Token in Perf Benchmark

vcanicTT:
We didn't have access to huggingface llama-3.2b model because of lack of the token. We add the token and fix model verbose for resnet and llama.

### DIFF
--- a/.github/workflows/on-nightly.yml
+++ b/.github/workflows/on-nightly.yml
@@ -1,6 +1,7 @@
 name: On nightly
 
 on:
+  push:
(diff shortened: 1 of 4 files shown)


pilkicTT (approved):

vcanicTT:
@dgolubovicTT @vmilosevic @pilkicTT Can you take a look? It would be nice if we can push this change today, so it can affect the next nightly run.

vmilosevic (approved):

Issue #1398, opened 2025-03-07
Record bringup_status property and update execution_stage recording without binding


chandrasekaranpradeep:
1. Remove **ExecutionPhase** enum present in the forge/forge/execution_tracker.py file
2. For recording test compilation and execution process, create new enum named **ExecutionDepth** and have below members.
    1. `FAILED_FE_COMPILATION` - Front end compilation fails (shlo->ttir or tvm->ttir)
    2. `FAILED_TTMLIR_COMPILATION` - TT-MLIR compilation fails, can't produce flatbuffer
    3. `FAILED_RUNTIME` - Runtime execution fails
    4. `INCORRECT_RESULT` - Flatbuffer executed, output incorrect
    5. `PASSED` - Outputs correct (pcc >= 0.99)
3. Record the execution_depth enum inside the `bringup_status` property which will be present inside the "tags" json property
Eg: `record_forge_property("tags.bringup_status", "FAILED_FE_COMPILATION")`
4. **ExecutionStage** can be still recorded which will give more granular compilations status and it will be specific to forge-fe  and  remove the binding with **ExecutionPhase**/**ExecutionDepth** in **ExecutionStage** data members
5. Modify the data members of the **ExecutionStage** to align with **ExecutionDepth**(i.e Record failing execution stage instead of passing ones)

nvukobratTT:
@chandrasekaranpradeep one more details, to be more aligned with execution depth, can we rename execution stage data members to follow similar philosophy? E.g. to record where they failed, instead of where they passed. 

My main reasoning is because model can pass multiple stages, but only fail on single. Therefore, recording failure stages is potentially more correct.

chandrasekaranpradeep:
Got it, will also update the execution stage data members  to align with execution depth(i.e recording failing stage)

chandrasekaranpradeep:
Closing the issue because the issue was addressed here - https://github.com/tenstorrent/tt-forge-fe/pull/1442

Issue #1361, opened 2025-03-04
Execution tracker unnecessary warnings


pilkicTT:
Execution tracker prints out unneeded warnings which can only confuse the user. From my understanding this tool should be used only as part of the test infrastructure, so in that sense it doesn't make sense to be active when we are not running tests.

If we want to expose it to the user then clear APIs need to be provided.

Example:

```
2025-03-04 10:40:34.563 | INFO     | forge.compile:compile_main:227 - Compiling module Add
2025-03-04 10:40:34.564 | INFO     | forge.compile:forge_compile_from_context:286 - Running compile stage init_compile
2025-03-04 10:40:34.564 | INFO     | forge.compile:forge_compile_from_context:286 - Running compile stage generate_initial_graph
2025-03-04 10:40:34.564 | WARNING  | forge.execution_tracker:record_execution_phase_and_stage:177 - Unable to find current test
2025-03-04 10:40:36.071 | WARNING  | forge.execution_tracker:record_execution_phase_and_stage:177 - Unable to find current test
2025-03-04 10:40:36.072 | WARNING  | forge.execution_tracker:record_execution_phase_and_stage:177 - Unable to find current test
2025-03-04 10:40:36.093 | WARNING  | forge.execution_tracker:record_execution_phase_and_stage:177 - Unable to find current test
2025-03-04 10:40:36.116 | WARNING  | forge.execution_tracker:record_execution_phase_and_stage:177 - Unable to find current test
2025-03-04 10:40:36.145 | WARNING  | forge.execution_tracker:record_execution_phase_and_stage:177 - Unable to find current test
2025-03-04 10:40:36.146 | DEBUG    | forge.tvm_to_python:_determine_node_dtype:2149 - Node 'x1' does not have a framework dtype specified. Using TVM generated dtype.
2025-03-04 10:40:36.146 | DEBUG    | forge.tvm_to_python:_determine_node_dtype:2149 - Node 'x2' does not have a framework dtype specified. Using TVM generated dtype.
2025-03-04 10:40:36.148 | DEBUG    | forge.verify.compare:compare_tensor_to_golden:289 - Tensors match on Framework vs. Forge codegen output 0
2025-03-04 10:40:36.148 | INFO     | forge.tvm_to_python:verify_framework_vs_forge_codegen:1955 - Verified python codegen agains framework
[...]

Issue #1333, opened 2025-02-26
Serialize and record config


nvukobratTT:
## Summary
For easier reproducibility, it'll be good to serialize compiler and verify configurations as part of the "config" key in the pytest property. 

These two can be duped in JSON, and stored as two sub-keys:
- config.compile
- config.verify

chandrasekaranpradeep:
The compiler and verify config was recoded in this [PR](https://github.com/tenstorrent/tt-forge-fe/pull/1442)  and merged into main. so closing the issue

PR #1442, opened 2025-03-12
Refactor forge properties recording and record execution depth and stage and compiler and verify config

chandrasekaranpradeep:
1. Reimplement forge properties recording logic  - https://github.com/tenstorrent/tt-forge-fe/issues/1452

In current Implementation, for recording forge properties like ExecutionPhase(i.e ExecutionDepth) and ExecutionStage, will collect the [current running test command before every test execution and save it as environmental flags](https://github.com/tenstorrent/tt-forge-fe/blob/7cb612649389dd0e26a4ceca3e9e74d9abca2a9a/forge/test/conftest.py#L77)  and will record ExecutionDepth and ExecutionStage and store in [json file with key as current running test command as key and properties as values](https://github.com/tenstorrent/tt-forge-fe/blob/7cb612649389dd0e26a4ceca3e9e74d9abca2a9a/forge/forge/execution_tracker.py#L127)and after the test execution completed will [record those properties inside the pytest record_property fixtures by fetching from json file](https://github.com/tenstorrent/tt-forge-fe/blob/7cb612649389dd0e26a4ceca3e9e74d9abca2a9a/forge/test/conftest.py#L192).

But recording properties through env flags produces [unnecessary warnings](https://github.com/tenstorrent/tt-forge-fe/issues/1361) and saving those properties inside the json through env flags(i.e current running test command) is not good way of doing

Created the `ForgePropertyHandler`(i.e utility class) to manage and record properties inside the test function and compiler. The ForgePropertyHandler can be instantiated by calling the `forge_property_recorder` pytest fixture inside the test function. With the help of the ForgePropertyHandler instance, we can record model_name, op_name, group and by passing it as argument to compile and verify function for recording the execution stage and depth and other properties from TT Binary in future.

2. Record bringup_status property and update execution_stage recording without binding -https://github.com/tenstorrent/tt-forge-fe/issues/1398
    1. Reimplemented the ExecutionPhase(i.e ExecutionDepth) with the new data members and added support for recording the new ExecutionDepth inside the bringup_status property.
eg: "bringup_status" : "FAILED_FE_COMPILATION"
     2. Decoupled ExecutionStage and ExecutionPhase enum and updated the data members of the ExecutionStage to follow similar philosophy.(i.e record where they failed, instead of where they passed) and added support for recording it with ForgePropertyHandler.
     
3. Serialize and record config - https://github.com/tenstorrent/tt-forge-fe/issues/1333
Added support for recording compiler and verify configuration as dict inside the config property
```
{
   'owner' : 'tt-forge-fe',
   'group' : 'priority'
   'tags'   : {.....}
  'config' : {
                  'compiler': {'enable_training': False, 'enable_recompute': False, etc...}
                  'verify': {'enabled': True, 'verify_size': True, 'verify_dtype': True, 'verify_shape': True, 'verify_values': True, 'value_checker': {'rtol': 1e-05, 'atol': 1e-08, 'pcc': 0.99, 'dissimilarity_threshold': 0.001}, 'dump_tensors': False, 'dump_tensors_path': ''},
                  }
 }
```
     
4. Modified all the push, nightly and models_ops test to use forge_property_recorder and depreciated record_forge_property fixture.


forge/csrc/forge_bindings.cpp
```diff
@@ -171,6 +172,35 @@ PYBIND11_MODULE(_C, m)
...
 
+    py::enum_<tt::property::ExecutionDepth>(m, "ExecutionDepth")
+        .value("FAILED_FE_COMPILATION", tt::property::ExecutionDepth::FAILED_FE_COMPILATION)
+        .value("FAILED_TTMLIR_COMPILATION", tt::property::ExecutionDepth::FAILED_TTMLIR_COMPILATION)
+        .value("FAILED_RUNTIME", tt::property::ExecutionDepth::FAILED_RUNTIME)
+        .value("INCORRECT_RESULT", tt::property::ExecutionDepth::INCORRECT_RESULT)
+        .value("PASSED", tt::property::ExecutionDepth::PASSED)
+        .export_values()
+        .def(
+            "to_json",
```
ashokkumarkannan1, line 183: why not `to_str` as we are returning string and not json? Same for `from_json` also?

forge/csrc/forge_bindings.cpp
```diff
@@ -171,6 +172,35 @@ PYBIND11_MODULE(_C, m)
...
+            {
+                std::stringstream ss;
+                ss << execution_depth;
+                return ss.str();
+            })
+        .def(
+            "from_json",
+            [](std::string const &encoded)
+            {
+                static std::unordered_map<std::string, tt::property::ExecutionDepth> decode = {
```
ashokkumarkannan1, line 194: Shall we use switch case here? and why static?

forge/csrc/passes/mlir_compiler.cpp
```diff
@@ -78,6 +79,9 @@ auto run_mlir_compiler_generic(tt::ForgeGraphModule& module)
     // Generate MLIR from the Forge graph.
     mlir::OwningOpRef<mlir::ModuleOp> mlir_module = lower_to_mlir(module, context);
 
+    tt::property::record_execution_depth(
+        forge_property_handler, tt::property::ExecutionDepth::FAILED_TTMLIR_COMPILATION);
```
ashokkumarkannan1, line 83: What if `else if constexpr (output == MLIROutputKind::Cpp)` block is executed after setting execution depth to `FAILED_TTMLIR_COMPILATION`? Because in that block we are not overriding it? Instead I would suggest using below approach.

```cpp
try {
    // Run MLIR pipeline.
     run_mlir_passes<output>(mlir_module);
} catch {
    tt::property::record_execution_depth(
         forge_property_handler, tt::property::ExecutionDepth::FAILED_TTMLIR_COMPILATION);
}

File: forge/forge/compile.py [hunk 4]
```diff
@@ -235,6 +237,11 @@ def compile_main(
     modules = [wrap_module(module, module_name)]
     training = training or optimizer is not None
 
+    if forge_property_handler is not None:
+        forge_property_handler.record_compiler_config(compiler_cfg)
+        forge_property_handler.record_execution_depth(ExecutionDepth.FAILED_FE_COMPILATION)
+        forge_property_handler.record_execution_stage(ExecutionStage.FAILED_TVM_RELAY_IRMODULE_GENERATION)
```
ashokkumarkannan1, line 243: Shall we have one more method `record_execution`?
Example:
```python
forge_property_handler.record_execution(ExecutionDepth.FAILED_FE_COMPILATION, ExecutionStage.FAILED_TVM_RELAY_IRMODULE_GENERATION)
```

ashokkumarkannan1: Just curious, will it be possible to auto detect execution_depth based on execution_stage value?

forge/forge/compile.py
```diff
@@ -972,20 +992,21 @@ def split_graph(context: CompileContext) -> CompileDepth:
...
 
 def run_mlir_compiler(context: CompileContext) -> CompileDepth:
     assert context.forge_module is not None
+    if context.forge_property_handler is not None:
+        context.forge_property_handler.record_execution_stage(ExecutionStage.FAILED_FORGE_MLIR_COMPILATION)
 
-    context.compiled_binary = forge._C.run_mlir_compiler(context.forge_module)
-
-    record_execution_phase_and_stage(ExecutionPhase.COMPILE_MLIR)
+    context.compiled_binary = forge._C.run_mlir_compiler(context.forge_module, context.forge_property_handler)
```
ashokkumarkannan1, line 1009: Shall we have try...except block here as mentioned above? because setting `ExecutionStage.FAILED_FORGE_MLIR_COMPILATION` doesn't make sense before getting any such failure?

forge/forge/forge_property_utils.py
```diff
@@ -0,0 +1,356 @@
...
+    atol: Optional[float] = None
+    execution_stage: str = ""
+    op_name: str = ""
+
+
+@dataclass_json
+@dataclass
+class ForgePropertyStore:
+    owner: str = "tt-forge-fe"
+    group: str = ""
```
ashokkumarkannan1, line 66: Let default value be `generality`

forge/csrc/forge_bindings.cpp
chandrasekaranpradeep: Other enums like `DataFormat`, `MathFidelity` are using `from_json` and `to_json` function names for converting to string or creating respective object so I am aligning with already existing enums.

@nvukobratTT Whats your thoughts on this function name updation suggestion, do we need to change method names from `to_json` to `to_str` and `from_json` to `from_str`?

forge/forge/compile.py
chandrasekaranpradeep: We can create method named record_execution for recording execution depth and execution stage like you mentioned.
The issue https://github.com/tenstorrent/tt-forge-fe/issues/1398 address to decouple execution stage and depth so we won't detect execution_depth based upon the execution stage

forge/csrc/passes/mlir_compiler.cpp
chandrasekaranpradeep: The If block(i.e` if constexpr (output == MLIROutputKind::Flatbuffer)`) will be executed for generating flatbuffer because we are only calling the `run_mlir_compiler` from the `forge/forge/compile.py` file and not calling the `run_mlir_compiler_to_cpp` which is used for executing the else if block(i.e `else if constexpr (output == MLIROutputKind::Cpp)`) and in that cases we don't need try catch block here.

We are updating the execution depth from FAILED_FE_COMPILATION to FAILED_TTMLIR_COMPILATION after generating TTIR(i.e after lower_to_mlir function).
We are updating the execution depth from FAILED_TTMLIR_COMPILATION to FAILED_RUNTIME after generating flatbuffer(i.e after run_mlir_passes and ttnnToFlatbuffer function).

forge/forge/forge_property_utils.py
chandrasekaranpradeep: We should set the group property for models present inside the forge/test/models path but this feature will be also used by other tests(i.e ops tests(i.e models_ops), forge/test/mlir). So in that cases, we should not set group with generality as default value

forge/csrc/passes/mlir_compiler.cpp
ashokkumarkannan1: @nvukobratTT  May I know your thoughts in this? Will it be fine to set `FAILED_TTMLIR_COMPILATION` even before we start running the `run_mlir_passes` ?

chandrasekaranpradeep: @ashokkumarkannan1  Can you check this one, it is related to above one - https://github.com/tenstorrent/tt-forge-fe/pull/1442#discussion_r1995182805?

forge/forge/compile.py
chandrasekaranpradeep: Created record_execution method

forge/forge/compile.py
ashokkumarkannan1: cc: @nvukobratTT  

Can you please share your thoughts in this?

forge/csrc/forge_bindings.cpp
nvukobratTT: I think @ashokkumarkannan1 has a point, we're not doing JSON conversion here, but pure string. We should name it that way to avoid confusions

forge/csrc/passes/mlir_compiler.cpp
nvukobratTT: The suggestion is to wrap in try/catch block is cleaner. However, I think the main issue is that our compiler passes aren't cleanelly structured. Because of it, it's not easy to define best place to put these properties. 

For now, I'm okay with keeping it this way as @chandrasekaranpradeep suggested. Once we start a push to restructure compiler passes, we'll also update where these properties are recorded.

forge/forge/compile.py
nvukobratTT: Similar to comment above, it's cleaner to have try/catche fro my side. However, I'm not sure if we have a  clean handling of errors with exceptions. If that is the case, e.g. we're raising expectations for all cases, then we can use try/catch blocks.

If not, it's safer to use this approach until we restructure the compile stages

nvukobratTT (approved):

forge/csrc/forge_bindings.cpp
chandrasekaranpradeep: I will update methods names from `to_json` to `to_str` and `from_json` to `from_str` for `ExecutionDepth` and `ExecutionStage` enum.
For updating it in other enum class like `DataFormat`, `MathFidelity`, we will create issue for addressing it in separate PR

chandrasekaranpradeep: Updated the `ExecutionDepth` and `ExecutionStage` enum methods

ashokkumarkannan1 (commented):
@chandrasekaranpradeep Let's proceed as @nvukobratTT  mentioned for now. Later while doing compiler cleanups and restructuring, We can do this in an expected way.

ashokkumarkannan1 (approved):

PR #1456, opened 2025-03-14
[CI] Nightly xfail workflow

vkovinicTT:
## Add workflow for testing xfailing nightly models
This PR adds a new workflow that runs after the nightly workflow completes, focusing specifically on tests marked with both `nightly` and `xfail`. This separation improves our CI process by:

1. Making the main nightly workflow faster by focusing only on tests expected to pass
2. Creating a dedicated workflow for tracking expected failures

### Nightly Workflow
- Modified the existing nightly workflow to save the Docker image name as an artifact

### New "On nightly xfail" Workflow

- Created a new workflow that runs automatically after the nightly workflow completes
- Retrieves the Docker image from the previous workflow to avoid rebuilding (or falls back to building a new Docker image if needed)
- Only runs tests marked with both `nightly` and `xfail`

This approach efficiently reuses resources from the main nightly run, making the process of running xfailing tests more efficient.


.github/workflows/test.yml
```diff
@@ -101,10 +105,19 @@ jobs:
...
+      with:
+        name: forge-wheel
+
+    - name: Download from specified workflow run
+      if: ${{ inputs.run-id }}
       uses: actions/download-artifact@v4
       with:
         name: forge-wheel
+        run-id: ${{ inputs.run-id }}
+        github-token: ${{ github.token }}
```
pilkicTT, line 120: Isn't `inputs.run-id` always set? In `on-nightly-xfail.yml`, line 43.

So no need for two steps, just have one which passes `inputs.run-id` into `download-artifacts`.

vkovinicTT: You ara right, we can just move line:
```
run-id: ${{ github.event.workflow_run.id || github.event.inputs.run_id || github.run_id }}
```
from `on-nightly-xfail.yml` to `test.yml`.

.github/workflows/on-nightly-xfail.yml
```diff
@@ -0,0 +1,43 @@
...
+    if: ${{ !github.event.inputs.run_id && !github.event.workflow_run.id }}
+    uses: ./.github/workflows/build.yml
+    secrets: inherit
+    with:
+      docker-image: ${{ needs.docker-build.outputs.docker-image }}
+  test_full_model_xpassing:
+    needs:
+      - docker-build
+      - build
+    if: ${{ needs.docker-build.result == 'success' && (needs.build.result == 'success' || needs.build.result == 'skipped') }}
```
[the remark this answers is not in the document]
vkovinicTT: But what happens if `build` is skipped, will `test_full_model_xpassing` still be executed?

.github/workflows/on-nightly-xfail.yml
```diff
@@ -0,0 +1,43 @@
+name: On nightly xfail
+
```
vmilosevic, line 2: Please add a comment describing the workflow purpose, maybe something like this

```
# This workflow runs after the "on-nightly" workflow.  
# It executes tests that are expected to fail (xFail).  
# Note: This workflow does NOT build the code.  
# Instead, it uses the WHL installer generated from the nightly run.  
```

PR #1457, opened 2025-03-14
[bench] make benchmarks repeatable

pilkicTT:
Resetting seeds at each benchmark run, to make them repeatable and reproducable.


vcanicTT (approved):

forge/test/conftest.py
```diff
@@ -203,33 +207,17 @@ def record_forge_property(record_property):
...
-
-    import tensorflow as tf
-
-    tf.random.set_seed(0)
-
-    yield
-
-    # clean up after each test
-    forge.forge_reset()
-    torch._dynamo.reset()
```
vkovinicTT, line 230: Can we also remove these functions?

vkovinicTT (changes requested):

forge/test/conftest.py
pilkicTT: removed `forge_reset()`; the other one is a torch function

vkovinicTT (approved):

PR #1459, opened 2025-03-14
Add support for recording pcc and atol

### LINKED ISSUE #1394: Record pcc and atol
Record below properties inside the "tags" forge property:

 1. ` pcc:` float - PCC; correlation accuracy
 2. `atol:` float - Absolute tolerance; numerical stability

### Note:
1. The pcc and atol should be calculated independent of verification enabled/disabled(i.e `verify_cfg.enabled = True/False`).
2. Create a utility function to calculate pcc and atol and this function should be called inside the verify function present in the `forge/forge/verify/verify.py` file after getting compiled and framework model output.

chandrasekaranpradeep:
Fixes #1394 

1. This PR introduces the `compute_pcc_and_atol` function for calculating **PCC** and **ATOL** values between framework and compiled model outputs. The calculated pcc and atol values will be recorded inside the **tags** property. The function calculates the Pearson correlation coefficient (PCC) for non-scalar outputs and the absolute tolerance (ATOL) for all outputs. For scalar outputs, only ATOL is computed.
In tt-torch, ATOL calculation handles special cases like Inf and NaN values, so replicated the ATOL calculation here.
Reference: https://github.com/tenstorrent/tt-torch/blob/c79248aa29f736e7977e22555d45b47fb2b9d12f/tt_torch/tools/utils.py#L579

2. Modified the forge loss test function present inside the forge/test/mlir/test_loss.py path to use verify function

Note:

1. If the number of framework outputs does not match the number of compiled outputs, the pcc and atol will not be recorded.
2. For non-scalar outputs, if the tensor shapes do not match, the PCC and ATOL calculations for that particular output are skipped.
3. The minimum PCC is computed among  non-scalar outputs, and the maximum ATOL is computed across all outputs.

### DIFF
--- a/forge/forge/verify/compare.py
+++ b/forge/forge/verify/compare.py
@@ -11,6 +11,7 @@ import tensorflow as tf
 import numpy as np
 from loguru import logger
 from scipy.spatial import distance
+from typing import Union, Tuple, List, Optional
 
 from forge.tensor import narrow_forge_tensor_to_pytorch
 
@@ -21,12 +22,14 @@ def compare_with_golden(
     pcc: float = 0.99,
     rtol: float = 1e-05,
     atol: float = 1e-08,
-    dissimilarity_threshold: float = 1e-03,
+    dissimilarity_threshold: float = 1e-03,  # threshold picked empirically. We will update it as TTNN evolves
 ):
     if golden.dtype == torch.bool:
-        result = compare_with_golden_bool(golden, calculated, dissimilarity_threshold)
+        calculated_dissimilarity = calculate_dissimilarity(golden, calculated)
+        result = compare_dissimilarity(calculated_dissimilarity, dissimilarity_threshold)
     if golden.flatten().size() != (1,):  # PCC for single values doesn't work
-        result = compare_with_golden_pcc(golden, calculated, pcc)
+        calculated_pcc = calculate_pcc(golden, calculated)
+        result = compare_pcc(calculated_pcc, pcc)
     else:
         # For scalar values, we can't calculate PCC, but we can compare golden and calculated values using relative and absolute tolerances
         golden = golden.flatten()[0]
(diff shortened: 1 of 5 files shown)


forge/forge/verify/compare.py
```diff
@@ -299,3 +300,109 @@ def compute_required_tolerances(a, b):
...
+    calculated[calculated_nan_mask] = 0
+
+    # Replace Inf values with 0 to avoid having NaN as ATOL
+    golden[golden_inf_mask] = 0
+    calculated[calculated_inf_mask] = 0
+
+    return torch.max(torch.abs(golden - calculated)).item()
+
+
+def compute_pcc_and_atol(
```
dgolubovicTT, line 358: Do we have to write this from scratch? Didn't we have something similar already? fyi @vkovinicTT

chandrasekaranpradeep: We are already having function for calculating pcc(i.e [calculate_pcc](https://github.com/tenstorrent/tt-forge-fe/blob/41b429c316a342f283176cc3e5f57652ad838930/forge/forge/verify/compare.py#L100C5-L100C18)) and atol (i.e [compute_required_tolerances](https://github.com/tenstorrent/tt-forge-fe/blob/41b429c316a342f283176cc3e5f57652ad838930/forge/forge/verify/compare.py#L286C5-L286C32)) but the atol  function(i.e [compute_required_tolerances](https://github.com/tenstorrent/tt-forge-fe/blob/41b429c316a342f283176cc3e5f57652ad838930/forge/forge/verify/compare.py#L286C5-L286C32) ) doesn't handle inf and nan values. There is no separate function for calculating the pcc and atol between framework and compiled model outputs. so I have created `compute_pcc_and_atol` function which utilize existing [calculate_pcc](https://github.com/tenstorrent/tt-forge-fe/blob/41b429c316a342f283176cc3e5f57652ad838930/forge/forge/verify/compare.py#L100C5-L100C18) for calculating pcc and for atol calculation I have taken calculate_atol function from tt-torch because it handles inf and nan values.

forge/forge/verify/verify.py
```diff
@@ -360,6 +371,13 @@ def verify(
 
     co_out = [co.to("cpu") for co in co_out]
 
+    if forge_property_handler is not None:
+        pcc, atol = compute_pcc_and_atol(fw_out, co_out)
+        if pcc is not None:
+            forge_property_handler.record_pcc(pcc=pcc)
+        if atol is not None:
+            forge_property_handler.record_atol(atol=atol)
```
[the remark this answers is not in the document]
chandrasekaranpradeep: @nvukobratTT told to record the pcc and atol values independent of verification is enabled or disabled.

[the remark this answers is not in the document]
nvukobratTT: Yes, even with verification disabled, we want to record these details into our DB for more granular overview in Superset.

forge/forge/verify/compare.py
chandrasekaranpradeep: In that cases, can we remove below tensor cloning and inplace operation and use torch.where?
Current Implementation:
```
     a = a.clone()
    a[torch.logical_or(torch.isnan(a), torch.logical_or(torch.isinf(a), torch.isneginf(a)))] = 0
    b = b.clone()
    b[torch.logical_or(torch.isnan(b), torch.logical_or(torch.isinf(b), torch.isneginf(b)))] = 0
```
Suggestion:
```
    invalid_a = torch.isnan(a) | torch.isinf(a) | torch.isneginf(a)
    invalid_b = torch.isnan(b) | torch.isinf(b) | torch.isneginf(b)
    a_fixed = torch.where(invalid_a, torch.zeros_like(a), a)
    b_fixed = torch.where(invalid_b, torch.zeros_like(b), b)
```

nvukobratTT: Can we try to unify these? 

E.g. we should have separated functions for:
- Calculating PCC
- Comparing PCC
- Calculating Atol
- Comparing Atol

For the `compute_required_tolerance`, we should use an updated version (or a combination of them if applicable) as these calculations should be aligned across FFE for all cases, both comparison and recording. 

Also, for this function `compute_pcc_and_atol ` we should have a better name to outline the main purpose.

Finally, for the `calculate_pcc` rewrite, this should be a separate effort and addressed in different PR.

vkovinicTT: > In that cases, can we remove below tensor cloning and inplace operation and use torch.where? Current Implementation:
> 
> ```
>      a = a.clone()
>     a[torch.logical_or(torch.isnan(a), torch.logical_or(torch.isinf(a), torch.isneginf(a)))] = 0
>     b = b.clone()
>     b[torch.logical_or(torch.isnan(b), torch.logical_or(torch.isinf(b), torch.isneginf(b)))] = 0
> ```
> 
> Suggestion:
> 
> ```
>     invalid_a = torch.isnan(a) | torch.isinf(a) | torch.isneginf(a)
>     invalid_b = torch.isnan(b) | torch.isinf(b) | torch.isneginf(b)
>     a_fixed = torch.where(invalid_a, torch.zeros_like(a), a)
>     b_fixed = torch.where(invalid_b, torch.zeros_like(b), b)
> ```

We already have someone working on it, so no need for changing it yourself.

chandrasekaranpradeep: Okay Thanks for letting me know

chandrasekaranpradeep: @nvukobratTT  Created a separate function for calculating pcc, atol and dissimilarity and also separate function for comparing pcc and dissimilarity.

Modified the `compare_with_golden` function to use pcc and dissimilarity calculation and comparision function and removed the `compare_with_golden_pcc` and `compare_with_golden_bool` function.

Updated the `compute_required_tolerances` function to use new atol calculation function(i.e `calculate_atol` from tt-torch).

Renamed `compute_pcc_and_atol` to `determine_consistency_limits`

In `determine_consistency_limits`, instead of recording the mean of all PCC and ATOL values, we are now extracting the minimum PCC and maximum ATOL from the calculated lists, following the[ tt-torch approach that records these extreme values](https://github.com/tenstorrent/tt-torch/blob/0d63245f020eec153a6336ec00bea8acfb48b1dd/tests/utils.py#L356).

nvukobratTT (approved):
Approving this change for now. However, we'll have to optimize our memory consumption during verification (WIP through separate effort). 

@chandrasekaranpradeep if you see that after this change we're hitting more OOM regressions, make sure to disable it by default until OOM issues are addressed.

saiarthiraguram (approved):

PR #1462, opened 2025-03-17
Revert "Streamline json file generation"

vcanicTT:
This commit is unnecessary, we should revert it.


nsumrakTT (approved):
Cool, thanks and sorry for the inconvenience...

Issue #1291, opened 2025-02-21
Surya OCR


nvukobratTT:

skhatriTT:
@saiarthiraguram do you maybe know our status on this?

saiarthiraguram:
> [@saiarthiraguram](https://github.com/saiarthiraguram) do you maybe know our status on this?

This model is on hold, will work on this once the torch is upgraded!

PR #1463, opened 2025-03-17
Surya_ocr Pytest #1291

saiarthiraguram:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-forge-fe/issues/1291)

### Problem description
- Added Pytest for Surya-ocr 

### What's changed
- The model consists of two PYPI dependencies.

1. Surya_ocr
2. Pydantic

- The surya-ocr 0.13.0 requires torch<3.0.0,>=2.5.1, but the tt-forge env have torch 2.1.0+cpu.cxx11.abi which is incompatible. Hence it is installed using no deps.
- Due to no-deps installation of Surya_ocr we face below errors.
`ImportError: cannot import name 'field_validator' from 'pydantic
ImportError: cannot import name 'computed_field' from 'pydantic'
`
-Hence pydantic is also installed using subprocess.
- The torch dependency leads to a Loader issue. 
` AttributeError: module 'torch.backends.cuda' has no attribute 'enable_cudnn_sdp'`
Hence the Loader is monkey patched using conftest.py.

### DIFF
--- a/forge/test/models/pytorch/vision/suryaocr/conftest.py
+++ b/forge/test/models/pytorch/vision/suryaocr/conftest.py
@@ -1,0 +1,57 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import sys
+from importlib.abc import MetaPathFinder
+from importlib.util import spec_from_loader
+from typing import Optional
+
+import torch
+
+
+class SuryaLoaderFinder(MetaPathFinder):
+    def find_spec(self, fullname, path, target=None):
+        if fullname == "surya.recognition.loader":
+            return spec_from_loader(fullname, SuryaLoader())
+
+
+class SuryaLoader:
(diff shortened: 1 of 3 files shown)


vkovinicTT:
I appreciate the effort in getting this to work, but relying on subprocesses to install dependencies might introduce maintainability issues. Would it be possible to explore upgrading PyTorch in our environment instead and see if that resolves the compatibility issues while not causing other problems?

vkovinicTT:
> * Torch upgradation might solve the issue, but I think we would alo need to take into account on the impact it might have on other models and PYPIs. Do you want me to take a look into that? @vkovinicTT

Yep that would be nice, since we should consider bumping up torch version in future :))

nvukobratTT:
@saiarthi-r @vkovinicTT I'm converting this one to Draft for easier PR prioritization until we determine if we can fix it with Torch upgrade. 

If we can't, let's push the ONNX variant with the offline loading until we determine strategy for pushing PyTorch models with tricky dependencies.

PR #1467, opened 2025-03-17
Add local model caching to fix transient CI failures

vkovinicTT:
## Problem
Sometimes on CI some tests fail to download model locally resulting in failing a test and not passing CI checks.

## Solution
This workaround enables us to cache model with method:

- `fetch_model` which downloads model with retry mechanisms from given url to our model cache repository if it's not present already in the model repo.  Cache repo location depends on whether env variable `FORGE_MODELS_CACHE` is set or not (having default location in latter case).

The changes help stabilize our CI pipeline and prevent unnecessary test failures.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -123,7 +123,7 @@ jobs:
       env:
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
         HF_HOME: /mnt/dockercache/huggingface
-        FORGE_MODELS_HUB: /mnt/dockercache/custom_models
+        FORGE_MODELS_HUB: /mnt/dockercache/forge_models_cache
         HF_HUB_DISABLE_PROGRESS_BARS: 1
         FORGE_DISABLE_REPORTIFY_DUMP: 1
         OPERATORS: ${{ inputs.operators }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -123,7 +123,7 @@ jobs:
       env:
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
         HF_HOME: /mnt/dockercache/huggingface
-        FORGE_MODELS_HUB: /mnt/dockercache/forge_models_cache
+        FORGE_MODELS_CACHE: /mnt/dockercache/forge_models_cache
         HF_HUB_DISABLE_PROGRESS_BARS: 1
         FORGE_DISABLE_REPORTIFY_DUMP: 1
         OPERATORS: ${{ inputs.operators }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -123,6 +123,7 @@ jobs:
       env:
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
         HF_HOME: /mnt/dockercache/huggingface
+        FORGE_MODELS_HUB: /mnt/dockercache/custom_models
         HF_HUB_DISABLE_PROGRESS_BARS: 1
         FORGE_DISABLE_REPORTIFY_DUMP: 1
         OPERATORS: ${{ inputs.operators }}
--- a/forge/test/utils.py
+++ b/forge/test/utils.py
@@ -35,6 +35,42 @@ def download_model(download_func, *args, num_retries=3, timeout=180, **kwargs):
     assert False, "Failed to download the model after multiple retries."
 
 
+def fetch_model(model_name, model_path, url, max_retries=3, timeout=30):
+    for attempt in range(1, max_retries + 1):
+        try:
+            print(f"Downloading {model_name} model, attempt {attempt}/{max_retries}...")
+            response = requests.get(url, timeout=timeout)
+            response.raise_for_status()
+
+            with open(f"{model_path}", "wb") as f:
+                f.write(response.content)
+
+            return True
+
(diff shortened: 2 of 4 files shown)


mstojkovicTT (approved):
LGTM

vkovinicTT:
> How is this caching anything? You should look into `/mnt/dockercache` which we use in the CI for caching (for now it looks like we only cache models from huggingface). We should do something similar for this.

Sorry, you are right, we are not caching anything, instead we are just downloading model everytime we run the test.

.github/workflows/test.yml
```diff
@@ -124,6 +124,7 @@ jobs:
       env:
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
         HF_HOME: /mnt/dockercache/huggingface
+        FORGE_MODELS_HUB: /mnt/dockercache/custom_models
```
pilkicTT, line 127: I would prefer something like:
```yaml
FORGE_MODELS_CACHE: /mnt/dockercache/forge_models_cache
```

forge/test/utils.py
```diff
@@ -35,6 +37,74 @@ def download_model(download_func, *args, num_retries=3, timeout=180, **kwargs):
     assert False, "Failed to download the model after multiple retries."
 
 
+def get_cache_dir() -> str:
+    """Get models cache directory from env var or use local default."""
+    cache_dir = os.environ.get("FORGE_MODELS_HUB")
+    if not cache_dir:
+        cache_dir = os.path.join(os.getcwd(), "cached_forge_models")
```
pilkicTT, line 44: make this directory `.forge_models_cache`
additionally you can add it to `.gitignore` as well...

pilkicTT (commented):
Minor comments, looks good otherwise... Have you ran the nightlies with this change?

.github/workflows/test.yml
vmilosevic: If this is specific to tt-forge-fe can it be under path /mnt/dockercache/tt-forge-fe/ ...

forge/test/utils.py
```diff
@@ -35,6 +37,74 @@ def download_model(download_func, *args, num_retries=3, timeout=180, **kwargs):
...
+    **kwargs: Any,
+) -> FrameworkModule:
+    """Fetch model from URL, cache it, and load it."""
+
+    model_file = model_name + ".pt"
+
+    model_path = os.path.join(get_cache_dir(), model_file)
+
+    # Download if needed
+    if not os.path.exists(model_path):
```
vmilosevic, line 82: Will the "fetch_model" recover in next run if download fails? It looks like that folder will be crated and it will not attempt the download again

forge/test/utils.py
```diff
@@ -35,6 +37,74 @@ def download_model(download_func, *args, num_retries=3, timeout=180, **kwargs):
...
+        os.makedirs(os.path.dirname(model_path), exist_ok=True)
+
+        for attempt in range(1, max_retries + 1):
```
vmilosevic, line 85: We are using shared (nfs) folder for file download we need to make sure that we handle concurrent "fetch_model" calls from multiple runners. Lokk into acquiring file lock before downloading

vmilosevic: https://chatgpt.com/share/67e28097-819c-8000-b35b-52f38087ca24

forge/test/utils.py
vkovinicTT: You are right, I changed that now.

forge/test/utils.py
vkovinicTT: I've made change already, thanks for the tips though :))

forge/test/utils.py
```diff
@@ -35,6 +38,95 @@ def download_model(download_func, *args, num_retries=3, timeout=180, **kwargs):
...
+                        print(f"Downloading {model_name}, attempt {attempt}/{max_retries}...")
+                        response = requests.get(url, timeout=timeout)
+                        response.raise_for_status()
+
+                        # Write to temporary file first
+                        temp_path = model_path + ".tmp"
+                        with open(temp_path, "wb") as f:
+                            f.write(response.content)
+
+                        # Atomic rename after successful download
```
pilkicTT, line 102: Why the rename step? why not directly writing it to the desired file?

vkovinicTT: This handles cases where a process might be terminated unexpectedly (e.g., by OOM killer, system shutdown, or network interruption) during download.
For example, if process A starts downloading a large model file and is killed halfway through, process B might see the incomplete file as valid and try to use it, causing errors. With temp file, process B would correctly detect that the download needs to be attempted again.

pilkicTT (approved):

.github/workflows/test.yml
```diff
@@ -123,6 +123,7 @@ jobs:
       env:
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
         HF_HOME: /mnt/dockercache/huggingface
+        FORGE_MODELS_CACHE: /mnt/dockercache/forge_models_cache
```
vmilosevic, line 126: If this is cache specific to tt-forge-fe project please set path to 
FORGE_MODELS_CACHE: /mnt/dockercache/tt-forge-fe/models_cache

vmilosevic (approved):

PR #1469, opened 2025-03-17
Fix nightly sweeps

vmilosevic:
### Ticket
N/A

### Problem description
Runs-on string wasn't properly substituted, causing "nightly sweeps" pipelines to stop

### What's changed
Fixed by splitting into two jobs, for on dispatch and on schedule

### Checklist
- [x ] New/Existing tests provide coverage for changes


.github/workflows/on-nightly-sweeps.yml
```diff
@@ -31,7 +28,24 @@ jobs:
     secrets: inherit
     with:
       docker-image: ${{ needs.docker-build.outputs.docker-image }}
-  test:
+
+  test_on_dispatch:
```
pilkicTT, line 32: is this step doing anything? also, couldn't we have both of these (`on_dispatch` and `on_schedule`) as a single step?

pilkicTT: Ah, sorry, it uses the `test.yml` workflow

pilkicTT (approved):

.github/workflows/on-nightly-sweeps.yml
vmilosevic: Btw whole thing is a workaround as string escape inside json doesn't work well and we can't set default values when running on schedule.

This is what we would like to have
```
runs-on: '[{"runs-on": "${{ github.event.inputs.runs-on || 'n150' "}]'
```
But the only way I managed to get it working is by adding two jobs with conditions

PR #1470, opened 2025-03-18
Add test files for gemma

meenakshiramanathan1:
gemma-1.1-2b-it is passing end-to-end and occupies 19.8 GB, hence it has been added with nightly and push markers.
gemma-1.1-7b-it requires over 50 GB, so it has been skipped.
gemma-2-2b-it fails with RuntimeError: Trying to access element outside of dimensions: 2 which has been added in xfail
google/gemma-2-9b-it requires over 50 GB, so it has been skipped.


ashokkumarkannan1 (approved):
LGTM!

PR #1484, opened 2025-03-18
[runtime] tensors on device aka runtime stitching

### LINKED ISSUE #1245: [runtime] introduce forge tensor
Introduce new type for the runtime tensor. The runtime tensor should help us with implementing runtime stitching.

The runtime tensors should be able to hold handles to both host side tensors and device runtime tensors.

Additionally, in the near future, it should help us with propagating gradients during training runs - where there are multiple separate backward graphs which feed into one another. Currently, there is a hack for that in the `compiled_graph_state.py`:
```python
        # Pass on the calculated gradients to the attached module
        if self.attached_module is not None:
            # pass on the calculated gradients and call the attached module's backward pass
            # HACK: we don't have a way to know which gradient outputs are tied to which gradient inputs
            # of the attached module. For now, just attach the first one since we are doing this only for
            # the loss module (which will have only one gradient output) and the model will need only one
            # gradient output to be passed to the loss module.
            assert len(self.gradient_outputs) == 1, "Passing gradients not properly implemented yet"
            assert len(self.attached_module.gradient_inputs) == 1, "Passing gradients not properly implemented yet"
            self.attached_module.gradient_inputs[0] = self.gradient_outputs[0]
            self.attached_module.backward()
```

### LINKED ISSUE #1329: [runtime] enable runtime stitching
The `tt-mlir` runtime now leaves every tensor on device and can accept inputs already on device. Use that to avoid unnecessary movements of tensors to and from the device.

For example, when running inference of a model in a loop, we are currently sending the weights/constant tensors on each iteration. This is obviosly not ideal.

Another example would be in training scenario. We can now reuse outputs of, let's say, `forward()` program on device as inputs to the `backward()` program. Without needing to move outputs of `forward()` to the host and then running the `backward()` program which will move them back to the device.

pilkicTT:
This change implements a mechanism to keep certain tensors on device, so that running programs in a loop (or chaining multiple programs) is more efficient. We call this - runtime stitching.

Previously, we were sending each input to device each time we ran a program - which is problematic, since this includes all of the weights, which for example remain the same during multiple invocations of the forward program (in inference) and don't need to be sent to the device each time.

To keep the tensors on the device we only need to hold on to the handles of the tensors provided by the device runtime (`tt-mlir` runtime) so that the tensors don't get deallocated; and pass them on when needed for executing a program. This change introduces several structures for enabling us to that. Tensors which are kept on the device are called `persistent`.

Summary of the new structures:
- `ModelState` which holds info for running the model (module), i.e. multiple programs (e.g. tensor pool, all program states)
- `ProgramState` - holds info needed for running a particular program (e.g. persistent input tensors).
- `TensorPool` simple container for holding named tensors; in the context of the `ModelState` it will contain all `persistent` tensors.
- `Tensor` which represents forge runtime tensor and holds both the handle to the runtime tensor (if exists) and the host storage handle - the torch tensor from which it was created. This tensor object is similar to the torch tensor, in the sense that the object contains (in this case `shared_ptr`) pointer to the actual tensor data. Thus, copying a tensor only creates a new handle to the underlaying data.

The rest of the change is minimal refactor of existing code to incorporate new structures. Additionally, some of the training tests were modified so that they remain functional; additional handling is needed to ensure proper updating of host/device weight tensors, because training tests combine executing some parts of the graphs/modules both on the device and cpu.

There is no additional test coverage added and the correctness relies on the existing e2e tests. More changes are yet to come with the goal of creating a more performant and robust forge runtime. More fine grained functional tests of the runtime component will be added.

Closes #1245, Closes #1329

### DIFF
--- a/forge/csrc/runtime/tensor.hpp
+++ b/forge/csrc/runtime/tensor.hpp
@@ -1,0 +1,343 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <cstring>
+#include <optional>
+#include <variant>
+
+#include "runtime/tt_device.hpp"
+#include "torch/torch.h"
+#include "tt/runtime/runtime.h"
+#include "tt/runtime/types.h"
+#include "utils/assert.hpp"
+
+namespace tt
+{
+
+target::DataType torch_scalar_type_to_dt(torch::ScalarType st);
+torch::ScalarType dt_to_torch_scalar_type(target::DataType df);
+
+template <typename T>
+std::vector<int64_t> as_vec_int64(std::vector<T> const& vec)
+{
+    std::vector<int64_t> result;
+    result.reserve(vec.size());
+    for (auto const& v : vec)
+    {
+        result.push_back(v);
+    }
+    return result;
+}
+
+// Class which wraps the host storage of a tensor.
+//
+// NOTE: currently this doesn't make much sense, since we only have one type of host storage (torch.Tensor).
+// However, i expect that we will soon introduce at least one additional type of storage, so i've decided to introduce
+// this class. Remove this if i turn out to be an optimist.
+class TensorHostStorage
+{
+   public:
+    explicit TensorHostStorage(torch::Tensor& tensor) : storage(tensor) {}
+
+    // Creates a new shared_ptr to the data of the tensor.
+    // NOTE: the shared_ptr is not owning the data and will not ever release it (see the lambda for the deleter),
+    // it is used only to provide a shared_ptr interface to the data which the device runtime (`tt-mlir` runtime)
+    // expects.
+    std::shared_ptr<void> borrow_data() const
+    {
+        return std::visit(
+            [](auto&& arg) -> std::shared_ptr<void>
+            {
+                using T = std::decay_t<decltype(arg)>;
+                if constexpr (std::is_same_v<T, torch::Tensor>)
+                {
+                    return std::shared_ptr<void>(arg.data_ptr(), [arg](void*) { (void)arg; });
+                }
+            },
+            storage);
+    }
+
+    void* data_ptr() const
+    {
+        return std::visit(
+            [](auto&& arg) -> void*
+            {
+                using T = std::decay_t<decltype(arg)>;
+                if constexpr (std::is_same_v<T, torch::Tensor>)
+                {
+                    return arg.data_ptr();
+                }
+            },
+            storage);
+    }
+
+    template <typename T>
+    static TensorHostStorage from_desc(runtime::TensorDesc desc)
+    {
+        static_assert(std::is_same_v<T, torch::Tensor>, "Currently only torch::Tensor is supported");
+
+        if constexpr (std::is_same_v<T, torch::Tensor>)
+        {
+            auto shape = as_vec_int64(desc.shape);
+            auto stride = as_vec_int64(desc.stride);
+            auto dtype = dt_to_torch_scalar_type(desc.dataType);
+
+            auto tensor = at::empty_strided(shape, stride, dtype);
+            return TensorHostStorage(tensor);
+        }
+    }
+
+    std::variant<torch::Tensor> storage;
+};
+
+// Core implementation behind the `Tensor` class.
+//
+// It holds the tensor description and the actual data.
+//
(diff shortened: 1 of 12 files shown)


pilkicTT:
P.S. Sorry for the large change folks...

forge/csrc/runtime/tensor.hpp
```diff
@@ -0,0 +1,307 @@
...
+        return host_storage->borrow_data();
+    }
+
+    runtime::Tensor& get_runtime_tensor()
+    {
+        TT_ASSERT(rt_tensor.has_value());
+        return *rt_tensor;
+    }
+
+    torch::Tensor to_torch() const
```
pglusacTT, line 105: Is the idea here to explicitly sync from device to host? If it's not on device, why create a copy, what's the use case? Should we have `get_host_tensor` similar to `get_runtime_tensor` that just gets the tensors, without the copy and sync?

Also, doesn't having `to_torch` kinda break the idea of `TensorHostStorage`, should torch specific logic be moved there? Should this be called `to_host` or `move_to_host` or something to that extent?

forge/csrc/runtime/tensor.hpp
```diff
@@ -0,0 +1,307 @@
...
+
+inline target::DataType torch_scalar_type_to_dt(torch::ScalarType st)
+{
+    switch (st)
+    {
+        case torch::ScalarType::Byte: return target::DataType::UInt8;
+        case torch::ScalarType::Char: return target::DataType::UInt8;
+        case torch::ScalarType::Short: return target::DataType::UInt16;
+        case torch::ScalarType::Int: return target::DataType::Int32;
+        case torch::ScalarType::Long: return target::DataType::UInt32;
```
pglusacTT, line 275: Isn't Long a 64-bit int? How are we handling this?

forge/forge/compiled_graph_state.py
```diff
@@ -313,27 +356,30 @@ def backward(self) -> List[torch.Tensor]:
...
 
         if self.optimizer_on_device():
             if self.gradient_outputs is None or len(self.gradient_outputs) == 0:
                 self.gradient_outputs = grads
             else:
+                # TODO: How to handle gradient accumulation?
                 assert len(self.gradient_outputs) == len(grads), "Number of gradients does not match number of outputs"
-                for idx, grad in enumerate(grads):
-                    self.gradient_outputs[idx] += grad
+                assert False, "Gradient accumulation not implemented yet"
```
pglusacTT, line 371: Could this be a new program type that just does grad acc?
It can take past and current gradients and just sums up and writes to output, what do you think? Just food for thought 😄

vladimirjovanovicTT: This is a good topic for the next meeting :)

forge/csrc/runtime/tensor.hpp
pilkicTT: modified the `to_torch` to return the host storage tensor and initiate `to_host()` if there is no host storage present

reegarding the second part of the question, `TensorHostStorage` is our internal structure - the user sees only `Tensor` and it should be valid to convert `Tensor` to `torch::Tensor` if the user wants it.

pglusacTT (approved):
🚀

forge/csrc/runtime/tensor.hpp
```diff
@@ -0,0 +1,343 @@
...
+        static_assert(std::is_same_v<T, torch::Tensor>, "Currently only torch::Tensor is supported");
+
+        if constexpr (std::is_same_v<T, torch::Tensor>)
+        {
+            auto shape = as_vec_int64(desc.shape);
+            auto stride = as_vec_int64(desc.stride);
+            auto dtype = dt_to_torch_scalar_type(desc.dataType);
+
+            auto tensor = at::empty_strided(shape, stride, dtype);
+            return TensorHostStorage(tensor);
```
dgolubovicTT, line 88: So if the tensor is created from TensorDesc, then TensorHostStorage owns the tensor? And if it created from existing pytorch tensor, then TensorHostStorage doesn't own the tensor?

pilkicTT: Sorry, missed this one... The ownership in both cases is shared, similar as it is the case with a `shared_ptr<T>`.

The torch tensors which we use for host storage, are refcounted so they won't be released until the last user finished with them. So, even in the case when we create `tt::Tensor` and its `TensorHostStorage` from `TensorDesc`, the user could ask for it via `Tensor::to_torch()` and use it as long as he holds the reference.

PR #1495, opened 2025-03-20
Increase batch size for resnet benchmark model

vcanicTT:
Increase batch size for resnet benchmark model.

### DIFF
--- a/.github/workflows/perf-benchmark.yml
+++ b/.github/workflows/perf-benchmark.yml
@@ -89,7 +89,7 @@ jobs:
       run: |
         source env/activate
         python forge/test/benchmark/benchmark.py -m mnist_linear -bs 32 -lp 32 -o benchmark_forge_e2e_mnist_32_32_${{ steps.fetch-job-id.outputs.job_id }}.json
-        python forge/test/benchmark/benchmark.py -m resnet50_hf -bs 1 -lp 32 -o benchmark_forge_e2e_rasnet50_1_32_${{ steps.fetch-job-id.outputs.job_id }}.json
+        python forge/test/benchmark/benchmark.py -m resnet50_hf -bs 3 -lp 32 -o benchmark_forge_e2e_rasnet50_1_32_${{ steps.fetch-job-id.outputs.job_id }}.json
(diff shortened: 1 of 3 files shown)


.github/workflows/perf-benchmark.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       run: |
         source env/activate
         python forge/test/benchmark/benchmark.py -m mnist_linear -bs 32 -lp 32 -o benchmark_forge_e2e_mnist_32_32_${{ steps.fetch-job-id.outputs.job_id }}.json
-        python forge/test/benchmark/benchmark.py -m resnet50_hf -bs 1 -lp 32 -o benchmark_forge_e2e_rasnet50_1_32_${{ steps.fetch-job-id.outputs.job_id }}.json
+        python forge/test/benchmark/benchmark.py -m resnet50_hf -bs 3 -lp 32 -o benchmark_forge_e2e_rasnet50_1_32_${{ steps.fetch-job-id.outputs.job_id }}.json
```
odjuricicTT, line 92: Could we use run_benchmark_wh.sh in the job so we don't have multiple places where we set batch size?

vcanicTT:
@odjuricicTT @pilkicTT Can you take a look?

odjuricicTT (approved):
Resnet :rocket:

forge/test/benchmark/benchmark/models/resnet_hf.py
```diff
@@ -67,9 +67,6 @@ def test_resnet_hf(
     if training:
         pytest.skip("Training is not supported")
 
-    if batch_size > 1:
```
pilkicTT, line 70: should we update the batch size configuration above?

pilkicTT (approved):

.github/workflows/perf-benchmark.yml
vcanicTT: I am not sure how we run these commands in CI, but I will communicate with CI people, to run one test in one job, because if something fails jobs are independent and don't affect other jobs and test. And bash script is more for internal use.

Issue #1502, opened 2025-03-21
[verifications] fix host OOM issues when running CI


pilkicTT:
There are some tests which use large amount of host memory which causes flakiness in the CI.

E.g. `test_specific_ops_llama32.py::test_transpose[param10]`

The issues in this test case, stem from large input/output tensor, which combined with current implementations of different verifications (both during compile and after the runtime is completed) cause memory consumption to be ~20x.

pilkicTT:
On branch [pilkic/memory-issue-locations](https://github.com/tenstorrent/tt-forge-fe/tree/pilkic/memory-issue-locations) there are parts of the code that are contributing to the abnormal usage of memory during different verification steps.

The biggest contribution is our `verify` function, more precisely, all of the atol, pcc checks.

PR #1503, opened 2025-03-21
[ci] skip large transpose test

pilkicTT:
Due to multiple issues in verification steps, this test uses large amount of host memory, which in turn causes CI flakinees (OOM issues).

The OOM issues don't always occur, since the amount of free memory depends on how fast garbage collector is able to clean things up (between tests).

Skipping for now, until #1502 is resolved.


vkovinicTT (approved):

pilkicTT:
Merging now to reduce issues with CI.

PR #1508, opened 2025-03-21
[Dtypes] Added support for uint8

vkovinicTT:
This PR mainly addresses issues with `Cast` OP:
- added support for `uint8`
- casting `boolean` to `uint8` (with warning)
- casting `int8` to `int32` since `int8` is not supported in `ttnn` (with warning)
- removed xfail from most of the test since changes above fix the failing reason
- for some tests I updated failing reason since it has changed


mstojkovicTT (approved):
Looks good! Also fixes a future bug with Llama 3.2 Decode :))

dgolubovicTT (approved):

PR #1517, opened 2025-03-24
Restructure stable_diffusion_xl test

meenakshiramanathan1:
Original stable_diffusion_xl has been restructured since model is loaded using a pipeline that includes multiple text encoders (CLIPTextModelWithProjection, CLIPTextModel), tokenizers, a EulerDiscreteScheduler, an UNet2DConditionModel, and an AutoencoderKL VAE.Since the pipeline lacks a direct forward function call that can be passed to forge.compile, the pipeline structure has been broken down.
All preprocessing steps have been manually extracted from StableDiffusionXLPipeline, and only the unet forward function call is now passed to forge.compile.
Model currently fails with `"NotImplementedError: The following operators are not implemented: ['aten::concat']"` during frontend conversion which will be debugged later.

### DIFF
--- a/forge/test/models/pytorch/multimodal/stable_diffusion/utils/model.py
+++ b/forge/test/models/pytorch/multimodal/stable_diffusion/utils/model.py
@@ -3,10 +3,11 @@
 
 # Stable Diffusion Demo Script
 
-from typing import List, Optional, Union
+
+from typing import List, Optional, Tuple, Union
 
 import torch
-from diffusers import StableDiffusion3Pipeline
+from diffusers import DiffusionPipeline, StableDiffusion3Pipeline
 from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion import (
     retrieve_timesteps,
 )
@@ -16,15 +17,23 @@ import forge
 from test.models.utils import Framework, build_module_name
 
 
-def load_pipe_v35(variant):
-    pipe = StableDiffusion3Pipeline.from_pretrained(f"stabilityai/{variant}", torch_dtype=torch.float32)
+def load_pipe(variant, variant_type):
+    if variant_type == "v35":
+        pipe = StableDiffusion3Pipeline.from_pretrained(f"stabilityai/{variant}", torch_dtype=torch.float32)
+        modules = [pipe.text_encoder, pipe.transformer, pipe.text_encoder_2, pipe.text_encoder_3, pipe.vae]
+    elif variant_type == "xl":
+        pipe = DiffusionPipeline.from_pretrained(f"stabilityai/{variant}", torch_dtype=torch.float32)
+        modules = [pipe.text_encoder, pipe.unet, pipe.text_encoder_2, pipe.vae]
+
+    # Move the pipeline to CPU
     pipe.to("cpu")
-    modules = [pipe.text_encoder, pipe.transformer, pipe.text_encoder_2, pipe.text_encoder_3, pipe.vae]
+
     for module in modules:
         module.eval()
         for param in module.parameters():
             if param.requires_grad:
                 param.requires_grad = False
+
     return pipe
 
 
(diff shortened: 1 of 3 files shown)


saiarthiraguram (approved):

forge/test/models/pytorch/multimodal/stable_diffusion/utils/model.py
```diff
@@ -217,7 +223,148 @@ def stable_diffusion_preprocessing_v35(
...
+        (num_images_per_prompt, num_channels_latents, height // pipe.vae_scale_factor, width // pipe.vae_scale_factor),
+        device=device,
+    )
+    latents = latents * pipe.scheduler.init_noise_sigma
+    add_text_embeds = pooled_prompt_embeds
+    if pipe.text_encoder_2 is None:
+        text_encoder_projection_dim = int(pooled_prompt_embeds.shape[-1])
+    else:
+        text_encoder_projection_dim = pipe.text_encoder_2.config.projection_dim
+    target_size = target_size or (height, width)
```
mstojkovicTT, line 317: Is there a reason to do this twice? (other one is at line 257)

forge/test/models/pytorch/multimodal/stable_diffusion/utils/model.py
```diff
@@ -217,7 +223,148 @@ def stable_diffusion_preprocessing_v35(
...
+        batch_size = prompt_embeds.shape[0]
+    num_channels_latents = pipe.unet.config.in_channels
+    shape = (
+        batch_size,
+        num_channels_latents,
+        int(height) // pipe.vae_scale_factor,
+        int(width) // pipe.vae_scale_factor,
+    )
+    latents = torch.randn(
+        (num_images_per_prompt, num_channels_latents, height // pipe.vae_scale_factor, width // pipe.vae_scale_factor),
```
mstojkovicTT, line 308: Should `batch_size` be factored in here as well? Something like
``` python
latents = torch.randn(
    (batch_size * num_images_per_prompt, num_channels_latents, height // pipe.vae_scale_factor, width // pipe.vae_scale_factor),
    device=device,
)
```

forge/test/models/pytorch/multimodal/stable_diffusion/utils/model.py
```diff
@@ -217,7 +223,148 @@ def stable_diffusion_preprocessing_v35(
...
+    timesteps, num_inference_steps = retrieve_timesteps(
+        pipe.scheduler,
+        num_inference_steps=num_inference_steps,
+        device=device,
+        timesteps=timesteps,
+        sigmas=sigmas,
+    )
+
+    # 3. Prepare latent variables
+    if prompt is not None and isinstance(prompt, str):
```
mstojkovicTT, line 294: No need to explicitly check `prompt is not None`, it's redundant due to the `isinstance` checks already handling that :))
Try something like:
``` python
if isinstance(prompt, str):
    batch_size = 1
elif isinstance(prompt, list):
    batch_size = len(prompt)
else:
    batch_size = prompt_embeds.shape[0]
```

mstojkovicTT (changes requested):

forge/test/models/pytorch/multimodal/stable_diffusion/utils/model.py
meenakshiramanathan1: Removed

forge/test/models/pytorch/multimodal/stable_diffusion/utils/model.py
meenakshiramanathan1: Modified

forge/test/models/pytorch/multimodal/stable_diffusion/utils/model.py
meenakshiramanathan1: Modified

mstojkovicTT (approved):

vmilosevic (approved):

PR #1521, opened 2025-03-24
Script for reading TTIR and parsing device perf data for E2E benchmark tests

vcanicTT:
Added script for writing device perf into the JSON file, that we later write into the database. First we generate ttir file by running benchmark tests. Then we modify those ttirs, and run ttrt on them, when we get device perf, we write it into the database.


vcanicTT:
@pilkicTT @nsumrakTT Can you take a look?

nsumrakTT:
Just a note: ttrt should run after all forge-fe related activities because its requires different versions of the similar py packages. So, we need to activate mlir env and reinstall those packages.

nsumrakTT (approved):

forge/test/benchmark/test_data/device_perf/ttir.mlir
```diff
@@ -0,0 +1,4 @@
...
+    "module": "MNISTLinear"
+}
```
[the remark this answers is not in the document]
vcanicTT: Yes, I used them during the development, and decided to leave them as simple tests. Also, inside the script in the main function, I made an example how to use the script with these tests.

forge/test/benchmark/device_perf.py
```diff
@@ -0,0 +1,283 @@
...
+    -------
+    parsed_args: dict
+        The parsed arguments from the command line.
+    """
+
+    parser = argparse.ArgumentParser(description="Get device perf for benchmark end-to-end tests.")
+
+    parser.add_argument("-ct", "--create_ttir", help="Create TTIR file from the given path.")
+
+    parser.add_argument("-cdp", "--create_device_perf", nargs=2, help="Create device performance data.")
```
pilkicTT, line 221: Majority of CLI tools use `--long-param-name` instead of `--long_param_name`, so it would be good to align on this...

forge/test/benchmark/device_perf.py
```diff
@@ -0,0 +1,283 @@
...
+    attr_definition = "attributes {tt.system_desc = #system_desc}"
+    attr_empty = "attributes {}"
+    content[1] = content[1].replace(attr_definition, attr_empty)
+    content = content[1:]  # Remove the first line
+
+    # At the beginning of the content, we need to add two commands
+    # The first command is to ttmlir-optimize the module
+    # The second command is to ttmlir-translate the module
+    ttmlir_optimize = (
+        '// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=%system_desc_path%" -o out.mlir %s'
```
pilkicTT, line 94: I think it would be best if we can leave these `RUN` commands for the workflow which will actually execute the benchmark...

pilkicTT (approved):
Little bit too many string literals for my taste, but for the initial setup i guess it's ok. Moving forward let's discuss on how can we tighten this up...

forge/test/benchmark/device_perf.py
vcanicTT: Yes, I'll change.

forge/test/benchmark/device_perf.py
vcanicTT: Yes, I'll test

forge/test/benchmark/device_perf.py
vcanicTT: Done.

PR #1539, opened 2025-03-25
[CI] Changed default execution depth from "" -> CI_FAILURE

vkovinicTT:
### Problem description
Previously execution depth wasn't set on beginning so if we were to fail test before compile stage (e.g. while downloading model), in our database `bringup_status` would be `N/A` and now we changed it to `CI_FAILURE` per @tt-mpantic request, so now if it fails during preprocessing it will show `CI_FAILURE`.

### What's changed
Changed default execution depth from "" -> `CI_FAILURE`.

### DIFF
--- a/forge/csrc/shared_utils/forge_property_utils.hpp
+++ b/forge/csrc/shared_utils/forge_property_utils.hpp
@@ -19,6 +19,7 @@ namespace tt::property
 
 enum class ExecutionDepth
 {
+    CI_FAILURE,                 // CI failure
     FAILED_FE_COMPILATION,      // Front end compilation fails (tvm->ttir)
     FAILED_TTMLIR_COMPILATION,  // TT-MLIR compilation fails, can't produce flatbuffer
     FAILED_RUNTIME,             // Runtime execution fails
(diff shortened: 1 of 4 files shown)


dgolubovicTT (approved):
LGTM

forge/csrc/shared_utils/forge_property_utils.hpp
```diff
@@ -19,6 +19,7 @@ namespace tt::property
 
 enum class ExecutionDepth
 {
+    CI_FAILURE,                 // CI failure (e.g. download of model weights fails)
```
nvukobratTT, line 22: cc @AleksKnezevic @mrakitaTT for alignment

Feel free to ping me on Slack if you disagree or have other opinions on having this addition on other FEs as well

nvukobratTT (approved):

forge/csrc/shared_utils/forge_property_utils.hpp
mrakitaTT: Looks good to me, I even wanted to suggest something similar because we currently have many models failing in CI because of OOM issues (specific for tt-xla), but we were using `FAILED_FE_COMPILATION` depth in the meantime.

PR #1551, opened 2025-03-26
Test for ONNX BiLSTM_CRF 

saiarthiraguram:
### Ticket
Fixes 
(https://github.com/tenstorrent/tt-forge-fe/issues/1550)

### DIFF
--- a/forge/test/models/onnx/text/bi_lstm_crf/test_bi_lstm_crf.py
+++ b/forge/test/models/onnx/text/bi_lstm_crf/test_bi_lstm_crf.py
@@ -1,0 +1,59 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import torch
+import onnx
+
+import forge
+from forge.verify.verify import verify
+
+from test.models.onnx.text.bi_lstm_crf.utils.model import get_model
+from test.models.utils import Framework, Source, Task, build_module_name
+
+# Opset 11 is the minimum version to support bi_lstm_crf in Torch.
+# Opset 17 is the maximum version in Torchscript.
+opset_versions = [11, 17]
+
+
+@pytest.mark.nightly
+@pytest.mark.xfail(
+    reason="NotImplementedError: The following operators are not implemented: ['aten::_pad_packed_sequence', 'aten::_pack_padded_sequence']"
+)
+@pytest.mark.parametrize("opset_version", opset_versions, ids=opset_versions)
+def test_birnn_crf_pypi(forge_property_recorder, tmp_path, opset_version):
+
+    # Build Module Name
+    module_name = build_module_name(
+        framework=Framework.PYTORCH,
+        model="BiRnnCrf_PyPI",
+        task=Task.TOKEN_CLASSIFICATION,
+        source=Source.GITHUB,
+    )
+
+    # Record Forge Property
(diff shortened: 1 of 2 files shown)


forge/test/models/onnx/text/bi_lstm_crf/test_bi_lstm_crf.py
```diff
@@ -0,0 +1,59 @@
...
+    # Build Module Name
+    module_name = build_module_name(
+        framework=Framework.PYTORCH,
+        model="BiRnnCrf_PyPI",
+        task=Task.TOKEN_CLASSIFICATION,
+        source=Source.GITHUB,
+    )
+
+    # Record Forge Property
+    forge_property_recorder.record_group("priority")
```
mstojkovicTT, line 36: This should be "red" instead, as per [PR](https://github.com/tenstorrent/tt-forge-fe/pull/1573) :))

forge/test/models/onnx/text/bi_lstm_crf/test_bi_lstm_crf.py
```diff
@@ -0,0 +1,59 @@
...
+
+import forge
+from forge.verify.verify import verify
+
+from test.models.onnx.text.bi_lstm_crf.utils.model import get_model
+from test.models.utils import Framework, Source, Task, build_module_name
+
+# Opset 11 is the minimum version to support bi_lstm_crf in Torch.
+# Opset 17 is the maximum version in Torchscript.
+opset_versions = [11, 17]
```
mstojkovicTT, line 17: Same question as in [PR](https://github.com/tenstorrent/tt-forge-fe/pull/1554)

forge/test/models/onnx/text/bi_lstm_crf/utils/model.py
```diff
@@ -0,0 +1,43 @@
...
+        if isinstance(tag_seq, list):
+            tag_seq = torch.tensor(tag_seq, dtype=torch.long)
+        return scores, tag_seq
+
+
+def get_model(test_sentence):
+    word_to_ix = {"apple": 0, "corporation": 1, "is": 2, "in": 3, "georgia": 4, "<PAD>": 5}
+    tag_to_ix = {"B": 0, "I": 1, "O": 2}
+
+    embedding_dim = 5
```
mstojkovicTT, line 26: Can this be calculated in some other way then having it hardcoded :))

saiarthiraguram: 1. In the [repo](https://github.com/jidasheng/bi-lstm-crf/blob/master/README.md) shared, the model loads the above parameters from a json file.
2.  The options are to train the model, save the model and load the params which would have the model load as a dependency or have some sample input as in the code right now.
3. Please let me know on how would you like me to proceed.

mstojkovicTT: I was thinking something like `word_to_ix` has 6 elements, and `embedding_dim` is surely somehow correlated to that, something like `len(word_to_ix)-1`. If that is the case, i would do that instead :))

saiarthiraguram: > I was thinking something like `word_to_ix` has 6 elements, and `embedding_dim` is surely somehow correlated to that, something like `len(word_to_ix)-1`. If that is the case, i would do that instead :))

Add the changes in recent commit, please have a look

PR #1557, opened 2025-03-27
[CI] Improving stability of nightly runs

vkovinicTT:
### What's changed
- Some tests that were reporting good `pcc` weren't actually checking `pcc` at all, so now that's fixed. While it might degrade our statistics, at least they will be correct :))
- some tests were missing `group` tag which caused `N/A` values in superset, so I added them. Also some values will still be `N/A` but the problem is not on our part, so we opened [ticket](https://tenstorrent.atlassian.net/browse/DATA-640) for database team
- skipped one model which caused our `On nightly xfail` workflow to crash, hopefully now we will have green `On nightly xfail`.

### DIFF
--- a/forge/test/models/pytorch/vision/monodle/test_monodle.py
+++ b/forge/test/models/pytorch/vision/monodle/test_monodle.py
@@ -15,6 +15,7 @@ from test.models.utils import Framework, Source, Task, build_module_name
 
 @pytest.mark.nightly
 @pytest.mark.xfail(reason="[Conv2dTranspose][Shape Calculation] TypeError: 'int' object is not subscriptable")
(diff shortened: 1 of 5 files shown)


forge/test/models/pytorch/vision/monodle/test_monodle.py
```diff
@@ -15,6 +15,7 @@
 
 @pytest.mark.nightly
 @pytest.mark.xfail(reason="[Conv2dTranspose][Shape Calculation] TypeError: 'int' object is not subscriptable")
+@pytest.mark.skip(reason="Floating point exception(core dumped)")
```
pilkicTT, line 18: It would be good to reproduce this on ttnn and file an issue...

pilkicTT (approved):

PR #1567, opened 2025-03-27
[verification] optimize memory usage

### LINKED ISSUE #1534: [verification] improve pcc check memory usage
In the existing implementation there are multiple new tensor being created, i.e. multiple copies of the data are created. Requiring 10x more memory than total input size (golden + calculated tensor).

### LINKED ISSUE #1562: [verification] cleanup atol calculation
Cleanup the function `calculate_atol()` which calculates maximum absolute difference between tensors. It's important to keep the memory usage as minimal as possible.

### LINKED ISSUE #1563: [verification] disable torch jit trace verification
When creating the torch jit trace, torch by default executes verification (`check_trace`) - which creates multiple copies (clones) of input tensors. This can be problematic with regards to the total memory usage if the input tensors are large.

Disable this verification by default, since we are anyways doing the verifications on our end (after TVM).

pilkicTT:
We've been hitting sporadic out of memory failures on CI; due to different verification code paths taking up significant amount of memory (Issue #1502).

This change modifies the following:
- disables `torch.jit.trace` verification by default (`check_trace = False`); we are already verifying everything after TVM is done
- optimizes memory usage for large tensors during pcc calculations; large tensors are split into chunks and processed in parallel to calculate pcc estimation - more info in the comments
- refactors `calculate_atol()` function

For one of the most extreme cases in our CI, the transpose test with 1GB input, goes from ~20GB to ~9GB total memory usage; which still isn't ideal but it's a step forward.

Added sanity tests to ensure that the pcc/atol calculation makes sense.

Closes #1534, Closes #1562, Closes #1563

### DIFF
--- a/forge/forge/verify/compare.py
+++ b/forge/forge/verify/compare.py
@@ -3,8 +3,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-from typing import Union
+from multiprocessing.pool import ThreadPool
 import os
+from typing import Union
 
 import torch
 import tensorflow as tf
@@ -12,8 +13,6 @@ import numpy as np
 from loguru import logger
 from scipy.spatial import distance
 from typing import Union, Tuple, List, Optional
-
-from forge.tensor import narrow_forge_tensor_to_pytorch
 
 # Compares golden and calculated tensors. Using allclose for scalar values, rogerstanimoto for bool tensors, pcc otherwise
 def compare_with_golden(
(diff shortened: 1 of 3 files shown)


forge/forge/verify/compare.py
```diff
@@ -75,50 +74,99 @@ def compare_dissimilarity(calculated_dissimilarity: float, dissimilarity_thresho
...
+      and is quite simple.
 
-    if a.dtype == torch.bfloat16:
+      The calculations over chunks are done in parallel using ThreadPool, to lower the execution time.
+
+      NOTES: original implementation is flattening both tensors and then calculating the PCC, this means
+      that the whole tensor is treated as a single random variable with N samples (N = number of elements in the tensor).
+      With the estimation procedure we are treating the tensor as a collection of random variables, each chunk is treated
+      like a single random variable (so N_chunks in total) with M samples (M = number of elements in the chunk). The final
+      PCC is the minimum of all the PCCs calculated over the chunks.
```
vkovinicTT, line 109: Why is it minimum of all pcc calculations over chunks?

nvukobratTT: +1 on this Q. Shouldn't we track average chunk PCC to have a more precise estimate?

forge/forge/verify/compare.py
```diff
@@ -75,50 +74,99 @@ def compare_dissimilarity(calculated_dissimilarity: float, dissimilarity_thresho
...
-    )
 
     if isinstance(pcc, np.ma.core.MaskedConstant):
         return 1.0
 
     return pcc
 
 
+def calculate_pcc(a: torch.Tensor, b: torch.Tensor) -> np.float64:
+    TENSOR_SIZE_THRESHOLD = 100 * 1000000
```
nvukobratTT, line 165: You can use this notation to make it cleaner: `1.e08`.

forge/forge/verify/compare.py
pilkicTT: I wanted to test if a more pesimistic estimation will still satisfy existing tests. Changed it to averaging the chunk pccs.

Issue #1529, opened 2025-03-25
ONNX ModelProto Exceeds 2GB Limit Causing Serialization Error in Forge Compilation


kamalrajkannan78:
## Description

-  While compiling the Phi2 ONNX model using Forge, the following error occurs due to the model size exceeding the 2GB limit for Protobuf serialization

## Reproduce

```
git checkout kkannan/onnx_inference_issue_in_large_models
git submodule update --recursive
pytest forge/test/models/onnx/text/phi2/test_phi2.py -svv
```

## Observed Behaviour
```
self = Module ModelProto
args = (tensor([[16594,   257,  6496, 23970,  1022, 19473,   290,   257, 46371,    13,
         50295, 50295, 50295, 50295, 5...0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]]))
kwargs = {}
ort = <module 'onnxruntime' from '/opt/ttforge-toolchain/venv/lib/python3.10/site-packages/onnxruntime/__init__.py'>
so = <onnxruntime.capi.onnxruntime_pybind11_state.SessionOptions object at 0x7f96545764b0>

    def forward(self, *args, **kwargs):
        import onnxruntime as ort
    
        so = ort.SessionOptions()
        so.inter_op_num_threads = 2
        so.intra_op_num_threads = 2
    
>       module_bytes = self.module.SerializeToString()
E       ValueError: Message onnx.ModelProto exceeds maximum protobuf size of 2GB: 11119721893
```

## Logs

- [phi2_onnx.log](https://github.com/user-attachments/files/19443081/mar25_phi2_onnx_using_opti.log)

kamalrajkannan78:
@sott0n , @nvukobratTT  Could you please share the reason behind the changes in [PR #1408](https://github.com/tenstorrent/tt-forge-fe/pull/1408/files)?

- Since in-memory serialization is restricted by the 2GB Protobuf limit, why can't we use file-based loading instead(i.e., revert the changes from this [PR #1408](https://github.com/tenstorrent/tt-forge-fe/pull/1408/files))? or is there a better approach to handle this type of cases?

[...]

PR #1515, opened 2025-03-23
Add ONNX tests for BERT, DETR model

kamalrajkannan78:
### summary 

This PR adds ONNX tests for BERT, DETR model

## Logs

- [bert_onnx.log](https://github.com/user-attachments/files/19430194/mar23_bert_sen_gen_1.log)
- [detr_onnx.log](https://github.com/user-attachments/files/19430199/mar23_detr_onnx_c1.log)

PR #1569, opened 2025-03-27
Resolve serialization issue in large ONNX models

kamalrajkannan78:
### Ticket

- https://github.com/tenstorrent/tt-forge-fe/issues/1529

### Problem description

- ONNX ModelProto Exceeds 2GB Limit Causing Serialization Error in Forge Compilation

### What's changed

- For ONNX models larger than 2GB (models with external data), SerializeToString is no longer used.
- Instead, the ONNX file path must be provided to forge.OnnxModule.
- The wrapped module is then passed to forge.compile, where the ONNX path is extracted from the wrapped module during compilation.

### logs 

- [phi2_onnx_before_fix.log](https://github.com/user-attachments/files/19491291/mar25_phi2_onnx_using_opti.log)
- [phi2_onnx_after_fix.log](https://github.com/user-attachments/files/19527399/mar30_z_phi2_export_pr.log)


saiarthiraguram (approved):

forge/forge/compile.py
```diff
@@ -181,6 +181,7 @@ def compile_main(
     module: AnyModule,
     sample_inputs: List[torch.Tensor],
     module_name: Optional[str] = None,
+    onnx_path: Optional[str] = None,
```
nvukobratTT, line 184: Instead of adding the ONNX file path as an additional compile argument, can we link the ONNX path to the ONNX module (`OnnxModule `)? 

This way, we'll only have to check if ONNX file path exists within our OnnxModule instead of propagating it through whole compile flow.

forge/test/models/onnx/text/phi2/test_phi2.py
```diff
@@ -0,0 +1,101 @@
...
+    module_name = build_module_name(
+        framework=Framework.ONNX,
+        model="phi2",
+        variant=variant,
+        source=Source.HUGGINGFACE,
+        task=Task.CAUSAL_LM,
+    )
+
+    # Record model details
+    forge_property_recorder.record_group("priority")
```
nvukobratTT, line 34: Let's update it to "red" :))

forge/forge/compile.py
kamalrajkannan78: - we are not passing the result of `framework_model = forge.OnnxModule(module_name, onnx_model)` to forge.compile.
- we are passing result of `onnx_model = onnx.load(onnx_path)` to the forge.compile.
- Hence onnx_path is mandatory argument in forge.compile

forge/forge/module.py
```diff
@@ -386,9 +389,12 @@ def forward(self, *args, **kwargs):
         so.inter_op_num_threads = 2
         so.intra_op_num_threads = 2
 
-        module_bytes = self.module.SerializeToString()
+        if self.onnx_path is not None:
+            model = self.onnx_path
+        else:
+            model = self.module.SerializeToString()
```
sott0n, line 395: When a ValueError is raised due to an ONNX model over 2GB and onnx_path is not set, should we display a message suggesting that the user should provide onnx_path in forge.compile() ?

forge/forge/compile.py
nvukobratTT: I see. To make this compile flow cleaner, should we extend it to be able to ingest `forge.OnnxModule` as well? 

E.g. for most cases where models are smaller, we can use direct binary and ONNX module. However, in these cases we'll need to have some additional information. Therefore, I'm more for an option to use `forge.OnnxModule` with additional path attached, rather than having a special case on compile function level for it. 

Let me know your thoughts Kamal

nvukobratTT: You'll see that compile funciton is intended to be flexible to cover different cases. 

Also, we're already wrapping those into our modules. We should be able to extend support further and extract path from OnnxModule.

kamalrajkannan78: - I completely agree with your suggestions & thanks for the input. I reflected your suggestions 

   - forge.compile function can ingest forge.OnnxModule directly. 
   - ONNX path is now extracted from OnnxModule.

forge/forge/module.py
kamalrajkannan78: - @sott0n Thanks for the suggestion! As per @nvukobratTT's feedback [here](https://github.com/tenstorrent/tt-forge-fe/pull/1569#discussion_r2020191858), we are now passing `onnx_path` to `forge.OnnxModule`, and the wrapped model is then passed to forge.compile.

- As you suggested, the error message is now displayed alongside the original error for better clarity. For cases where the ONNX model over 2GB and onnx_path is not set, the following message will be shown:


```
E               RuntimeError: The model exceeds the 2GB limit and cannot be serialized directly. 
E               To proceed, please provide an 'onnx_path' to load the model
E               Original error: Message onnx.ModelProto exceeds maximum protobuf size of 2GB: 11119721893
```

- log - [phi2_export_when_onnx_path_is_not_set.log](https://github.com/user-attachments/files/19527375/mar30_z_phi2_export_m2.log)

sott0n (approved):

saiarthiraguram (approved):

forge/forge/module.py
```diff
@@ -361,20 +361,23 @@ class OnnxModule(Module):
...
+    def __init__(self, name: str, module: onnx.onnx_ml_pb2.ModelProto, onnx_path: Optional[str] = None):
         """
         Create Onnx module wrapper.
 
         Parameters
         ----------
         module: onnx.onnx_ml_pb2.ModelProto
             onnx module
+        onnx_path: Optional[str]
+        Path to the ONNX model file. Used to directly load the model when its size exceeds 2GB (with external data).
```
nvukobratTT, line 373: Let's indent this line

forge/forge/module.py
```diff
@@ -386,9 +389,20 @@ def forward(self, *args, **kwargs):
...
+            try:
+                model = self.module.SerializeToString()
+            except Exception as e:
+                raise RuntimeError(
+                    "The model exceeds the 2GB limit and cannot be serialized directly. \n"
+                    "To proceed, please provide an 'onnx_path' to load the model\n"
+                    f"Original error: {e}"
+                )
+        else:
+            model = self.onnx_path
```
nvukobratTT, line 402: We should check if model path is provided, assert if not.

forge/forge/tvm_calls/forge_compile.py
```diff
@@ -225,6 +225,7 @@ def compile_tvm_graph(
         onnx_inputs = [x.detach().numpy() for x in inputs]
         json_graphs, _ = compile_onnx_for_forge(
             module,
+            path,
```
nvukobratTT, line 228: Let's have a cleaner name (e.g. serialized_file_path or something similar)

forge/forge/tvm_to_python.py
```diff
@@ -2134,7 +2134,9 @@ def compile_tvm_to_python(
 
     # Path is needed for TFLite model verification against TVM compile.
     path = None
-    if isinstance(framework_mod, TFLiteModule):
+    if isinstance(framework_mod, OnnxModule):
+        path = framework_mod.onnx_path
```
nvukobratTT, line 2138: Cleaner name instead of pure 'path'

forge/test/models/onnx/text/phi2/test_phi2.py
```diff
@@ -0,0 +1,100 @@
...
+from test.models.utils import Framework, Source, Task, build_module_name
+
+from test.utils import download_model
+import subprocess
+
+variants = ["microsoft/phi-2"]
+
+
+@pytest.mark.nightly
+@pytest.mark.xfail(reason="AssertionError: Encountered unsupported op node type: scatter_nd, on device: tt")
```
nvukobratTT, line 19: Let's remove xfail reasons for models, they're hard to maintain. 

We'll keep them for op tests only (generated ones)

forge/test/models/onnx/text/phi2/test_phi2.py
```diff
@@ -0,0 +1,100 @@
...
+        return_tensors="pt",
+        max_length=256,
+        pad_to_max_length=True,
+        truncation=True,
+    )
+    input_ids = inputs["input_ids"]
+    attn_mask = inputs["attention_mask"]
+
+    # Export model to ONNX
+    onnx_path = f"{tmp_path}/model.onnx"
```
nvukobratTT, line 57: I assume this export will be mostly standardized across multiple models. Should we put it as some common/util function?

forge/forge/module.py
kamalrajkannan78: else block will execute only when onnx_path is not None, so the assertion may not be necessary?

forge/forge/tvm_calls/forge_compile.py
kamalrajkannan78: - I am using the existing path argument from the compile_tvm_graph function, which is also used by the TFLite framework.
- Renaming path to `onnx_file_path` or `serialized_file_path` would require updating all occurrences of path, leading to additional changes.
- To avoid this, we could introduce an extra line to assign path to a new variable & pass it to compile_onnx_for_forge like below  :

```
onnx_file_path = path

json_graphs, _ = compile_onnx_for_forge(
    module,
    onnx_file_path,
    *onnx_inputs,
    graph_name=graph_name,
    compiler_cfg=compiler_cfg,
    verify_cfg=verify_cfg,
    forge_property_handler=forge_property_handler,
)

```
- Let me know if you still prefer renaming it, and I can update it accordingly.

forge/forge/tvm_to_python.py
kamalrajkannan78: - This `path` variable is used by ONNX & TFLite framework. we need a unified name to pass it as single argument to load_tvm_graph. that's why I preferred 'path'

forge/test/models/onnx/text/phi2/test_phi2.py
kamalrajkannan78: May I put it in `forge/test/utils.py`?

nvukobratTT (approved):
Minor changes left, I'm approving to unblock the merge of this PR once these are resolved

forge/test/models/onnx/text/phi2/test_phi2.py
kamalrajkannan78: As @vkovinicTT suggested [here](https://github.com/tenstorrent/tt-forge-fe/pull/1515#discussion_r2022890891), It will better if we move this export code in  `forge/test/models/models_utils.py `instead of `forge/test/models/utils.py`. 
what do you think @nvukobratTT ?

vkovinicTT: > As @vkovinicTT suggested [here](https://github.com/tenstorrent/tt-forge-fe/pull/1515#discussion_r2022890891), It will better if we move this export code in `forge/test/models/models_utils.py `instead of `forge/test/models/utils.py`. what do you think @nvukobratTT ?

I commented on that PR, maybe `forge/test/models/utils.py` is better name. Sorry for the confusion 😕

kamalrajkannan78: @vkovinicTT may I move those helper functions from `forge/test/models/models_utils.py` to f`orge/test/models/utils.py`  And I noticed forge/test/models/utils.py already exists in latest main with [contents ](https://github.com/tenstorrent/tt-forge-fe/blob/main/forge/test/models/utils.py)related to build_module_name. is it ok put this helper functions in the same file ?

kamalrajkannan78: @vkovinicTT , I Moved the contents from `forge/test/models/models_utils.py` to `forge/test/models/utils.py` now & added new fucntion in same file to construct export command using optimum cli. Let me know, if additional changes needed.

kamalrajkannan78: I am merging this PR now to unblock onnx bringup for large onnx models. if additional changes related to utils sctipt path is needed , will handle it in other/separate PR

vkovinicTT: Yep I missed that `forge/test/models/utils.py` already exists, we can move things related to model inputs/outputs preprocessing into `forge/test/models/models_utils.py`, but leave things like building of the module name in the `forge/test/models/utils.py`.

PR #1572, opened 2025-03-28
Add onnx test for phi1

meenakshiramanathan1:
This PR adds onnx test for phi1


forge/test/models/onnx/text/phi1/test_phi1.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # Record model details
+    module_name = build_module_name(
+        framework=Framework.ONNX,
+        model="phi1",
+        variant=variant,
+        source=Source.HUGGINGFACE,
+        task=Task.CAUSAL_LM,
+    )
+    forge_property_recorder.record_group("priority")
```
vkovinicTT, line 35: Just change this to `red` since we are transitioning groups from `priority` -> `red`.

PR #1574, opened 2025-03-28
Enable ONNX Models

nvukobratTT:
Enable ONNX models to run as part of CI.

Majority of the vision tests (all except ResNet) require further changes before we can enable them (full restructuring). Pushing this change to enable test runs for few text models beside ResNet, as well as enable run for upcoming ONNX models for Red Team.

### DIFF
--- a/forge/test/models/onnx/vision/yolo/test_yolo_v5.py
+++ b/forge/test/models/onnx/vision/yolo/test_yolo_v5.py
@@ -6,11 +6,9 @@ import requests
 import pytest
 import torch
 from PIL import Image
-from pathlib import Path
 import cv2
 import numpy as np
-from yolov5.utils.dataloaders import exif_transpose, letterbox
-import onnx, pytest
+import onnx
 
 # TODO: These are old forge, we should update them to the currently version.
 # import forge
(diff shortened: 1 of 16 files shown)


nvukobratTT:
@sott0n I'm having the same issue, but I'm unable to reproduce it locally for some reason. I'll also dig in further to figure out where the issue is

nvukobratTT:
@sott0n few doesen runs and experiments, and I managed to figure out the root cause...

In sum, Yolo v5 imports had conflicts with existing libraries, causing some of the core Python libraries to produce issues, as you can see. I removed them, however, the majority of vision models require full restructure before we can enable them. 

However, this change should be good to be merged now to enable current ONNX workloads to run, and to unblock further push for ONNX as well.

forge/test/models/onnx/vision/yolo/test_yolo_v5.py
```diff
@@ -6,11 +6,9 @@
 import pytest
 import torch
 from PIL import Image
-from pathlib import Path
 import cv2
 import numpy as np
-from yolov5.utils.dataloaders import exif_transpose, letterbox
```
sott0n, line 12: Just to clarify my understanding — are you saying the issue was caused by the yolov5 library? From what I can tell using git grep, it looks like this library is only used in the ONNX test.
Anyway, everything's working well now! Once this patch is merged, I’ll go ahead and close my PR.

sott0n (approved):

forge/test/models/onnx/vision/yolo/test_yolo_v5.py
nvukobratTT: It's prbably the best. Before removing it, I want to take a more detailed look what happened. Just to prevent it from happening in the future.

PR #1582, opened 2025-03-28
[OP] Index - from decomposition to direct mapping of TTIR Index OP

vkovinicTT:
## What's Changed

- **Index Operation Support**: Replaced decomposition of `Index` OP with direct mapping to `MLIR`'s native `Index` OP
- **Fixed Branching Logic**: Corrected a bug in `compare_with_golden` where improper conditional structure (two separate `if` statements instead of `if/elif`) allowed execution of both paths
- **Memory Layout Handling**: Updated `calculate_dissimilarity` to use `reshape(-1)` instead of `view(-1)` for tensor flattening. This is critical because:
  - `view(-1)` requires tensors to have contiguous memory layout
  - Certain tensor manipulation operations create non-contiguous memory layouts
  - `view(-1)` throws runtime errors when applied to non-contiguous tensors
  - `reshape(-1)` intelligently handles both cases - reusing the same memory when contiguous and creating a properly organized copy when non-contiguous
- **Test Improvements**: Removed `xfail` mark from index tests as they now pass successfully
- **Code Cleanup**: Eliminated redundant code that was likely legacy from pybuda


forge/test/models_ops/test_index.py
```diff
@@ -1917,7 +1917,6 @@ def ids_func(param):
                 "op_params": {"dim": "-2", "start": "0", "stop": "1", "stride": "1"},
```
nvukobratTT, line 1917: There is no need to remove these by hand. 

It's okay for this CI to fail with xpass. @chandrasekaranpradeep has a script that automatically removes/edits these xfails based on the model ops nightly results. There is not bad side or removing them here as well, but just an advice if there are many of these failures across different ops :))

PR #1668, opened 2025-04-01
- Adding training intermediate verification.

### LINKED ISSUE #1667: Enable golden verification for training
Currently we have golden intermediate verification for forward pass. 
For successful debugging of training graph (forward, backward, and optimizer) we need following golden comparisons:

intermediate gradients (gradients with respect to outputs of intermediate operations),
parameter gradients,
input (activations) gradients (if inputs require grad),
parameter updates (if optimizer is on device)

dgolubovicTT:
### Ticket
Closes #1667 

### What's changed
- Training verification includes comparison of:
		intermediate gradients (gradients with respect to outputs of intermediate operations),
		parameter gradients,
		input (activations) gradients (if inputs require grad),
		parameter updates (if optimizer is on device)
- Adding a small end to end test with verification included.
- Cleanup redundant code. Remove redundant calls to calculate_grads and _run_pytorch_backward. Calling backward once for output tensor is enough to have golden gradients for comparison

Training verification can be performed in compile stages after autograd pass (since before we don't have backward graph). Graph is evaled op by op (forward and backward) and corresponding gradients are compared to the result of pytorch backward.

Few implementation details not to miss:
- Previously, compile context attributes `context.outputs `and `context.parameter_dict` were not from the same computational graph. Namely, context.outputs was output of ForgeModule forward, while context.parameter_dict was from PytorchModule. So essentially, we took PytorchModule extracted parameters, then translated to ForgeModule and took outputs from forward. This is why outputs.backward() didn't produce change in parameter gradients… 
Solution: use ForgeModule to extract parameter_dict

- From now, we pass optimizer to do_verify function. Optimizer parameters are inputs to the graph, so we use them to calculate updates of trainable parameters.

### DIFF
--- a/forge/csrc/graph_lib/python_bindings.cpp
+++ b/forge/csrc/graph_lib/python_bindings.cpp
@@ -50,7 +50,8 @@ eval_graph(
     float pcc,
     std::string const &dump_tensors_path,
     bool allow_modified_shapes,
-    bool return_intermediates);
+    bool return_intermediates,
+    std::optional<py::object> optimizer = std::nullopt);
 
 py::object get_constant_input_value(graphlib::Node *node, bool is_forge);
 py::object eval_input(
@@ -164,6 +165,7 @@ void GraphModule(py::module &m_graph)
         .def("get_ordered_input_tile_dims", &Graph::get_ordered_input_tile_dims)
         .def("get_ordered_parameter_tile_dims", &Graph::get_ordered_parameter_tile_dims)
         .def("get_ordered_constant_tile_dims", &Graph::get_ordered_constant_tile_dims)
+        .def("contains_bwd_nodes", &Graph::contains_bwd_nodes)
         .def(
             "get_input_runtime_tensor_transforms",
             [](Graph *graph) -> std::vector<graphlib::RuntimeTensorTransform>
@@ -628,7 +630,8 @@ void GraphModule(py::module &m_graph)
            const std::vector<py::object> &losses,
            const std::vector<py::object> &targets,
            std::string const &dump_tensors_path,
-           bool allow_modified_shapes)
+           bool allow_modified_shapes,
+           std::optional<py::object> optimizer)
         {
             auto ret = eval_graph(
                 graph,
@@ -641,7 +644,8 @@ void GraphModule(py::module &m_graph)
                 pcc,
                 dump_tensors_path,
                 allow_modified_shapes,
-                false);
+                false,
+                optimizer.value_or(py::none()));
             return std::make_tuple(std::get<0>(ret), std::get<1>(ret), std::get<2>(ret), std::get<3>(ret));
         },
         py::arg("graph"),
@@ -653,7 +657,8 @@ void GraphModule(py::module &m_graph)
         py::arg("losses") = std::vector<py::object>(),
         py::arg("targets") = std::vector<py::object>(),
         py::arg("dump_tensors_path") = "",
-        py::arg("allow_modified_shapes") = false);
+        py::arg("allow_modified_shapes") = false,
+        py::arg("optimizer") = py::none());
 
     m_graph.def(
         "remove_node",
@@ -774,16 +779,6 @@ py::object eval_relu(py::object tensor, graphlib::OpType type)
         tensor = eval_op(relu, inputs, graphlib::IRLevel::IR_TT_FORGE);
     }
     return tensor;
-}
-
-py::object eval_t_streaming_tms(py::object tensor, graphlib::Graph *graph, graphlib::Node *node, std::string const &dir)
-{
-    return tensor;
-}
-
-py::object eval_t_streaming_tms(py::object tensor, graphlib::Graph *graph, graphlib::Node *node)
-{
-    return eval_t_streaming_tms(tensor, graph, node, "Redo");
 }
 
 py::object eval_golden_transforms(graphlib::Node *node, py::object tensor, bool eval_for_output = false)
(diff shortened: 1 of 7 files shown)


dgolubovicTT:
Here is how it looks in log:
intermediate_training_verification

forge/forge/compile.py
```diff
@@ -302,6 +302,12 @@ def forge_compile_from_context(context: CompileContext) -> CompiledModel:
 
         # Execute the current stage.
         next_stage = stage_to_func[current_stage](context)
+        in_training = context.training and current_stage.value >= CompileDepth.AUTOGRAD.value
+
+        #  Set optimizer parameters from module, because we need optimizer parameters in intermediate verification.
+        if in_training and context.optimizer_on_device():
+            module_params = context.modules[0].get_parameters()
+            context.optimizer.set_parameters_to_optimize(module_params)
```
pilkicTT, line 310: this should be done once in `run_autograd_pass`, not here

forge/forge/compile.py
```diff
@@ -316,7 +322,6 @@ def forge_compile_from_context(context: CompileContext) -> CompiledModel:
         if (
             verify_cfg.verify_all or (verify_cfg.verify_last and should_early_stop_compilation) or should_verify
         ) and can_verify:
-            in_training = context.compiler_cfg.enable_training and current_stage.value >= CompileDepth.AUTOGRAD.value
             do_verify(
                 current_stage.name.lower(),
                 in_training,
```
pilkicTT, line 327: it feels like the `do_verify()` should figure out this by itself... it has a graph, so it can check if we have bwd subgraph in there, etc...

forge/forge/verify/verify.py
```diff
@@ -37,33 +37,21 @@
...
+def _run_pytorch_backward(outputs, losses):
     retain_graph = True
     for i, o in enumerate(outputs):
         if o.requires_grad:
-            if device.loss_module is None:
-                loss = narrow_forge_tensor_to_pytorch(losses[i], o.value().shape)
-                o.value().backward(loss, retain_graph=retain_graph)
-            else:
-                o.value().backward(retain_graph=True)  # this is loss
+            loss = narrow_forge_tensor_to_pytorch(losses[i], o.value().shape)
```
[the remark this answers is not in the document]
pilkicTT: we can do this in a separate PR, of course...

pilkicTT (changes requested):

forge/forge/verify/verify.py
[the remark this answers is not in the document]
dgolubovicTT: Yeah, I have a list of things related to pybuda that should be cleaned as a follow up here. As we talked, I opted to do it in a separate PR. Will create an issue

forge/forge/compile.py
```diff
@@ -168,11 +168,11 @@ def calculate_grads(outputs: Tuple[Tensor, ...], intermediate_golden_tensors: Di
 
     if not losses or run_backward:
```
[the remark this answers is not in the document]
dgolubovicTT: Good catch, we don't need it now :)

forge/forge/compile.py
dgolubovicTT: True

forge/csrc/graph_lib/python_bindings.cpp
```diff
@@ -1389,60 +1388,59 @@ std::unordered_map<std::string, py::object> get_graph_input_mapping(
...
+            // E.g. for 'l1.weight' trainable parameter, the optimizer parameters could be learning rate and momentum.
+            // If that was the only trainable parameter in our example, the optimizer_params dictionary would look like
+            // this:
+            // {
+            // ('l1.weight', 'lr'): 0.01,
+            // ('l1.weight', 'momentum'): 0.9,
+            // }
+            // So key to this dictionary is a tuple of parameter name and optimizer parameter name.
+
+            if (optimizer_params.is_none())
```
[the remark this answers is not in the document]
dgolubovicTT: You are right, inside this if `input->is_optimizer_parameter()` `optimizer_params.is_none()` is not a valid state. I can just add an assert

forge/forge/compile.py
dgolubovicTT: There is mapping for each node Id to SubgraphId `node_id_to_subgraph_id_`. However, I don't see which subgraph_id is Backward. 
We have 
```
enum class SubgraphType
{
    Forward,
    Backward,
    Loss,
    Optimizer,
};
```
but seems we are not using it. I figured I could use `nodes_by_subgraph` if I knew the Id of Backward graph, but I don't. Can you give some insight?

pilkicTT: i think you should be able to use `graph->contains_bwd_nodes()` - if it does, then it's in training...

forge/forge/verify/verify.py
```diff
@@ -37,33 +37,21 @@
...
+            o.value().backward(loss, retain_graph=retain_graph)
```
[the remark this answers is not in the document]
pilkicTT: hm, i am still not sure, because whenever you call `backward()` on a tensor, it should create its own graph, as far as i understand...

maybe we should test this scenario?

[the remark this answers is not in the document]
dgolubovicTT: Here is a pytorch example:

```
import torch

x = torch.tensor(2.0, requires_grad=True)
y = x * x

# First backward pass
y.backward()  # OK

# Second backward pass on same output
try:
    y.backward() 
except RuntimeError as e:
    print("Expected error:", e)
```

Here is error when trying to call .backward second time:

`Expected error: Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed). Saved intermediate values of the graph are freed when you call .backward() or autograd.grad(). Specify retain_graph=True if you need to backward through the graph a second time or if you need to access saved tensors after calling backward.`

forge/forge/compile.py
dgolubovicTT: Done, tnx

forge/forge/verify/verify.py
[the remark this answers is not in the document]
pilkicTT: but this is calling `backward()` on the same tensor... you are mentioning multiple outputs, i.e. different tensors, so i was wondering if it is necessary in this case?

forge/test/mlir/test_golden_verification.py
```diff
@@ -74,3 +77,45 @@ def forward(self, x):
...
+        torch.rand(shape2, requires_grad=train),
+    ]
+
+    framework_model = Matmul()
+    framework_model.eval() if not train else framework_model.train()
+    verify_cfg = DepricatedVerifyConfig()
+    # Use verify_all to verify all stages of compilation.
+    # verify_cfg.verify_all = True
+    # Use this stages_for_intermediate_verification to specify stage in which you want to verify.
+    verify_cfg.stages_for_intermediate_verification = {CompileDepth.AUTOGRAD}
```
pilkicTT, line 112: can we have a test which will test golden verification on each stage individually and also when `verify_all=True`?

pilkicTT (approved):

forge/test/mlir/test_golden_verification.py
dgolubovicTT: done

PR #1675, opened 2025-04-02
Enable torch loss to be compiled as loss

ndrakulicTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist


forge/csrc/passes/python_bindings.cpp
ndrakulicTT, line 1: This file is not needed per say to enable Torch losses but is needed to decompose log_softmax that is produced by tvm

forge/forge/op/eval/forge/nn.py
ndrakulicTT, line 1: fix for log_softmax decomposition

nvukobratTT:
Is this one stoll active?

PR #1680, opened 2025-04-02
[build] add option for skipping tt-mlir env build

pilkicTT:
Introducing `-DTTFORGE_SKIP_BUILD_TTMLIR_ENV` cmake option, when building `tt-forge-fe` environment.

By default, as part of building our environment, we also build `tt-mlir` env as well. This can be problematic for developers which want to work on multiple active repos (either `tt-mlir` or other FEs) and already have the `tt-mlir` toolchain built - because our env build will (by default) rebuild the `tt-mlir` toolchain from scratch, even if it is not necessary.

In those cases, the introduced option can be used, so that you can point to the already built `tt-mlir` toolchain dir (via `TTMLIR_TOOLCHAIN_DIR` env variable) and skip rebuilding it when building `tt-forge-fe` env.

```bash
cmake -B env/build -DTTFORGE_SKIP_BUILD_TTMLIR_ENV=ON env
```

NOTE: if you skip the `tt-mlir` env build, you need to make sure that your already existing `tt-mlir` env (toolchain) version is compatible with `tt-forge-fe`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.20.0)
+cmake_minimum_required(VERSION 3.24.0)
 
 project(ttforge LANGUAGES CXX)
 
--- a/docs/src/build.md
+++ b/docs/src/build.md
@@ -35,9 +35,9 @@ python3 --version
 ```
 
 ## Build environment
-This is one off step to build the toolchain and create virtual environment for tt-forge. Generally you need to run this step only once, unless you want to update the toolchain (LLVM).
+This is one off step to build the toolchain and create virtual environment for `tt-forge-fe`.Generally, you need to run this step only once, unless you want to update the toolchain. Since `tt-forge-fe` is using `tt-mlir`, this step also builds the `tt-mlir` environment (toolchain).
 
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.20.0)
+cmake_minimum_required(VERSION 3.24.0)
```
nvukobratTT, line 1: We should update our requirements in doc as well:
- https://docs.tenstorrent.com/tt-forge-fe/build.html#prerequisites

nvukobratTT (approved):
Looks good! One minor comment (approving nonetheless).

dgolubovicTT (approved):

CMakeLists.txt
pilkicTT: yeah, I am still looking into how to handle cmake in our build workflow... maybe adding a separate script for installing everything that is required for build.

anyhow, I'll update the relevant docs with a follow up change...

nvukobratTT: Yup, a script sounds good. However, it should handle the cases when the user isn't building on top of vanilla Ubuntu (which will often be the case). 

But maybe we can simplify it as much as possible. E.g., if the user has a cmake installed, and it's on an older version, we can notify him to remove it and re-run the script or manually upgrade the version to the required one. Just to make our life easier managing different edge cases..

PR #1699, opened 2025-04-04
Support Jax/Flax

sott0n:
### Ticket
N/A

### Problem description
The JAX support implementation is currently inactive (commented out), and it is outdated and contains known bugs.

### What's changed
This patch adds support for flax.linen.Module as the JAX frontend, along with transformers.FlaxPreTrainedModel. Most basic operations now pass, but a few tests are currently skipped due to ttnn runtime errors. Additionally, the ResNet-50 test from transformers has not yet fully passed due to unresolved issues. I’ve added TODO comments and documented the reasons for each skipped test. I’m actively working on resolving these issues.

### Checklist


forge/test/mlir/test_ops_jax.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+@pytest.mark.push
+def test_add(forge_property_recorder):
+    class Add(nn.Module):
+        @nn.compact
+        def __call__(self, x, y):
+            return x + y
+
+    key = random.PRNGKey(0)
```
nvukobratTT, line 22: As this key gen is pretty repetitive, let's create a util/common function for Jax that will generate and return random inputs. E.g. give arbitrary number of shapes, data formats, etc. and return generated inputs.

nvukobratTT: Potentially, to clean up the tests, we can have a function that will also bind variables to the framework model (as it's a pretty repetitive action as well. E.g., return generated inputs and bindend model. In sum think about it and let's see can we make these tests look a bit clener :))

forge/test/models/jax/vision/resnet/test_resnet.py
```diff
@@ -0,0 +1,56 @@
...
+    forge_property_recorder.record_model_name(module_name)
+
+    framework_model = FlaxResNetForImageClassification.from_pretrained(variant)
+
+    key = random.PRNGKey(0)
+    input_sample = [random.normal(key, shape=(1, 3, 224, 224))]
+
+    # TODO: The constant folding pass causes the graph to break around the multiply operation at the end.
+    # Therefore, this pass is skipped until the issue is resolved.
+    os.environ["FORGE_DISABLE_CONSTANT_FOLDING"] = "1"
```
nvukobratTT, line 45: Can you open up issues for this and add it as a comment here? 

Before saying that this model is functional, we should fix the bug once TTNN issues is resolved as well

forge/test/models/jax/vision/resnet/test_resnet.py
```diff
@@ -0,0 +1,56 @@
...
+    compiled_model = forge.compile(
+        framework_model, sample_inputs=input_sample, forge_property_handler=forge_property_recorder
+    )
+
+    verify(
+        input_sample,
+        framework_model,
+        compiled_model,
+        VerifyConfig(value_checker=AutomaticValueChecker(pcc=0.95)),
```
nvukobratTT, line 54: Let's target 0.99 for PCC. If you hit the same accuracy issues once the TTNN problem is resolved, then let's reduce the expected precision

forge/forge/module.py
```diff
@@ -579,29 +581,48 @@ def __init__(self, name: str, module):
...
-        return []  # TODO
+        parameters = []
+
+        def _set_parameters(d, prefix=""):
+            for key, value in d.items():
+                if isinstance(value, (dict, flax.core.frozen_dict.FrozenDict)):
+                    _set_parameters(value, f"{prefix}{key}." if prefix else key + ".")
+                else:
+                    param_tensor = to_pt_tensor(value)
+                    parameters.append(Parameter(param_tensor, requires_grad=True, name=f"{prefix}{key}"))
```
nvukobratTT, line 615: Do we want to set `requires_grad=True` by default? 

I think should be inferred from the Jax parameters or set to `False` by default, as default is inference run unless training is explicitly requested before compiling the model.

forge/forge/tvm_calls/forge_utils.py
```diff
@@ -263,69 +266,67 @@ def construct_tvm_ir(framework: str, model, tvm_mod, params, compiler_cfg: Compi
...
+        param_name_lookup = {}
+        non_weight_params = {}
+        model_params = {}
+
+        if hasattr(model, "params"):  # For Tranformer's FlaxPreTrainedModel
+            model_params = model.params
+        elif hasattr(model, "variables") and "params" in model.variables:
+            model_params = model.variables["params"]._dict
+
+        model_params = flatten(model_params)
```
nvukobratTT, line 291: Let's document each main step here. E.g.
- Params flattening
- Param mapping
- TVM constant prop option

Just to have clearly outlined sections that are easier to follow with the main steps explained.

forge/forge/module.py
sott0n: Thanks for the catch! We should set requires_grad=False by default. In JAX, parameters don’t carry a train/eval state since it’s a functional design—train vs. eval is typically controlled by an explicit argument to the function. So if we want to support training mode, we’d need to include a train flag to the sample_input at forge.compile and trace it.

forge/test/models/jax/vision/resnet/test_resnet.py
sott0n: opend at https://github.com/tenstorrent/tt-forge-fe/issues/1709

forge/forge/module.py
nvukobratTT: I see. Can you create a feature request to track this as well? 

Once we stabilize Jax's path for inference, we want to have it in functional for training as well. However, this will require more exploration as Jax has its autograd, and we likely have to model it differently compared to other frameworks (possibly skip our autograd). In sum, let's track it through separate issues :))

forge/forge/tvm_calls/forge_compile.py
```diff
@@ -956,21 +958,45 @@ def get_frozen_graph_for_large_jax_model(
...
+                {"params": params, "batch_stats": batch_stats}, *input
+            )
+
+        # Convert model from Jax to TensorFlow
         tf_fn = jax2tf.convert(predict_fn, enable_xla=compiler_cfg.enable_xla_jax_convert)
 
-        params_vars = tf.nest.map_structure(lambda param: tf.Variable(param, trainable=True), params)
-        tf_func = tf.function(lambda inputs: tf_fn(params_vars, inputs), autograph=False, jit_compile=True)
+        # Convert parameters TF variables.
+        tf_params = tf.nest.map_structure(lambda param: tf.Variable(param, trainable=True), params)
```
nvukobratTT, line 995: Should `trainable=True` be set by default?

nvukobratTT (approved):
Looks good! One minor comment :))

forge/forge/module.py
```diff
@@ -981,11 +1002,29 @@ def wrap_module(module, name: str) -> Module:
...
         raise RuntimeError("Unsupported module type: " + str(type(module)))
 
 
+def is_supported_module(module):
+    """
+    Check that the module is supported by Forge.
+    """
+    if isinstance(module, AnyModule):
+        return True
+    elif hasattr(module, "module") and isinstance(module.module, AnyModule):
```
pilkicTT, line 1020: Can you explain why this is needed (it's probably related to he pretrained flax models, but i don't have any context on them)? Generally, I would like to avoid these kinds of subtle complications whenever possible.

pilkicTT (changes requested):

forge/forge/tvm_calls/forge_compile.py
sott0n: It should be set to True during training, but during inference, it is treated as a fixed value. So I believe trainable=True is the correct setting.

forge/forge/module.py
sott0n: Yes, this change is to support FlaxPreTrainedModel. In order to determine whether a model is a flax.linen.Module, we need to inspect its internal module, so I added a condition to check the internal module.
One way to simplify this might be to add FlaxPreTrainedModel to AnyModule, but do you have any better ideas?

nvukobratTT:
@sott0n one more note, did you tried to upgrade Jax to the latest version as we're bringing it up from the scratch once again?

sott0n:
@nvukobratTT 
> one more note, did you tried to upgrade Jax to the latest version as we're bringing it up from the scratch once again?

I tested this change with `jax==0.4.13`, which is the version currently pinned in Forge.
Would you like me to try upgrading to the latest JAX version?

nvukobratTT:
> @nvukobratTT
> 
> > one more note, did you tried to upgrade Jax to the latest version as we're bringing it up from the scratch once again?
> 
> I tested this change with `jax==0.4.13`, which is the version currently pinned in Forge. Would you like me to try upgrading to the latest JAX version?

Yup! Let's do it. It'll be much easier to upgrade it to the latest one now that we don't have many models attached to it

env/core_requirements.txt
```diff
@@ -13,9 +13,9 @@ dill==0.3.5.1
 flatbuffers==23.5.26
 # This is needed to prevent AttributeError: module 'ml_dtypes' has no attribute 'float8_e4m3b11'
 ml-dtypes==0.2.0
-flax==0.6.0
-jax==0.4.13
-jaxlib==0.4.11
+flax==0.9.0
+jax==0.4.30
+jaxlib==0.4.30
```
sott0n, line 18: @nvukobratTT Bumped JAX version from 0.4.13 to 0.4.30.
Did not upgrade to the latest version (0.5.3) because it requires updates to several other dependencies: numpy, tensorflow, ml-dtypes, and flatbuffers, which could have a wide-ranging impact. Additionally, the latest version causes an error in tvm.relay tracing. It's better to create a new PR if we upgrade the version further.

forge/forge/module.py
sott0n: opened https://github.com/tenstorrent/tt-forge-fe/issues/1717

forge/forge/module.py
sott0n: @pilkicTT Simplified checking by adding FlaxPreTrainedModel to AnyModule instead of checking `module.module` at https://github.com/tenstorrent/tt-forge-fe/pull/1699/commits/5c7bfca199f4c205eadb4f0c95f90b200befeb3a.
If you have any better ideas, please let me know.

pilkicTT: thanks! i think it's slightly more straightforward know - you just need to look into `AnyModule` to see what is supported...

forge/test/models/jax/vision/resnet/test_resnet.py
```diff
@@ -0,0 +1,57 @@
...
+    os.environ["FORGE_DISABLE_CONSTANT_FOLDING"] = "1"
+    compiled_model = forge.compile(
+        framework_model, sample_inputs=input_sample, forge_property_handler=forge_property_recorder
+    )
+
+    verify(
+        input_sample,
+        framework_model,
+        compiled_model,
+        VerifyConfig(value_checker=AutomaticValueChecker(pcc=0.99)),
```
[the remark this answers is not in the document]
sott0n: removed it at https://github.com/tenstorrent/tt-forge-fe/pull/1699/commits/02126fbb89cec449590ebb2fb417ca0903a80c15

forge/test/mlir/test_ops_jax.py
```diff
@@ -0,0 +1,222 @@
...
+
+    compiled_model = forge.compile(
+        framework_model, sample_inputs=inputs, forge_property_handler=forge_property_recorder
+    )
+
+    verify(inputs, framework_model, compiled_model, forge_property_handler=forge_property_recorder)
+
+
+@pytest.mark.push
+@pytest.mark.skip(reason="unsupported broadcast operation in ttnn at runtime")
```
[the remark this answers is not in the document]
sott0n: updated at https://github.com/tenstorrent/tt-forge-fe/pull/1699/commits/02126fbb89cec449590ebb2fb417ca0903a80c15

Issue #1599, opened 2025-03-31
Single nightly pipeline for model runs


nvukobratTT:
### Summary
At the moment, we run models across different nightly pipelines:
- Passing models (On nightly)
- Benchmarks (On nightly)
- Failing models (On nightly xfail)

The main goal of this task is to unify them and make them flexible in terms of running specific parts of the pipeline. 

### Details
More precisely, we need a single nightly pipeline that will run all models' workloads (the one outlined above). Besides, we need the flexibility of running specific parts of the pipeline per need and on specific branches. E.g. running only benchmarking on the custom branch (not only main). 

Next to model nightlies, we have a few nightlies focused on model analysis, sweep testing, etc. It'll be good if those can be invoked once the models' nightly pipelines finish with execution. However, if we manually run only the nightly pipeline, the rest shouldn't be triggered as well. For example, model nighlty is scheduled and run at 12 AM, it finishes at 2 AM and triggers Sweeps nighlty, etc. 

Additional cases to consider:
1. Ability to run single and group of tests only.
    - Useful for cases when we hit regression and want to patch it. 
    - Especially when nightly grows to 6h runs. When we hit that closer to that, it'll be mandatory to be more granular when regressions occur while testing. 
1. Ability to skip build job (mostly useful for manual runs; check to skip the build if desired)
    - Useful in cases when we're doing test skips or moving into success/fail groups
    - Also, useful when there are only interpreter level changes
    - Additionally, it is great for performance testing when we're only updating model configurations, without any compiler changes

nvukobratTT:
cc @pilkicTT @dgolubovicTT @vkovinicTT

nvukobratTT:
@nsumrakTT I added additional cases to consider based on our discussion. Let me know if you have additional ideas that can further enhance our nightlies :)))

nsumrakTT:
Looks good, but I have to say this issue will be implemented in multiple steps because when designing new workflow, I can't test or run it until it hits main branch.
So, first we should have redesigned nightly, and after that, when it is in main, we can tune your needs for individual test runs and similar features.

nvukobratTT:
Sounds great! 💪

nvukobratTT:
If easier, feel free to open sub issues to easily track where we are :))

vbrkicTT:
In nightly sweeps tests there is one specific regarding test selection/filtering via parameters specified via env variables.

[...]

PR #1704, opened 2025-04-05
Nightly build refurbishment

nsumrakTT:
### Ticket
https://github.com/tenstorrent/tt-forge-fe/issues/1599

### Problem description
Running multiple nightly pipelines, scheduled to run at different schedules

### What's changed
Single nightly pipeline that run all models' workloads with flexibility of running specific parts of the pipeline per need and on specific branches.

### Checklist
Successful refurbished nightly build.
Successful PR, push builds.
tt-forge release run should pickup "fail-notify" job as success indicator.
After merged in main, subsequent PR will test-fix-tune running partial parts of nightly tests.

New On Nightly: https://github.com/tenstorrent/tt-forge-fe/actions/runs/14292203331


.github/workflows/build.yml
```diff
@@ -7,6 +7,16 @@ on:
         description: 'Git SHA of commit in tenstorrent/tt-mlir or branch name'
         required: false
         type: string
+      build:
+        description: 'Build type (Release or Debug)'
+        required: false
+        default: 'All'
+        type: choice
+        options:
+          - All
```
nvukobratTT, line 16: Can we add descriptions next to each option?

.github/workflows/build.yml
```diff
@@ -116,6 +136,8 @@ jobs:
 
 
   unit-test-build:
+    if: ${{ inputs.build == 'All' || inputs.build == 'Debug' }}
```
nvukobratTT, line 139: Unit tests should all pass for release build

.github/workflows/on-nightly.yml
```diff
@@ -6,36 +6,95 @@ on:
...
     needs:
       - docker-build
       - build
-    uses: ./.github/workflows/test.yml
+    uses: ./.github/workflows/test-sub.yml
     secrets: inherit
     with:
       test_mark: 'nightly and not xfail'
       test_group_cnt: 4
       test_group_ids: '[1,2,3,4]'
```
nvukobratTT, line 31: Should we use all resources? Potentially doing split into more groups?

nvukobratTT (approved):
Few minor comments, not blocker to merge and continue with further enhancements and testing

.github/workflows/build.yml
nsumrakTT: This is an option in build to build all, release only, debug only or none.
It is useful for feature to manually run tests without building anything or, building just release version.
Also, nightly is designed to run from beginning to end even if build fails (it will try to find last successful build with installation) and in the end its resolution is in "fail-notify" job where it decides is nightly successful or failed (and send slack notification).

.github/workflows/build.yml
nsumrakTT: Done

PR #1716, opened 2025-04-07
[CI] Removed xfail mark from passing models

vkovinicTT:
### What's changed
Removed xfail mark from passing models.


nvukobratTT (approved):
Nice! Are there some OOM issues should consider as well?

PR #1718, opened 2025-04-07
[Ci] Skipping models with OOM problems

vkovinicTT:
### What's changed
- skipping models with OOM problems
- removed `xfail` from passing models

### DIFF
--- a/forge/test/models/onnx/text/ministral/test_ministral.py
+++ b/forge/test/models/onnx/text/ministral/test_ministral.py
@@ -16,7 +16,7 @@ variants = ["ministral/Ministral-3b-instruct"]
 
 
 @pytest.mark.nightly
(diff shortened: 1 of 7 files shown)


forge/test/models/onnx/text/ministral/test_ministral.py
```diff
@@ -16,7 +16,7 @@
 
 
 @pytest.mark.nightly
-@pytest.mark.xfail
+@pytest.mark.skip(reason="Out of memory")
```
nvukobratTT, line 19: To be more precise, let's mention that these are transient failures on CI (when a bunch of other models are run). Just in case is someone tries to run them, and don't hit OOM issues, to have a notion why this test is still skipped

PR #1734, opened 2025-04-08
[verification] optimize memory usage vol. 2

### LINKED ISSUE #1733: [verification] further optimize memory usage when calculating pcc/atol
There are op sweep tests which are testing the ops with large tensors (>1GB). In those cases, the current methods of calculating atol/pcc are still using too much memory. Implement custom torch cpu kernels for calculating atol/pcc in-place.

pilkicTT:
Implements in-place torch cpu kernels for calculating pcc, maximum absolute difference between two tensors and all close. These kernels do not have additional memory overhead, i.e. there are no intermediate tensor allocations.

Modified verification/comparison functions (in `compare.py`) to use these kernels by default, whenever possible. There are cases where we fallback to previous implementation; for example when tensors have non-floating point data types, when they contain special values (NaNs/Infs). These cases should be infrequent, but we can expand on the kernels in the future, to cover these as well.

Unit tests for testing basic functionallity of implemented ops are added.

Issue #1502, Closes #1733


forge/forge/verify/compare.py
```diff
@@ -215,10 +233,16 @@ def compare_tensor_to_golden(
             if relative_atol is None:
                 relative_atol = 0.1
 
-            if torch.all(torch.isnan(golden)):
+            if verif.has_special_values(golden):
```
pilkicTT, line 236: TODO: Fix this

forge/csrc/verif/verif_ops.cpp
```diff
@@ -0,0 +1,514 @@
...
+        });
+
+    return iter.output().item<double>();
+}
+
+bool unsupported_dtypes(torch::Tensor& a)
+{
+    switch (a.scalar_type())
+    {
+        case torch::kFloat:
```
nvukobratTT, line 324: Which unsupported data formats were you hitting? 

I see k* DFs are used internally. Do you have more details about what else we can get in there?

forge/csrc/verif/verif_ops.cpp
```diff
@@ -0,0 +1,514 @@
...
+                    auto delta = (a - b).abs() - atol_vec - b.abs() * rtol_vec;
+                    return delta;
+                },
+                std::numeric_limits<scalar_t>::lowest());
+        });
+
+    return iter.output().item<double>() <= 0.0;
+}
+
+double cov_ij(
```
nvukobratTT, line 377: Can we add a high-level doc above this function? Just mentioning the purpose of it (e.g., calculating covariance)

forge/csrc/verif/verif_ops.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+    TT_ASSERT(a_flat.numel() == b_flat.numel(), "Input tensors must have the same number of elements");
+
+    if (unsupported_dtypes(a_flat) || unsupported_dtypes(b_flat))
+    {
+        // Fallback to torch impl.
+        return at::min(at::corrcoef(at::stack({a_flat, b_flat}))).item<double>();
+    }
+
+    auto iter = at::TensorIteratorConfig()
```
nvukobratTT, line 457: What is the purpose of this iterator?

forge/csrc/verif/verif_ops.cpp
```diff
@@ -0,0 +1,514 @@
...
+    //
+    // Finally, the resulting PCC value is taken as the minimum element of the PCC matrix.
+    //
+    auto cov = at::empty({4}, options);
+    cov[0] = cov_ij(a_flat, a_flat, a_mean, a_mean);
+    cov[1] = cov_ij(a_flat, b_flat, a_mean, b_mean);
+    cov[2] = cov_ij(b_flat, a_flat, b_mean, a_mean);
+    cov[3] = cov_ij(b_flat, b_flat, b_mean, b_mean);
+
+    auto std_a = std::sqrt(cov[0].item<double>());
```
nvukobratTT, line 497: Can this converge to 0 in low variance tensors?

nvukobratTT: Should we add `1e-12` to avoid division by zero?

forge/csrc/verif/verif_ops.cpp
pilkicTT: Thanks, removed it...

forge/forge/verify/compare.py
pilkicTT: Fixed this so that we call `calculate_atol` which handles special values (instead of handling it here).

forge/csrc/test/verif/test_ops.cpp
```diff
@@ -0,0 +1,135 @@
...
+{
+    auto dtype = GetParam();
+
+    torch::Tensor a = torch::tensor({1.0, 2.0, 3.0}, at::TensorOptions().dtype(dtype));
+    torch::Tensor b = torch::tensor({1.0, 2.0, 3.0}, at::TensorOptions().dtype(dtype));
+    EXPECT_TRUE(tt::calculate_tensor_pcc(a, b) > 0.99);
+
+    a = torch::tensor({1.0, 2.0, 3.0}, at::TensorOptions().dtype(dtype));
+    b = torch::tensor({0.0, 2.0, 3.0}, at::TensorOptions().dtype(dtype));
+    EXPECT_TRUE(tt::calculate_tensor_pcc(a, b) < 0.99);
```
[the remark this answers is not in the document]
pilkicTT: I've added some of the degenerate cases as unit tests. For the targeting of specific value (outside of the -1, 1, nans) i don't see much value - we are already testing the calculation on pretty large tensors, so it should be enough.

forge/test/test_verify.py
```diff
@@ -8,25 +8,29 @@
...
 
 from forge.verify.compare import calculate_pcc, calculate_or_estimate_pcc, calculate_atol
+import pytest
 
+import forge._C.verif as verif
 
-def test_pcc_calculation():
+
+@pytest.mark.push
+def test_pcc_estimation():
```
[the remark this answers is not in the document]
pilkicTT: These tests are mostly for verifying the high level functions (`calculate_atol`, `calculate_pcc`) which have additional handlings for special values Infs/NaNs, etc.

The main testing point for individual kernels are the added unit tests.

forge/csrc/verif/verif_ops.cpp
pilkicTT: I've modified slightly what we support. Initially, i was scoping out the infrequently used data types. Now we can handle all integer and floating point types (except bool) - they are all being tested in the unit tests.

As for what more there is in torch, there are some data types supporting complex numbers and also quantization stuff, from what i quickly glanced through.

nvukobratTT (approved):

Issue #1731, opened 2025-04-08
Port Model analysis doc generation to superset


chandrasekaranpradeep:
Report : https://docs.google.com/document/d/17iMyD8qWwOxfTEeGJ7SYkuKw5XhRR3WRqM467X7nkHg/edit?tab=t.0#heading=h.73w08m9orlbl

Based upon the discussion with respect to porting it to superset,

Record Below properties:
1. forge_op_name 
2. operand -> node_type, shape, dataformat, torch_dtype
3. args
4. model_names
5. refined_error_message -> For both failed and xfailed test

PR #1740, opened 2025-04-08
Record model analysis details for porting model analysis docs generation to the superset. 

chandrasekaranpradeep:
Fixes https://github.com/tenstorrent/tt-forge-fe/issues/1731

This PR helps for porting the model analysis documentation generation workflow to the superset dashboard by recording model analysis details such as **forge_op_name**, **operands details(nodetype, shape, dataformat, torch_dtype)**, **args** and **model names** where the respective operation configuration are present. In addition, it also records **refined error message** and **failure category** which will be useful for replicating model analysis documentation generation in superset.

1. ​The **ForgePropertyStore** dataclass now includes an **op_info** field to store model ops test details such as the operation name (forge_op_name), operand details, operation arguments, and model names.
2. Created a ForgePropertyHandler method(i.e `record_singleop_operands_info`) to extract operand details  such as **NodeType**(i.e Activation/Parameter/Constant), **Shape**, **tt-dataformat** and **torch dtype** from the Framework model(Similar to how we are constructing initial forge graph from the derived class of th eForgeModule)  for the models ops tests and record these details in `ForgePropertyStore` inside` tags -> op_info -> operands`
3. Enhanced the `forge_property_recorder` pytest fixture to capture failures from both standard `failing tests` and tests expected to fail (`xfailed`); the captured failure data is passed into the `extract_refine_error_message` function to yield a **refined error messag**e, which is then used to derive a `failure category` that facilitates grouping and comparing similar errors across tests. Created `compiler_exception_rules` for extracting the failure_category from refined error message.
4. In models ops tests generation,  modified the metadata names from `op_name` to `forge_op_name` and `op_params` to `args` and updated the `write_pytest_function` method in `ForgeWritter` to use forge property handler methods such as `record_forge_op_args`, `record_op_model_names` and `record_forge_op_name`.
5. Updated all the models ops test present in the `forge/test/models_ops` to record these operation details inside the tags->op_info by modifying the __call__ method with respective ForgePropertyHandler methods(i.e `record_forge_op_args`, `record_op_model_names` and `record_forge_op_name`).

Recorded model analysis data - [superset preview](https://superset.tenstorrent.com/superset/dashboard/105/?native_filters_key=pUGzdfa_RXF_Cvs9j8o8aIXsPc1Qmy97rJCJPzwRzS18cAcWj37KlrwuinI3-gwD)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -22,18 +22,57 @@ def record_test_timestamp(record_property):
 
 
 @pytest.fixture(scope="function")
-def forge_property_recorder(record_property):
+def forge_property_recorder(request, record_property):
+    """
+    Pytest fixture to initialize and manage a ForgePropertyHandler instance for recording test properties.
+
+    The fixture sets up a property store along with a handler configured with default parameters:
+      - Execution depth is set to 'CI_FAILURE'
+      - Execution stage is set to 'FAILED_BEFORE_FORGE_COMPILATION_INITIATION'
+
+    After the test runs, the fixture extracts any refined error message and failure category attached to the test item
+    (if available) and records these details. Finally, the properties are stored using the provided record_property
+    function regardless of the test outcome.
+
+    Yields:
+        An instance of ForgePropertyHandler.
+    """
+
+    # Create a property store that will hold all the properties recorded during test execution.
     forge_property_store = ForgePropertyStore()
 
+    # Create a handler that uses the property store; the handler is responsible for recording and managing property details.
     forge_property_handler = ForgePropertyHandler(forge_property_store)
 
     # Set CI_FAILURE as default execution depth and FAILED_BEFORE_FORGE_COMPILATION_INITIATION as default execution stage
     forge_property_handler.record_execution_depth(ExecutionDepth.CI_FAILURE)
     forge_property_handler.record_execution_stage(ExecutionStage.FAILED_BEFORE_FORGE_COMPILATION_INITIATION)
 
+    # Provide the handler instance to the test function so it can record properties during test execution.
     yield forge_property_handler
 
-    forge_property_handler.store_property(record_property)
+    try:
+        # Retrieve any refined error message that might have been set during the test execution
+        refined_error_message = getattr(request.node, "refined_error_message", None)
+
+        # Check if:
+        # 1. The refined error message exists.
+        # 2. The handler is configured to record single operation details (record_single_op_details flag is True).
+        # If either of these checks fail, exit without further recording.
+        if refined_error_message is None or not forge_property_handler.record_single_op_details:
+            return
+
+        # Record the refined error message in the handler's property store.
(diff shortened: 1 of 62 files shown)


conftest.py
```diff
@@ -33,9 +39,29 @@ def forge_property_recorder(record_property):
 
     yield forge_property_handler
 
+    report = getattr(request.node, "rep_call", None)
```
nvukobratTT, line 42: Let's simplify this with a clear statement of what isn't supported, instead of having nested if conditions. E.g.
```py
report = getattr(request.node, "rep_call", None)

if not report or report.when != "call":
    return

xfail = hasattr(report, "wasxfail")
is_expected_skip = report.skipped and xfail
is_unexpected_fail = report.failed and not xfail

if not (is_expected_skip or is_unexpected_fail):
    return

error_message = getattr(report, "longreprtext", None)
if not error_message:
    return

refined_error_message = extract_refined_error_message(error_message)
if refined_error_message is None or not forge_property_handler.record_single_op_details:
    return

forge_property_handler.record_refined_error_message(refined_error_message)

failure_category = extract_failure_category(refined_error_message)
if failure_category is not None:
    forge_property_handler.record_failure_category(failure_category)

```

setup.py
```diff
@@ -78,13 +78,14 @@ def build_forge(self, ext):
 
 # Find packages as before
 packages = [p for p in find_packages("forge") if not p.startswith("test")]
+packages.extend(["scripts", "scripts.model_analysis"])
```
nvukobratTT, line 81: Why are we packaging model analysis as part of the wheel?

nvukobratTT: I saw this
> Updated setup.py file to include the scripts and scripts.model_analysis in the packages list, and scripts in the package_dir mapping. This change addresses a ModuleNotFoundError encountered in conftest.py during CI tests, where extract_failure_category is imported from scripts.model_analysis - [CI issues](https://github.com/tenstorrent/tt-forge-fe/actions/runs/14331041833/job/40167456721).

But let's think of a different way. Packaging scripts shouldn't be part of the release wheel. It should contain only core compiler logic.

conftest.py
chandrasekaranpradeep: Updated as you mentioned

forge/forge/compile.py
```diff
@@ -704,6 +704,9 @@ def generate_initial_graph(context: CompileContext) -> CompileDepth:
             target_tensors=context.targets,
         )
 
+    if context.forge_property_handler is not None and context.forge_property_handler.record_single_op_details:
```
nvukobratTT, line 707: Should this be here? Can we localize it closer to the rest of the op analysis and generation?

chandrasekaranpradeep: Updated the PR to extract and record the operands details from the derived class of the ForgeModule, instead of recording it inside the compile function

nvukobratTT (approved):

PR #1752, opened 2025-04-09
[test] fix imports in yolov5

pilkicTT:
The pre-commit is failing for all the branches rebased on top of latest main due to the bad imports in this test.


nvukobratTT (approved):
This is my bad.. Thanks for fixing it ❤️

Issue #1516, opened 2025-03-24
Forge compilation on Blackhole


gfengTT:
## Summary

Make sure forge compiles on Blackhole.

The workflow on BH should be identical to that on WH, except a line of `export ARCH_NAME=blackhole` at the very beginning.

### 1. Making sure compilation works locally.

It seems a default WH soc descriptor is always used when lowering to mlir:

https://github.com/tenstorrent/tt-forge-fe/blob/67047bc30ea5c52450b31fe4ac55b66dc160cfed/forge/csrc/passes/lower_to_mlir.cpp#L154

https://github.com/tenstorrent/tt-mlir/blob/9fed4d8e29fb712802c148c1a011ed8c37c477dc/lib/Dialect/TT/IR/TTOpsTypes.cpp#L42

So were we really compiling for BH even when ARCH_NAME is properly set?

### 2. Add BH CI tests.

On branch `gfeng/blackhole-ci`. Need to add: add ARCH_NAME env var to [build step](https://github.com/tenstorrent/tt-forge-fe/blob/main/.github/workflows/build.yml).

### 3. Rerun BH tests.

When the steps above are done, re-run all the BH tests. For those that don't pass, either fix them or label them as xfail.

### 4. Developer experience enhancements.

[Need a better way to distinguish BH arch](https://github.com/tenstorrent/tt-forge-fe/pull/1352#discussion_r1986428976)

In addition, `env/activate` silently sets ARCH_NAME to wormhole_b0 if it was unset, which easily leads to errors. Something better be fixed.

jmcgrathTT:
Thanks, @gfengTT. I can add that and test that out.

I also saw this, which might be an issue; I see a hardcoded `arch` here for wormhole.

https://github.com/tenstorrent/tt-mlir/blob/9fed4d8e29fb712802c148c1a011ed8c37c477dc/lib/Dialect/TT/IR/TTOpsTypes.cpp#L114

In the same file, a similar function uses the `arch` var based on a case statement above.

Func call:
https://github.com/tenstorrent/tt-mlir/blob/9fed4d8e29fb712802c148c1a011ed8c37c477dc/lib/Dialect/TT/IR/TTOpsTypes.cpp#L316

Case statement:
https://github.com/tenstorrent/tt-mlir/blob/9fed4d8e29fb712802c148c1a011ed8c37c477dc/lib/Dialect/TT/IR/TTOpsTypes.cpp#L236-L247

PR #1783, opened 2025-04-14
DO NOT MERGE: workaround for blackhole compilation

gfengTT:
This allows the lowering to use a BH system descriptor.

### Ticket
https://github.com/tenstorrent/tt-forge-fe/issues/1516

### Problem description
We were always lowering on to a hard-coded WH system (https://github.com/tenstorrent/tt-mlir/blob/003c711842c5c757c143ef4f64d1fcc478d9e5da/lib/Dialect/TT/IR/TTOpsTypes.cpp#L41).

### Changes
Please first create a system descriptor by running `ttrt query --disable-eth-dispatch --save-artifact` in tt-mlir. (see https://github.com/tenstorrent/tt-mlir/issues/2632 if you run into any issues)

Then change the path in `lower_to_mlir.cpp`.

A more proper fix is needed. Maybe we should hardcode the BH values and add arch_name as an additional parameter to `mlir::tt::SystemDescAttr::getDefault()`?

### Checklist

### DIFF
--- a/forge/csrc/passes/lower_to_mlir.cpp
+++ b/forge/csrc/passes/lower_to_mlir.cpp
@@ -160,8 +160,9 @@ class MLIRGenerator
     mlir::ModuleOp emit_mlir(tt::ForgeGraphModule &module)
     {
         graphModule_ = mlir::ModuleOp::create(get_module_location(module), module.name());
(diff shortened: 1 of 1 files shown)


nvukobratTT:
@gfengTT, as we're working on this PR https://github.com/tenstorrent/tt-forge-fe/pull/2102, can we close this one?

PR #1805, opened 2025-04-16
Enable lower data format user override.

### LINKED ISSUE #1804: Enable user defined data format override
First step in supporting lower data formats in compiler is to be able to set lower data format from compiler config.

dgolubovicTT:
### Ticket
Closes #1804 

### Problem description
Optimizer needs a way to use lower precision data formats. Since TVM requires traced model to be in float32 (due to its numpy base), generated initial graph is in float32. To transfer this to bfloat16 (or some other lower data format) two things needed to be done:
1. cast all input tensors (Inputs, weights, constants) to lower data format.
2. convert output data format in whole graph

We used existing config `default_df_override`. By setting this to some data format, compiler will cast the input tensors and operations to that lower data format. Therefore enabling optimizer to have smaller memory footprint and enable higher batch size. 

### What's changed
- Set user-defined data format override in post-initial graph pass.
  Insert cast nodes on all input edges to ensure that constants, buffers, and parameters are in the lowered dtype.
- Move TVM module generation inside the block where the traced model is in float32, since that is the format supported by TVM (due to its NumPy base).
- Extend ConvertEmulatedDtypes to convert model buffers (e.g., BatchNorm runing mean and variance) to the desired dtype. Buffers are not considered parameters but are inputs to the computational graph and must be aligned in data format with the rest of the model.

- Added pytorch tests for casting model to bfloat16.

### 
What is not done:

Casting model to bfp8_b. There are some concerns regarding bfp8_b implementation in forge (mainly because torch doesn't support float8 and block formats and we rely on torch for verification and model tracing before TVM). Also currently there is no support for lowering bfp8_b to mlir.

### DIFF
--- a/forge/csrc/forge_passes.cpp
+++ b/forge/csrc/forge_passes.cpp
@@ -80,6 +80,7 @@ run_post_initial_graph_passes(
     std::shared_ptr<void> compiler_cfg = make_shared_py_object(compiler_cfg_object);
 
     passes::print_graph(graph, "INITIAL");
+    passes::apply_user_data_format_override(graph, compiler_cfg_object);
     passes::generate_initial_flops_estimate(graph);
     passes::decompose_nd_reshape_split(graph);
     passes::erase_unnecessary_4d_tm_sequence(graph);
@@ -211,12 +212,6 @@ graphlib::Graph *run_pre_lowering_passes(graphlib::Graph *graph, const std::opti
     // Manually convert broadcast ops to tms, so insert tile broadcast ops can work generically
     // Note this is not lowering, these are still forge tms
     convert_broadcast_ops_to_tms(graph);
-
-    //
-    // Data formats
-    //
-    // Apply user overrides
-    passes::configure_output_data_formats(graph, default_df_override);
 
     passes::remove_nops(graph);
 
--- a/forge/csrc/passes/dataformat.cpp
+++ b/forge/csrc/passes/dataformat.cpp
@@ -230,7 +230,57 @@ void lower_fallback_data_formats(graphlib::Graph *graph, DataFormat fp32_fallbac
     }
 }
 
-// Apply user overrides
+// Inserts a cast node after every input node in the graph,
+// placing it on the first user edge and re-routing all other edges.
+// The cast node will convert outputs to the given DataFormat.
+static void insert_cast_on_input_nodes(graphlib::Graph *graph, DataFormat df_override)
+{
+    for (Node *node : graph->nodes())
+    {
+        if (node->node_type() != graphlib::NodeType::kInput)
+            continue;
+
+        std::vector<Edge> user_edges = graph->user_data_edges(node);
+        if (user_edges.empty())
+            continue;
+
+        // Generate cast node
+        std::stringstream ss;
+        ss << df_override;
+        std::string dtype_str = ss.str();
+        graphlib::OpType::Attrs named_attr = graphlib::OpType::Attrs{{"dtype", dtype_str}};
+        graphlib::OpType op_type = graphlib::OpType("cast", {}, {}, named_attr);
+
+        Node *cast_node = graph->add_node(
+            graphlib::create_node<graphlib::PyOpNode>("cast_input_" + node->name(), op_type),
+            graph->get_subgraph_id_for_node(node->id()));
+
+        cast_node->set_shape(node->shape());
+        cast_node->set_output_df(df_override);
+
+        // First edge: insert cast node on it.
+        graphlib::insert_node_on_edge(graph, user_edges[0], cast_node);
+        TT_ASSERT(cast_node != nullptr, "Cast node should not be null");
+
+        // Remaining edges: reconnect edges to cast node.
+        for (size_t i = 1; i < user_edges.size(); ++i)
+        {
+            Edge &edge = user_edges[i];
+
+            Edge new_edge = Edge(
+                cast_node->id(),
+                0,  // output port
+                edge.consumer_node_id,
+                edge.consumer_input_port_id,
+                edge.edge_type);
+            graph->add_edge(new_edge);
+            graph->copy_edge_attributes(edge, new_edge);
+            graph->remove_edge(edge);
+        }
+    }
+}
+
+// Sets output data formats of all nodes to user specified data format.
 void configure_output_data_formats(graphlib::Graph *graph, std::optional<DataFormat> default_df_override)
 {
     for (Node *node : graph->nodes())
@@ -241,6 +291,24 @@ void configure_output_data_formats(graphlib::Graph *graph, std::optional<DataFor
             node->set_output_df(preserve_lower_precision_cast(node->output_df(), *default_df_override));
         }
     }
+}
+
+// Inserts cast nodes (user specified data format) on all input nodes in the graph,
+// and sets output data formats of all nodes to user specified data format.
+void apply_user_data_format_override(graphlib::Graph *graph, py::object compiler_cfg_object)
+{
+    // Extract optional default_df_override from Python
+    std::optional<DataFormat> default_df_override = std::nullopt;
+    py::object attr = compiler_cfg_object.attr("default_df_override");
+    if (!attr.is_none())
+        default_df_override = attr.cast<DataFormat>();
+
+    // Skip everything if no override is provided
+    if (!default_df_override)
+        return;
+
+    insert_cast_on_input_nodes(graph, *default_df_override);
+    configure_output_data_formats(graph, default_df_override);
 }
 
 void configure_input_data_formats(graphlib::Graph *graph)
--- a/forge/forge/tvm_calls/forge_compile.py
+++ b/forge/forge/tvm_calls/forge_compile.py
@@ -448,6 +448,13 @@ class ConvertEmulatedDtypes:
                 self.param_dfs[name] = param.dtype
                 param.data = param.data.to(self.fallback)
 
+        # Convert buffers
+        self.buffer_dfs = {}
+        for name, buf in self.model.named_buffers():
+            if buf.dtype in self.emulated_dfs:
+                self.buffer_dfs[name] = buf.dtype
+                buf.data = buf.data.to(self.fallback)
+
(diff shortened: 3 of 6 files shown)


forge/csrc/passes/dataformat.cpp
```diff
@@ -230,9 +230,65 @@ void lower_fallback_data_formats(graphlib::Graph *graph, DataFormat fp32_fallbac
...
+                edge.consumer_input_port_id,
+                edge.edge_type);
+            graph->add_edge(new_edge);
+            graph->copy_edge_attributes(edge, new_edge);
+            graph->remove_edge(edge);
+        }
+    }
+}
+
 // Apply user overrides
```
nvukobratTT, line 283: Let's be more specific with this doc here.

forge/csrc/passes/dataformat.cpp
```diff
@@ -230,9 +230,65 @@ void lower_fallback_data_formats(graphlib::Graph *graph, DataFormat fp32_fallbac
...
 {
+    // Skip everything if no override is provided
+    if (!default_df_override)
+        return;
+
+    insert_cast_on_input_nodes(graph, *default_df_override);
+
     for (Node *node : graph->nodes())
     {
         bool disallow_default_override = is_integer_data_format(node->output_df());
```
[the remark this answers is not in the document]
nvukobratTT: Similar note applies for `preserve_lower_precision_cast`. Let's get back to the drawing board once the holidays are over (in 2 - 3 weeks).

forge/forge/tvm_calls/forge_compile.py
```diff
@@ -448,6 +448,13 @@ def __enter__(self):
                 self.param_dfs[name] = param.dtype
                 param.data = param.data.to(self.fallback)
 
+        # Convert buffers
```
nvukobratTT, line 451: Is this needed with proposed logic of converting data formats after TVM pass?

forge/test/mlir/test_dataformats.py
```diff
@@ -0,0 +1,143 @@
...
+import onnx.numpy_helper
+
+import forge
+from forge.verify.verify import verify
+from forge.config import CompilerConfig
+from forge._C import DataFormat
+
+
+# PyTorch test remains the same
+@pytest.mark.parametrize("shape", [(1, 3, 8, 8)])
```
nvukobratTT, line 22: **Not a blocker for this PR**

Also, we should expand these tests for all supported PT data formats, and after that, to all framework data formats. Even if some are failing, let's mark them as xfail so we know what works and what doesn't.

nvukobratTT: I.e. testing conv2d (as shown here), embeddings, etc.

forge/csrc/passes/dataformat.cpp
```diff
@@ -230,9 +230,65 @@ void lower_fallback_data_formats(graphlib::Graph *graph, DataFormat fp32_fallbac
...
+        cast_node->set_output_df(df_override);
+
+        // First edge: insert cast node on it.
+        graphlib::insert_node_on_edge(graph, user_edges[0], cast_node);
+
+        // Remaining edges: reconnect edges to cast node.
+        for (size_t i = 1; i < user_edges.size(); ++i)
+        {
+            Edge &edge = user_edges[i];
+            TT_ASSERT(cast_node != nullptr, "Cast node should not be null");
```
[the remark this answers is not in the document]
dgolubovicTT: Yeah, I moved it after creation, before loop

forge/csrc/forge_passes.cpp
```diff
@@ -78,8 +78,13 @@ run_post_initial_graph_passes(
     graphlib::Graph *graph, py::object compiler_cfg_object, passes::FractureGroups const &fracture_groups)
 {
     std::shared_ptr<void> compiler_cfg = make_shared_py_object(compiler_cfg_object);
-
+    // Extract optional default_df_override from Python
```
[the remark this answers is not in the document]
dgolubovicTT: makes sense, done

forge/csrc/passes/dataformat.cpp
dgolubovicTT: Yeah, that is not informative. I've split the function to:
1. configure_output_data_formats - That sets output data formats for all nodes in graph (as it was always)
2. new: insert_cast_on_input_nodes - that adds cast nodes after each input node.
3. new: apply_user_data_format_override - that calls both 1. and 2.

everything commented in the code now

forge/csrc/passes/dataformat.cpp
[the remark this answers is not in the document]
dgolubovicTT: I wouldn't say this is wrong, and here is why:

1. TTNN embedding doesn't allow input to be float. Therefore if we allowed cast of integers we would hit assert in ttnn.
2. preserve_lower_precision_cast keeps us from loosing exponent (would prevent casting to bfp_8). Since float32 and bfloat16 have the same exponent this is not the issue for now. 

I agree that we will have to rewrite many things in data format logic in next period. I am fully aware of these two points.

forge/forge/tvm_calls/forge_compile.py
dgolubovicTT: This bug is not related to converting data formats after TVM. This bug was exposed when I tried compiling ResNet with model.to(torch.bfloat16)

Since our torch model can be in dtype other than float32, context manager(ConvertEmulatedDtypes) ensures that both our model and inputs are in float32 inside with block `(with ConvertEmulatedDtypes(torchmod, inputs):)`

Inside that block, model is traced using `torch.jit.trace` and traced model is input to TVM.
TVM requires whole model to be in data formats that are convertible to numpy (bfloat16 is not, therefore context manager block converts everything to float32). Turns out, one part of the model wasn't being converted to float32 and stayed in bfloat16 which caused error. Those were buffers of model.

What I added is converting model buffers dtype same as parameters were converted. 

Buffers are non learnable parameters that usually keep internal states. One example (the one I caught the bug) is BatchNorm. BatchNorm layers internally keep running mean and variance — they are buffers, not parameters. That is present in ResNet and I caught the bug as soon as I tried running compile for model.to(torch.bfloat16).

forge/forge/tvm_calls/forge_compile.py
```diff
@@ -491,25 +504,23 @@ def compile_pytorch_for_forge(
...
         traced_model = torch.jit.trace(torchmod, inputs, check_trace=False, strict=False)
-
-    # Extract flatten inputs
-    flattened_inputs, flattened_input_names, flattened_name_map, input_structure = extract_flatten_inputs(
-        framework="pytorch",
-        model=traced_model,
-        inputs=inputs,
-        input_names=input_names,
-    )
+        # Extract flatten inputs
```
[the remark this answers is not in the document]
dgolubovicTT: I gave an extensive explanation in previous comment. 
TL;DR: This change is orthogonal to dataformat overrides. Now we are calling tvm relay pass on traced model whlie traced model is still in FLOAT32 (inside with block). The purpose of with block is to enable traced_model to be in float32 for TVM to consume. Therefore `tvm.relay.frontend.from_pytorch` should be callled inside with block.

The reason why we haven't seen any issues before with this being outside with block is because models were already in float32 coming from torch.

forge/test/benchmark/benchmark/models/resnet_hf.py
```diff
@@ -71,14 +73,18 @@ def test_resnet_hf(
...
     # Random data
-    input_sample = [torch.rand(batch_size, channel_size, *input_size)]
+    input_sample = [torch.rand(batch_size, channel_size, *input_size).to(torch.bfloat16)]
 
     # Load framework model
-    framework_model = download_model(ResNetForImageClassification.from_pretrained, variant, return_dict=False)
+    framework_model = download_model(
+        ResNetForImageClassification.from_pretrained, variant, return_dict=False, torch_dtype=torch.bfloat16
+    )
+    framework_model = framework_model.to(dtype=torch.bfloat16)
```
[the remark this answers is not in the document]
dgolubovicTT: Fully agree, i need to look into it more...

vcanicTT:
@dgolubovicTT Can you change the batch size for resnet in these two files?

https://github.com/tenstorrent/tt-forge-fe/blob/main/.github/workflows/perf-benchmark-sub.yml#L22

https://github.com/tenstorrent/tt-forge-fe/blob/main/forge/test/benchmark/scripts/run_benchmark_wh.sh#L14

vcanicTT (approved):

odjuricicTT (approved):
🚀 Thanks for getting this in!

forge/forge/tvm_calls/forge_compile.py
nvukobratTT: Great! Thanks for clarification

pilkicTT (approved):

nvukobratTT (approved):

dgolubovicTT:
@odjuricicTT @nvukobratTT @vcanicTT Even though ResNet benchmark test passes on this branch locally with batch size 8 and bfloat16, when I run nightly it fails. To avoid regressing the nightlies, I will merge this PR without changes in ResNet test, but here is the branch with those changes where you can verify if it works or not( [dgolubovic/enable-bfloat16-user-override-with-resnet-test-changes](https://github.com/tenstorrent/tt-forge-fe/tree/dgolubovic/enable-bfloat16-user-override-with-resnet-test-changes)). I don't have time now to look into the discrepancy with CI nightly.

vcanicTT:
> @odjuricicTT @nvukobratTT @vcanicTT Even though ResNet benchmark test passes on this branch locally with batch size 8 and bfloat16, when I run nightly it fails. To avoid regressing the nightlies, I will merge this PR without changes in ResNet test, but here is the branch with those changes where you can verify if it works or not( [dgolubovic/enable-bfloat16-user-override-with-resnet-test-changes](https://github.com/tenstorrent/tt-forge-fe/tree/dgolubovic/enable-bfloat16-user-override-with-resnet-test-changes)). I don't have time now to look into the discrepancy with CI nightly.

ok, I'll take a look

PR #1835, opened 2025-04-22
Update model analysis documentation

vmilosevic:
This PR will update model analysis documentation.


nvukobratTT:
@chandrasekaranpradeep is this still an valid and active PR? Should we consider to push effort to depricate this logic now that we're closer to stable Superset for the data overview? :))

nvukobratTT:
> > @chandrasekaranpradeep is this still an valid and active PR? Should we consider to push effort to depricate this logic now that we're closer to stable Superset for the data overview? :))
> 
> The generated models ops docs PR is valid but it was generated 2 weeks before. May be we can remove the github workflow scheduling part and later we can remove the logic of generating docs. What do you think?

Sounds like a plan! Let's create issues for that and push removal for scheduling in a following days, and latter depricate the logic in place :))

chandrasekaranpradeep:
> > > @chandrasekaranpradeep is this still an valid and active PR? Should we consider to push effort to depricate this logic now that we're closer to stable Superset for the data overview? :))
> > 
> > 
> > The generated models ops docs PR is valid but it was generated 2 weeks before. May be we can remove the github workflow scheduling part and later we can remove the logic of generating docs. What do you think?
> 
> Sounds like a plan! Let's create issues for that and push removal for scheduling in a following days, and latter depricate the logic in place :))

Okay Nikola

chandrasekaranpradeep:
> > > @chandrasekaranpradeep is this still an valid and active PR? Should we consider to push effort to depricate this logic now that we're closer to stable Superset for the data overview? :))
> > 
> > 
> > The generated models ops docs PR is valid but it was generated 2 weeks before. May be we can remove the github workflow scheduling part and later we can remove the logic of generating docs. What do you think?
> 
> Sounds like a plan! Let's create issues for that and push removal for scheduling in a following days, and latter depricate the logic in place :))

@nvukobratTT  Created a issues: https://github.com/tenstorrent/tt-forge-fe/issues/1985

PR #1837, opened 2025-04-22
Enable program cache in benchmark models mobilenet and efficientnet.

vcanicTT:
Program cache works again on mobilenet and efficientnet, so we enable it.


vcanicTT:

pilkicTT (approved):

PR #1840, opened 2025-04-22
Enable device perf collection on CI for Llama benchmark model.

vcanicTT:
Now, we are able to run device run on Llama benchmark model, too. So, we enable it. The issue with TTRT is solved. There is the [issue](https://github.com/tenstorrent/tt-mlir/issues/2752#issuecomment-2786438092).


vcanicTT:

pilkicTT (approved):

PR #1848, opened 2025-04-22
Support Jax and TF framework in models ops test generation

chandrasekaranpradeep:
Fixes https://github.com/tenstorrent/tt-forge-fe/issues/1850
1. Added support for TensorFlow and JAX framework in models ops tests generation pipeline
2. Added resnet50 models test for TensorFlow framework.
3. Updated the advindex op maximum int calculation for indicies tensor generation due to Forge AdvIndex op refactoring in this [commit](https://github.com/tenstorrent/tt-forge-fe/commit/bee42144423d24fa3187cb6108c490088e3b62fb) 
4. Fixed an issue in `format_tvm_graph_weights` (in `forge/forge/tvm_calls/forge_compile.py`) where all Keras weights—including non-trainable BatchNorm statistics (moving_mean/moving_variance)—were incorrectly marked requires_grad=True, causing them to be treated as learnable parameters instead of constants; this change updates the logic to set requires_grad=True only for weights present in module.trainable_weights, so that BatchNorm moving stats remain exported as constants, and added a batchnorm tests to verify it

### DIFF
--- a/forge/forge/tvm_to_python.py
+++ b/forge/forge/tvm_to_python.py
@@ -2737,7 +2737,7 @@ def compile_tvm_to_python(
 
         modules.append(writer)
 
-        if (framework in ["pytorch", "paddle", "onnx"] and compiler_cfg.extract_tvm_unique_ops_config) or (
+        if (framework in ["pytorch", "paddle", "onnx", "jax"] and compiler_cfg.extract_tvm_unique_ops_config) or (
             framework == "pytorch" and compiler_cfg.tvm_generate_unique_ops_tests
         ):
 
--- a/forge/forge/tvm_to_python.py
+++ b/forge/forge/tvm_to_python.py
@@ -2737,7 +2737,7 @@ def compile_tvm_to_python(
 
         modules.append(writer)
 
-        if (framework in ["pytorch", "paddle", "onnx", "jax"] and compiler_cfg.extract_tvm_unique_ops_config) or (
+        if compiler_cfg.extract_tvm_unique_ops_config or (
             framework == "pytorch" and compiler_cfg.tvm_generate_unique_ops_tests
         ):
 
--- a/forge/test/models/tensorflow/vision/resnet/test_resnet.py
+++ b/forge/test/models/tensorflow/vision/resnet/test_resnet.py
@@ -1,0 +1,51 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+
+from tensorflow.keras.applications import ResNet50
+
+import forge
+from forge.forge_property_utils import Framework, Source, Task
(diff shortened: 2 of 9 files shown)


.github/workflows/model-analysis.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
   model-analysis:
     needs: docker-build
-    runs-on: tt-beta-ubuntu-2204-n150-large-stable
+    runs-on: runner
```
vmilosevic, line 27: Why are we changing this back to run on runners? Are there some issues with shared runner pool?

chandrasekaranpradeep: I tried triggering the models ops tests generation pipeline with `tt-beta-ubuntu-2204-n150-large-stable-wrt28-runner-m4tm2` runner which take less than a day but it was failling with no space left on device issues - https://github.com/tenstorrent/tt-forge-fe/actions/runs/14666435600

so I have changed it to runner but even in the runners, the pytest process was killed - https://github.com/tenstorrent/tt-forge-fe/actions/runs/14669518818. 

I am trying to generate models ops test(i.e [forge/test/models_ops](https://github.com/tenstorrent/tt-forge-fe/tree/main/forge/test/models_ops)) from this branch because the models ops test was not updated for last 3 weeks. So the plan is not change the runner in this PR.

chandrasekaranpradeep: Removed the runner updation because it was changed for generating models ops tests in this PR branch

forge/forge/tvm_to_python.py
```diff
@@ -2737,9 +2737,10 @@ def delete_unneeded_outputs(ops, returns):
 
         modules.append(writer)
 
-        if (framework in ["pytorch", "paddle", "onnx"] and compiler_cfg.extract_tvm_unique_ops_config) or (
-            framework == "pytorch" and compiler_cfg.tvm_generate_unique_ops_tests
-        ):
+        if (
```
nvukobratTT, line 2740: Is there any Framework that we're currently not supporting for unique ops? If no, we can simplify this check a bit.

forge/test/models/tensorflow/vision/resnet/test_resnet.py
```diff
@@ -0,0 +1,51 @@
...
+from forge.forge_property_utils import Framework, Source, Task
+from forge.verify.config import VerifyConfig
+from forge.verify.value_checkers import AutomaticValueChecker
+from forge.verify.verify import verify
+from test.models.tensorflow.vision.resnet.utils.image_utils import get_sample_inputs
+
+
+@pytest.mark.push
+@pytest.mark.nightly
+def test_resnet_tensorflow(forge_property_recorder):
```
nvukobratTT, line 19: FYI @odjuricicTT, we expect to have ResNet TF variant functional once this PR lands on main.

nvukobratTT (approved):
Minor comment, approving as it's not a blocking change

PR #1849, opened 2025-04-22
Change verification of Resnet Benchmark Model

vcanicTT:
Add standard way of verification of result correctness for benchmark model Resnet. No, resnet, like other models, uses verify function.

### DIFF
--- a/forge/test/benchmark/benchmark/models/resnet_hf.py
+++ b/forge/test/benchmark/benchmark/models/resnet_hf.py
@@ -18,9 +18,9 @@ from datasets import load_dataset
 
 # Forge modules
 import forge
-from forge.verify.value_checkers import AutomaticValueChecker
-from forge.verify.verify import verify
 from forge.verify.config import VerifyConfig
+from forge.verify.verify import verify
+from forge.verify.value_checkers import AutomaticValueChecker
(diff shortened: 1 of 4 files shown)


vcanicTT:

forge/test/benchmark/benchmark/models/resnet_hf.py
```diff
@@ -75,11 +78,18 @@ def test_resnet_hf(
 
     # Load framework model
     framework_model = download_model(ResNetForImageClassification.from_pretrained, variant, return_dict=False)
-    fw_out = framework_model(*input_sample)
 
     # Compile model
     compiled_model = forge.compile(framework_model, *input_sample)
 
+    # Model Verification
+    verify(
```
pilkicTT, line 86: You should run verify instead of the warm up run... So that `verify()` will serve both as golden verification and warm up for the cache... Additionally, I would advise to run the checker manually once again, after all of the iterations are done - to verify that the output after Nth iteration is still the same.

pilkicTT (changes requested):

forge/test/benchmark/benchmark/models/resnet_hf.py
vcanicTT: Changed

pilkicTT (approved):

PR #1855, opened 2025-04-23
[CI tests] removed xfailed and skipped one test

vkovinicTT:
### What's changed
Skipped one test due to transient failure and removed `xfail` mark from one test as it is passing now.

### DIFF
--- a/forge/test/models/onnx/text/qwen/test_qwen_v2.py
+++ b/forge/test/models/onnx/text/qwen/test_qwen_v2.py
@@ -32,7 +32,7 @@ import torch
         ),
         pytest.param(
             "Qwen/Qwen2.5-0.5B-Instruct",
-            marks=pytest.mark.xfail,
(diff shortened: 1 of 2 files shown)


forge/test/models/onnx/text/qwen/test_qwen_v2.py
```diff
@@ -32,7 +32,7 @@
         ),
         pytest.param(
             "Qwen/Qwen2.5-0.5B-Instruct",
-            marks=pytest.mark.xfail,
+            marks=pytest.mark.skip("Transient test - Out of memory due to other tests in CI pipeline"),
```
pilkicTT, line 35: Can you run this one locally to see how much RAM it uses?

pilkicTT (approved):

forge/test/models/onnx/text/qwen/test_qwen_v2.py
vkovinicTT: 0.5B is small model but I skipped it nevertheless, until we add all the OOM fixes.

PR #1858, opened 2025-04-23
[CI] Change model-analysis runner

nsumrakTT:
### Ticket
f2f

### Problem description
Performance runners are for performance benchmarks and not for long runs.

### What's changed
Changed to shared company runner.

### DIFF
--- a/.github/workflows/model-analysis.yml
+++ b/.github/workflows/model-analysis.yml
@@ -24,7 +24,7 @@ jobs:
 
   model-analysis:
     needs: docker-build
-    runs-on: performance
+    runs-on: tt-beta-ubuntu-2204-n150-large-stable
     timeout-minutes: 5760 # Set job execution time to 4 days(default: 6 hours)
 
     container:


vcanicTT (approved):

.github/workflows/model-analysis.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
   model-analysis:
     needs: docker-build
-    runs-on: performance
+    runs-on: tt-beta-ubuntu-2204-n150-large-stable
```
pilkicTT, line 27: Should we use shared runner for this? Mainly because the job duration is extremely long...

nsumrakTT: Well, name has "beta" in it, so let's test it :-)

chandrasekaranpradeep: The model-analysis documentation generation pipeline is expected to run for four days, while the model-ops test generation pipeline will complete in one day. I recommend triggering both pipelines on the updated runner and verifying that everything continues to function correctly.

PR #1870, opened 2025-04-24
[compile] set argument type when lowering to mlir

### LINKED ISSUE #1867: [compile] Mark weights for mlir consteval

pilkicTT:
The attribute `tt.argument_type` is used to distinguish between model parameters (weights), constants or regular inputs (which are expected to change on each forward call).

This attribute is used in consteval logic in `tt-mlir` to hoist parts of the graph which can be evaluated only once in runtime, cached on device and reused afterwards (in subsequent program invocations).

Closes #1867


odjuricicTT (approved):
Looks good :)

PR #1873, opened 2025-04-24
[mlir] basic support for configuring MLIR pipeline

### LINKED ISSUE #1869: Expose mlir compile options in compile config

pilkicTT:
Adding structure `MLIRConfig` to enable setting some of the pipeline options in `tt-mlir`. This structure is part of the `CompilerConfig` and is set to `None` by default.

For now, the only config we expose is the one controlling consteval in `tt-mlir`. Settings for enabling optimizer will follow.

Also, marking tests in `test_api.py` to be ran on `push`.

Closes #1869

### DIFF
--- a/forge/csrc/forge_bindings.cpp
+++ b/forge/csrc/forge_bindings.cpp
@@ -209,6 +209,22 @@ PYBIND11_MODULE(_C, m)
 
     py::register_exception<UnsupportedHWOpsError>(m, "UnsupportedHWOpsError");
 
+    py::class_<tt::passes::MLIRConfig>(m, "MLIRConfig")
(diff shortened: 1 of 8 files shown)


odjuricicTT (approved):
I was under the impression that we are just going to be passing a string instead of the `MLIRConfig` struct.

The reason for this being that we already have something which tries to solve the same problem as the MLIRConfig struct in tt-mlir:
https://github.com/tenstorrent/tt-mlir/blob/main/include/ttmlir/Dialect/TTNN/Utils/OptimizerOverrides.h#L17
With the idea that this can be standardised and used by all frontends.

Not that strongly opinionated tho.

PR #1891, opened 2025-04-28
[data collection] skip recording details from fb

pilkicTT:
When executing model tests, skip recording details from the flatbuffer, since this results in large amount of data being collected.


pilkicTT:
@chandrasekaranpradeep FYI

vmilosevic (approved):

PR #1906, opened 2025-04-30
Disable cache for mamba model and modify CastWhereConditionToBool TVM Callback

meenakshiramanathan1:
Cache has been disabled for this model which was causing tvm issue
`InternalError: Check failed: (!axes.defined() || static_cast<int>(axes.size()) == ndim) is false: 
Dimension mismatch: axes has 3 elements, but data.ndim = 6`

After this change, model faced `Encountered unsupported op node type: nn.dense, on device: tt`, root cause of this issue was from CastWhereConditionToBool TVM Callback. This callback was using the pre matched graph for modifications, which unintentionally reintroduced nn.dense ops that had already been decomposed. This caused failures as unsupported nn.dense. Changing the pass to use post ensures it operates on the latest graph state, preserving prior rewrites and preventing unsupported ops from being reintroduced.

Logs:
[mamba_before_fix.log](https://github.com/user-attachments/files/20053331/mamba_before_fix.log)
[mamba_after_fix.log](https://github.com/user-attachments/files/20053334/mamba_after_fix.log)

### DIFF
--- a/forge/forge/tvm_calls/relay/op/forge_passes.py
+++ b/forge/forge/tvm_calls/relay/op/forge_passes.py
@@ -1293,8 +1293,8 @@ class CastWhereConditionToBool(DFPatternCallback):
         self.pattern = is_op("where")(wildcard(), wildcard(), wildcard())
 
     def callback(self, pre, post, node_map):
-        cond = tvm.relay.cast(pre.args[0], "bool")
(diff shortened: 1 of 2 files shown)


saiarthiraguram (approved):

mstojkovicTT (approved):
Looks good :))

PR #1908, opened 2025-04-30
[mlir] construct hierarchical locations

pilkicTT:
The current logic for creating locations, forms them in the following way:

`NameLoc(graph_name, child_loc)` - where graph_name is in most cases `"forward"` and `child_loc` contains the actual location: `module/layer1/layer2/node_name`.

There are two problems with this:
 1. `NameLoc` contains only value `"forward"` as its value, so we cannot discern between nodes by looking at it (we would need to get the child location instead)
 2. even if the direct location would contain the layer information, the string is far too large

Since MLIR supports nesting of locations, logic for creating the locations is modified so that it parses the location string and creates nested instances of `NameLoc`. In this way, we have the `NameLoc` attached to a particular op containing its name. If we wish to see more details about where this particular op belongs, we can go through the chain of locations and reconstruct everything.

Example (location string -> `NameLoc`):

`"module_name/layer/op_name"` ->
`NameLoc("op_name", NameLoc("layer", NameLoc("module_name")))`

Also, modify the way we dump the module for `reportify` so that the locations are included in the dump.

### DIFF
--- a/forge/csrc/passes/lower_to_mlir.cpp
+++ b/forge/csrc/passes/lower_to_mlir.cpp
@@ -8,6 +8,7 @@
 #include <cstdint>
 #include <stdexcept>
 #include <string>
+#include <string_view>
 #include <utils/assert.hpp>
 #include <variant>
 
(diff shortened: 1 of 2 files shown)


forge/csrc/passes/lower_to_mlir.cpp
```diff
@@ -672,30 +673,55 @@ class MLIRGenerator
...
+            std::string_view name = layer_tag.substr(pos + 1);
+
+            mlir::StringAttr name_attr = builder_.getStringAttr(name);
+            return mlir::NameLoc::get(name_attr, parse_layer_tag(layer_tag.substr(0, pos)));
+        }
+        else
+        {
+            mlir::StringAttr name_attr = builder_.getStringAttr(layer_tag);
+            return mlir::NameLoc::get(name_attr);
+        }
```
azecevicTT, line 696: Since you are linking this with LLVM, `StringRef` has a lot of very convenient member functions, this could be rewritten as
```suggestion
        StringRef layer_tag;
        auto [prefix, sufix] = layer_tag.rsplit('/');
        if (!sufix.empty())
        {
            mlir::StringAttr name_attr = builder_.getStringAttr(sufix);
            return mlir::NameLoc::get(name_attr, parse_layer_tag(prefix));
        }
        else
        {
            mlir::StringAttr name_attr = builder_.getStringAttr(prefix);
            return mlir::NameLoc::get(name_attr);
        }
```

dgolubovicTT (approved):

odjuricicTT (approved):

forge/csrc/passes/lower_to_mlir.cpp
pilkicTT: i'll keep the `std` version, just to be aligned with rest of `tt-forge-fe` - we're using the MLIR/LLVM stuff where we need to...

PR #1916, opened 2025-05-02
Monkey patch: fix ONNX Where type error in Gemma2DecoderLayer by matching dtype

kamalrajkannan78:
### Problem description
While running the exported gemma_v2 model in ONNX Runtime, the following type mismatch error was encountered:

```
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Load model from /tmp/pytest-of-kkannan/pytest-1/test_gemma_v2_onnx_google_gemm1/gemma_v2.onnx failed:Type Error: Type parameter (T) of Optype (Where) bound to different types (tensor(double) and tensor(float) in node (/model/layers.0/Where).
```

This error occurs because [torch.finfo(...).min](https://github.com/huggingface/transformers/blob/5d7739f15a6e50de416977fe2cc9cb516d67edda/src/transformers/models/gemma2/modular_gemma2.py#L510C29-L510C65) returns a Python scalar of type float (i.e., double in ONNX), which causes a type mismatch when used in a [torch.where(...)](https://github.com/huggingface/transformers/blob/5d7739f15a6e50de416977fe2cc9cb516d67edda/src/transformers/models/gemma2/modular_gemma2.py#L514C34-L514C93) operation alongside attention_mask(float32)

### What's changed

To resolve this issue, a monkey patch was applied to Gemma2DecoderLayer.forward. The patch ensures that the scalar is cast to a tensor with the same data type as hidden_states, maintaining type consistency in the torch.where(...) operation and avoiding ONNX type errors.

### Logs

- [gemma2_before_fix.log](https://github.com/user-attachments/files/20014471/may2_gemma_before_fix.log)
- [gemma2_2b_after_fix.log](https://github.com/user-attachments/files/20014482/may2_gemma_2b_after_fix.log)

### DIFF
--- a/forge/test/models/onnx/text/gemma/test_gemma_v2.py
+++ b/forge/test/models/onnx/text/gemma/test_gemma_v2.py
@@ -13,6 +13,10 @@ from forge.verify.verify import verify
 from forge.forge_property_utils import Framework, Source, Task
 from test.utils import download_model
 import onnx
+from transformers.models.gemma2.modeling_gemma2 import Gemma2DecoderLayer
+from test.models.models_utils import Gemma2DecoderLayer_patched_forward
+
+Gemma2DecoderLayer.forward = Gemma2DecoderLayer_patched_forward
 
 
 @pytest.mark.nightly
--- a/forge/test/models/models_utils.py
+++ b/forge/test/models/models_utils.py
@@ -9,6 +9,8 @@ from transformers import AutoImageProcessor
 import torch
 from tabulate import tabulate
 import json
+from typing import Optional, Tuple
+from transformers import Cache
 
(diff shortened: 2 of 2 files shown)


saiarthiraguram (approved):

forge/test/models/onnx/text/gemma/test_gemma_v2.py
```diff
@@ -13,6 +13,74 @@
 from forge.forge_property_utils import Framework, Source, Task
 from test.utils import download_model
 import onnx
+from transformers.models.gemma2.modeling_gemma2 import Gemma2DecoderLayer
```
saiarthiraguram, line 16: can you move this patch to utils file?

PR #1919, opened 2025-05-03
Update forge-fe requirements

nsmithtt:
This commit updates forge-fe core requirements such that they are compatible with the other frontends, tt-torch and tt-xla.  This enables all of their respective wheels to be installed inside of the same venv.

### DIFF
--- a/env/core_requirements.txt
+++ b/env/core_requirements.txt
@@ -1,22 +1,17 @@
 setuptools>=61
-# This is needed to avoid issue https://yyz-gitlab.local.tenstorrent.com/devops/devops/-/issues/95
-# jax requires any version of optax which requires any version of chex which in turn
-# requires jax>=0.4.6 which conflicts with our jax == 0.3.16
-# TODO: Remove when jax library is upgraded
-chex==0.1.6
 dataclasses-json==0.5.7
 datasets==2.14.6
 decorator==5.1.1
-flatbuffers==23.5.26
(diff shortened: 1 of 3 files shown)


nsmithtt:
Note, I'm mostly just kicking this off as a litmus test.  I'm not necessarily intending to land this, just curious since this could be a nice user QOL by enabling them to install any FE wheel into the same venv.

nsmithtt:
Getting this error from some gtests (though I'm not sure these tests are relevant):
```
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.
```

nvukobratTT:
Closing this PR as the PyTorch lib push will be done through this issue:
- https://github.com/tenstorrent/tt-forge-fe/issues/1984

Once this issue is resolved, we're pushing effort on lib cleanup (mostly on removing unused and stalled dependencies). Tracking issues:
- https://github.com/tenstorrent/tt-forge-fe/issues/1980
- https://github.com/tenstorrent/tt-forge-fe/issues/1981

cc @saiarthiraguram

PR #1921, opened 2025-05-05
Disable device perf because of troubles with TTRT.

vcanicTT:
TTRT causes troubles with performance benchmark tests. Until we solve the problems, we disable preserving device perf, because without disabling we can't preserve even end-to-end perf into the database.


vcanicTT:
@nsumrakTT please, take a look!

nsumrakTT (approved):

PR #1926, opened 2025-05-05
[mlir] enable using tt-mlir optimizer

### LINKED ISSUE #1598: Enable optimizer for benchmark models
Dependent on #1597 

This should be done by passing the following options string into the ttnn pipeline:
```
--ttir-to-ttnn-backend-pipeline="enable-optimizer=true memory-layout-analysis-enabled=true"
```
Not sure if `system-desc-path` is also required.

Start with enabling it on mnist benchmark. Resnet might run into some errors on first try.

pilkicTT:
To be able to run the optimizer, it is required to build `tt-mlir` with certain flags - hence, passing these flags to the build.

The knobs to enable optimizer, memory layout analysis are added.

Also, there is a way to provide a custom string to be passed to MLIR - for use in testing or for configs not properly exposed.

Made all config fields `std::optional`, so that user can specify only one config flag, and the rest will not be specified in the pipeline options string, i.e. they'll have their default values defined in mlir.

Enabled the optimizer and memory layout analysis on MNIST benchmark - however, due to an issue we need to disable the device program cache.

Also optimizer without memory layout analysis is enabled for Resnet benchmark.

Closes #1598


vkovinicTT (approved):

odjuricicTT (approved):
Awesome!

PR #1927, opened 2025-05-05
Test EmitC path in verify fn

svuckovicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/3109

### Problem description
N/A

### What's changed
This change implements testing infra for EmitC. As is currently written, each verify call will end up compiling a dylib (.so), running it, and comparing it vs the output of the regular tt-mlir runtime path.

### Checklist


.github/workflows/test-sub.yml
```diff
@@ -152,7 +152,7 @@ jobs:
         shell: bash
         run: |
           source env/activate
-          pytest --splits ${{ inputs.test_group_cnt }} \
+          pytest -svv --splits ${{ inputs.test_group_cnt }} \
```
svuckovicTT, line 155: remove

forge/csrc/passes/mlir_compiler.cpp
```diff
@@ -131,6 +140,70 @@ auto run_mlir_compiler_generic(
...
+
+        if (!fs::exists(in_wheel_path) && !fs::exists(in_source_path))
+        {
+            throw std::runtime_error("Neither tt-metal wheel nor source path exists.");
+        }
+
+        fs::path metal_src_dir;
+        fs::path metal_lib_dir;
+        fs::path standalone_dir;
+        if (fs::exists(in_wheel_path))
```
svuckovicTT, line 182: Could improve this by setting something in `forge/forge/__init__.py::set_home_paths` which already checks whether we're "in wheel" or "in source".

forge/csrc/passes/mlir_compiler.cpp
```diff
@@ -131,6 +140,70 @@ auto run_mlir_compiler_generic(
...
+            assert(in_source_path == std::string(TT_METAL_HOME));
+
+            metal_src_dir = fs::path(std::string(TT_METAL_HOME));
+            metal_lib_dir = fs::path(std::string(TT_METAL_HOME)).parent_path() / "tt-metal-build/lib";
+            standalone_dir =
+                fs::path(std::string(FORGE_HOME)).parent_path() / "third_party/tt-mlir/tools/ttnn-standalone";
+        }
+
+        std::string soPathStr = compileCppToSo(
+            cpp_source, "/tmp/", metal_src_dir.string(), metal_lib_dir.string(), standalone_dir.string());
```
svuckovicTT, line 201: `/tmp/` hardcoded here, maybe we want something per-test? I played around with pytest's `tmp_path` but there were some slight issues I encountered that I can't recall.

svuckovicTT:
We also need to add data to superset DB - @vmilosevic @nsumrakTT let's sync on this tmr!

svuckovicTT:
@pilkicTT friendly ping

forge/csrc/CMakeLists.txt
```diff
@@ -11,7 +11,9 @@ set(TT_MLIR_ROOT_DIR ${CMAKE_SOURCE_DIR}/third_party/tt-mlir)
 set(TTMLIR_INCLUDE_DIRS
     ${TT_MLIR_ROOT_DIR}/include
     ${TT_MLIR_ROOT_DIR}/build/include
-    ${TT_MLIR_ROOT_DIR}/runtime/include)
+    ${TT_MLIR_ROOT_DIR}/runtime/test/include
```
svuckovicTT, line 14: Not needed anymore - remove

forge/forge/verify/verify.py
```diff
@@ -439,6 +440,36 @@ def verify(
...
 
+
+    if "TT_METAL_HOME" in os.environ:
+        print(f"metal: {os.environ['TT_METAL_HOME']}")
+    if "CPM_SOURCE_CACHE" in os.environ:
+        print(f"CPM: {os.environ['CPM_SOURCE_CACHE']}")
+
+
+    # Compile EmitC .so
+    so_path = compiled_model.export_to_shared_object()
```
svuckovicTT, line 451: We need a mechanism to exclude emitc from some tests - so I think we should add a bool in `VerifyConfig` to denote whether to run EmitC, and then have each model that pass set it to True (all but one currently). Additionally, a global flag to turn emitc on/off, that overrides the individual one, might be useful. Thoughts?

nsumrakTT (approved):

forge/forge/verify/verify.py
```diff
@@ -439,6 +438,29 @@ def verify(
...
         )
 
+    # Compile EmitC .so
+    so_path = compiled_model.export_to_shared_object()
+    # Run EmitC .so
+    all_outputs = compiled_model.runtime_model_state.get_outputs(ProgramType.Forward)
+    consts_and_params = compiled_model.runtime_model_state.get_persistent_inputs(ProgramType.Forward)
+    fwd_func_name = "forward"
+    fwd_func_name_len = len(fwd_func_name)
+    fwd_func_sym = f"_Z{fwd_func_name_len}{fwd_func_name}St6vectorIN2tt8tt_metal6TensorESaIS2_EE"
```
pilkicTT, line 448: why not create an `extern "C"` API in the shared library, which uses C arrays, i.e. `void*` - to avoid mangled symbols?

forge/forge/verify/verify.py
pilkicTT: yeah, we can add a flag in `VerifyConfig` and i think there's no need for a global override.

what i'm interesed in is what is the test duration overhead with this additional testing of all models?

forge/csrc/runtime/python_bindings.cpp
```diff
@@ -52,14 +52,31 @@ void RuntimeModule(py::module &m_runtime)
             "run_program",
             [](ModelState &self, ProgramType program_type, std::vector<tt::Tensor> &act_inputs)
             { self.run_program(program_type, act_inputs); })
+        .def(
+            "get_persistent_inputs",
```
pilkicTT, line 56: let's sync on this, but i don't like mixing actual user APIs with the ones needed for testing - i feel it can create confusion for the user

forge/forge/verify/verify.py
azecevicTT: We have an issue for this https://github.com/tenstorrent/tt-mlir/issues/2478 after a discussion I had with you after uplift failure. I've tinkered a bit with it, it should be doable. But have to agree that there is a zero reason for FE to know parts of the mangled name, as a workaround this should be placed at least in `tt-mlir` so it's transparent to the FEs.

forge/forge/verify/verify.py
svuckovicTT: Will do.

I did a rough estimate by comparing `main` vs `main + diffs`. There are some tests failing in main, but also some compilations failing in emitc path after I rebased, so the real diffs are going to be slightly bigger:
```
21m 38s vs 44m 09s
45m 29s vs 39m 29s  <--- strange
42m 19s vs 52m 28s
27m 14s vs 42m 46s
28m 53s vs 36m 26s
31m 53s vs 40m 07s
45m 50s vs 46m 16s
27m 08s vs 32m 03s
```

Runs:
- main: https://github.com/tenstorrent/tt-forge-fe/actions/runs/15098796775
- main + diffs: https://github.com/tenstorrent/tt-forge-fe/actions/runs/15099425714

I'd assume that the compilation of the dylibs is what's taking the most time. We're not closing the device between the canonical path and dylib, so I'm hoping that some form of caching is helping here - I'll have to investigate at some point.

Is this a concern at this point?

forge/forge/verify/verify.py
svuckovicTT: Completely agree!

I believe we shouldn't focus on this within this change, so many things are changing in tt-mlir that it feels like trying to hit a moving target already.

forge/csrc/runtime/python_bindings.cpp
```diff
@@ -54,14 +54,31 @@ void RuntimeModule(py::module &m_runtime)
...
             "get_outputs",
             [](ModelState &self, ProgramType program_type)
             {
                 std::optional<ProgramState> &opt_program_state = self.program_states[program_idx(program_type)];
                 TT_ASSERT(opt_program_state.has_value(), "Program state for {} not initialized", program_type);
                 return opt_program_state.value().outputs;
-            });
+            })
+        .def(
+            "test_so",
```
svuckovicTT, line 74: move to `runtime/test`

forge/csrc/runtime/runtime.hpp
```diff
@@ -19,4 +19,13 @@ std::vector<tt::Tensor> run_program_from_file(
...
 std::vector<tt::Tensor> run_program(runtime::Binary& binary, int program_idx, std::vector<tt::Tensor>& inputs);
 
+void* open_so(std::string path);
+void close_so(void* handle);
+std::vector<tt::runtime::Tensor> run_so_program(
+    void* so_handle,
+    std::string func_name,
+    std::vector<tt::Tensor>& inputs,
+    std::vector<tt::Tensor>& consts_and_params);
+bool compareOuts(std::vector<tt::runtime::Tensor>& lhs, std::vector<tt::runtime::Tensor>& rhs);
```
svuckovicTT, line 29: move to .cpp

pilkicTT (commented):
@svuckovicTT sorry for the delays... we're almost there...

forge/csrc/passes/mlir_compiler.cpp
```diff
@@ -119,6 +128,70 @@ auto run_mlir_compiler_generic(tt::ForgeGraphModule& module, const std::optional
...
+        if (FORGE_HOME == nullptr)
+        {
+            throw std::runtime_error("FORGE_HOME environment variable is not set.");
+        }
+
+        const fs::path in_wheel_path = fs::path(FORGE_HOME) / "forge/tt-metal";
+        const fs::path in_source_path =
+            fs::path(FORGE_HOME).parent_path() / "third_party/tt-mlir/third_party/tt-metal/src/tt-metal";
+
+        if (!fs::exists(in_wheel_path) && !fs::exists(in_source_path))
```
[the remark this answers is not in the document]
svuckovicTT: Added both `FORGE_IN_WHEEL` and `FORGE_IN_SOURCE` to be able to warn if someone already has `TT_METAL_HOME` set for other reasons.

forge/csrc/runtime/runtime.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include "tensor.hpp"
 #include "tt/runtime/runtime.h"
+#include "tt/runtime/test/ttnn/dylib.h"
```
pilkicTT, line 11: unneeded include

forge/csrc/runtime/state.cpp
```diff
@@ -4,8 +4,11 @@
 #include "runtime/state.hpp"
 
 #include <utils/logger.hpp>
+#include <vector>
 
+#include "runtime/runtime.hpp"
 #include "tt/runtime/runtime.h"
+#include "tt/runtime/types.h"
```
pilkicTT, line 11: unneeded includes

forge/forge/__init__.py
```diff
@@ -33,14 +33,21 @@ def set_home_paths():
...
         else:
             logger.error(
                 f"TT_METAL_HOME environment variable is not set. Tried setting it to {in_wheel_path} and {in_source_path}, but neither exists. Something is wrong with the installation."
             )
             exit(1)
 
+    if "FORGE_IN_WHEEL" not in os.environ and "FORGE_IN_SOURCE" not in os.environ:
+        logger.error(
+            "Neither FORGE_IN_WHEEL nor FORGE_IN_SOURCE environment variables are set. Please set one of them to indicate whether you are running from a wheel or from source."
+        )
```
pilkicTT, line 49: we should figure this out on our own and print out a warning only...

additionally, as it is right now, if `TT_METAL_HOME` is overriden by the user, neither of these flags will be set...

forge/csrc/runtime/testutils/python_bindings.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "runtime/state.hpp"
+#include "runtime/tensor.hpp"
+#include "runtime/testutils/runtime_testutils.hpp"
+
+namespace tt
+{
+
+void RuntimeTestUtilsModule(py::module &m_runtime_testutils)
+{
+    m_runtime_testutils
```
pilkicTT, line 16: can we move these bindings to `forge/csrc/runtime/python_bindings.cpp`? see the `experimental` module...

pilkicTT (approved):

PR #1934, opened 2025-05-06
Evaluate Benchmark Resnet on Dataset

vcanicTT:
Add support to run Resnet Benchmark model on real dataset, instead of random data.

### DIFF
--- a/.github/workflows/perf-benchmark-sub.yml
+++ b/.github/workflows/perf-benchmark-sub.yml
@@ -18,14 +18,14 @@ jobs:
       fail-fast: false
       matrix:
         build: [
-          { runs-on: "n150", name: "mnist_linear",             dir: "MNISTLinear",                  bs: 32, lp: 32, df: 'float32' },
-          { runs-on: "n150", name: "resnet50_hf",              dir: "ResNetForImageClassification", bs: 8,  lp: 32, df: 'bfloat16', allow-fail: true },
-          { runs-on: "n150", name: "llama",                    dir: "LlamaModel",                   bs: 1,  lp: 32, df: 'float32' },
-          { runs-on: "n150", name: "mobilenetv2_basic",        dir: "MobileNetv2Basic",             bs: 1,  lp: 32, df: 'float32' },
-          { runs-on: "n150", name: "efficientnet_timm",        dir: "EfficientNetTimmB0",           bs: 1,  lp: 32, df: 'float32' , allow-fail: true },
-          { runs-on: "n150", name: "segformer_classification", dir: "SegformerClassification",      bs: 1,  lp: 32, df: 'float32' },
-          { runs-on: "n150", name: "vit_base",                 dir: "ViTBase",                      bs: 1,  lp: 32, df: 'float32' },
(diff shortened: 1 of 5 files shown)


vcanicTT:
@pilkicTT @dgolubovicTT @nsumrakTT please, take a look!

.github/workflows/perf-benchmark-sub.yml
```diff
@@ -18,14 +18,14 @@ jobs:
...
-          { runs-on: "n150", name: "resnet50_hf",              dir: "ResNetForImageClassification", bs: 8,  lp: 32, df: 'bfloat16', allow-fail: true },
-          { runs-on: "n150", name: "llama",                    dir: "LlamaModel",                   bs: 1,  lp: 32, df: 'float32' },
-          { runs-on: "n150", name: "mobilenetv2_basic",        dir: "MobileNetv2Basic",             bs: 1,  lp: 32, df: 'float32' },
-          { runs-on: "n150", name: "efficientnet_timm",        dir: "EfficientNetTimmB0",           bs: 1,  lp: 32, df: 'float32' , allow-fail: true },
-          { runs-on: "n150", name: "segformer_classification", dir: "SegformerClassification",      bs: 1,  lp: 32, df: 'float32' },
-          { runs-on: "n150", name: "vit_base",                 dir: "ViTBase",                      bs: 1,  lp: 32, df: 'float32' },
-          { runs-on: "n150", name: "vovnet_osmr",              dir: "VovnetOSMR",                   bs: 1,  lp: 32, df: 'float32' , allow-fail: true }
+          { runs-on: "n150", name: "mnist_linear",             dir: "MNISTLinear",                  ts: 'na',             bs: 32, lp: 32, df: 'float32' },
```
nsumrakTT, line 21: It is possible to have default parameter, in case of "ts" it is obviously "na", so you can only add ts where it is different (in line with "ts: 'classification'"), just like allow-fail parameter. As number of parameters grow, it might give clearer view on test matrix.

nsumrakTT (approved):

.github/workflows/perf-benchmark-sub.yml
```diff
@@ -150,7 +150,7 @@ jobs:
       run: |
         source env/activate
         mkdir -p ${{ steps.strings.outputs.perf_report_path }}
-        python forge/test/benchmark/benchmark.py -m ${{ matrix.build.name }} -bs ${{ matrix.build.bs }} -df ${{ matrix.build.df }} -lp ${{ matrix.build.lp }} -o ${{ steps.strings.outputs.perf_report_path }}/benchmark_forge_e2e_${{ matrix.build.name }}_${{ matrix.build.bs }}_${{ matrix.build.lp }}_${{ steps.fetch-job-id.outputs.job_id }}.json
```
nsumrakTT, line 153: This line will make default parameter in shell variable ts:
``ts=if[ -z "${{matrix.build.ts}}"]; then echo "na" else echo "${{matrix.build.ts}}" fi``
And then instead of ${{matrix.build.ts}} use just $ts

PR #1940, opened 2025-05-06
Disable device perf

vcanicTT:
Disable device perf, until we solve the CI issue.


vcanicTT:
@nsumrakTT please, take a look!

nsumrakTT (approved):

PR #1955, opened 2025-05-07
Add benchmark codeowners

odjuricicTT:
### Problem description
I want to be notified on benchmark test changes.

### What's changed
Add @vcanicTT @odjuricicTT to benchmark codeowners.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,3 +6,6 @@
 
 # Models
 forge/test/models @nvukobratTT @pilkicTT @dgolubovicTT @vkovinicTT @mstojkovicTT
+
+# Benchmarks
+forge/test/benchmark @vcanicTT @odjuricicTT @pilkicTT @nvukobratTT @dgolubovicTT


nsumrakTT (approved):
It seems to me a good model to prevent often changes to CODEOWNERS is what is proposed in tt-mlir. Take a look:
https://github.com/tenstorrent/tt-mlir/pull/3160

nvukobratTT:
> It seems to me a good model to prevent often changes to CODEOWNERS is what is proposed in tt-mlir. Take a look: [tenstorrent/tt-mlir#3160](https://github.com/tenstorrent/tt-mlir/pull/3160)

I like this idea; I'm just not sure if it's overkill for frontends! For example, we'll need subgroups for each FE and groups for performance (some from the optimizer, some from other MLIR groups, etc.) in order to properly handle these.

nsumrakTT:
@nvukobratTT From my point of view, it is the difference between changing CODEOWNERS through repo change (and PR, build, review), or clicking through GH interface. Anyway, I mentioned that just as FYI

PR #1962, opened 2025-05-08
Added data format for performance benchmark tests.

vcanicTT:
Added data format for performance benchmark tests.

### DIFF
--- a/forge/test/benchmark/benchmark/models/efficientnet_timm.py
+++ b/forge/test/benchmark/benchmark/models/efficientnet_timm.py
@@ -19,8 +19,10 @@ from tqdm import tqdm
 import forge
 from forge.verify.value_checkers import AutomaticValueChecker
 from forge.verify.verify import verify
+from forge.verify.config import VerifyConfig
 from forge._C.runtime.experimental import configure_devices, DeviceSettings
 from forge.config import CompilerConfig, MLIRConfig
+from forge._C import DataFormat
 
 from test.utils import download_model
 from test.benchmark.utils import load_benchmark_dataset, evaluate_classification
@@ -37,6 +39,11 @@ TASK = [
 # Batch size configurations
 BATCH_SIZE = [
     1,
+]
+
+# Data format configurations
+DATA_FORMAT = [
+    "bfloat16",
 ]
 
 # Input size configurations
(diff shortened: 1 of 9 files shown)


vcanicTT:
@dgolubovicTT Let me know, when you finish your feature for data formats, so we can align the benchmarks.

nsumrakTT (approved):

forge/test/benchmark/benchmark/models/efficientnet_timm.py
```diff
@@ -31,6 +33,11 @@
     1,
 ]
 
+# Data format configurations
+DATA_FORMAT = [
+    "bfloat16",
```
odjuricicTT, line 38: Why do we need this both here and into the workflow?

vcanicTT: In CI in workflow we use only the best combination

forge/test/benchmark/benchmark/models/vovnet.py
```diff
@@ -99,14 +116,25 @@ def test_vovnet_osmr(
...
     # Run for the first time to warm up the model, it will be done by verify function.
     # This is required to get accurate performance numbers.
-    verify(input_sample, framework_model, compiled_model)
+    pcc = 0.99
+    verify_cfg = VerifyConfig()
+    if data_format == "bfloat16":
+        # Set smaller pcc for bfloat16
+        pcc = 0.97
+    verify_cfg.value_checker = AutomaticValueChecker(pcc=pcc)
+    verify(
```
odjuricicTT, line 125: Can we verify PCC after we measure inference performance? It would be really usefull to have the test run inference and print the result with samples per second and only then fail if PCC is bad.

vcanicTT: I know, it's useful for perf debugging, we can add option, parameter, to run without verification, but I think functionality always should be before performance. Let me to think about this, separately.

forge/test/benchmark/benchmark/models/efficientnet_timm.py
```diff
@@ -62,16 +70,25 @@ def test_efficientnet_timm(training, batch_size, input_size, channel_size, loop_
...
+        # Convert model to bfloat16
+        framework_model = framework_model.to(torch.bfloat16)
     framework_model.eval()
     fw_out = framework_model(*input_sample)
 
     # Compiler configuration
     compiler_config = CompilerConfig()
     # Turn on MLIR optimizations.
     compiler_config.mlir_config = MLIRConfig().set_enable_consteval(True).set_enable_optimizer(True)
+    if data_format == "bfloat16":
```
dgolubovicTT, line 89: Nitpick, it would be cleaner if you set everything needed for bfloat16 in one if statement.

dgolubovicTT (approved):

forge/test/benchmark/benchmark/models/vovnet.py
odjuricicTT: Let's discuss if needed. The idea being that it makes a smoother development process if we can at least see perf before needing to debug PCC issues.

PR #1972, opened 2025-05-08
Handle None min/max in Clip op properly in tvm

kamalrajkannan78:
## summary 

This PR tracks [this](https://github.com/tenstorrent/tt-tvm/pull/80) fix and adds sanity to verify the fix.


saiarthiraguram (approved):

forge/test/test_model_parts.py
```diff
@@ -54,3 +56,46 @@ def forward(self, x):
...
+
+    # Export to ONNX
+    torch.onnx.export(model, (inputs[0],), "temp.onnx", opset_version=17)
+
+    # Load and check ONNX
+    onnx_model = onnx.load("temp.onnx")
+    onnx.checker.check_model(onnx_model)
+
+    compiled_model = forge.compile(
+        onnx_model, inputs, verify_cfg=DepricatedVerifyConfig(verify_forge_codegen_vs_framework=True)
```
nvukobratTT, line 99: Do we need to verify codegen with this fix? If there is an issues, last verify should pick it up

forge/test/test_model_parts.py
```diff
@@ -54,3 +56,46 @@ def forward(self, x):
         forge_property_handler=forge_property_recorder,
     )
     verify(inputs, framework_model, compiled_model, forge_property_handler=forge_property_recorder)
+
+
+@pytest.mark.parametrize(
+    "input_shape, clamp_min, clamp_max",
+    [
+        ((64, 3, 64, 64), None, math.log(100.0)),
```
nvukobratTT, line 64: Should we add int test cases? I see 5.0 and -103.0 test cases, but these are considered as float.

kamalrajkannan78: int test cases are added now. for this cases pcc drop observed at runtime. I confirmed that issue is from backend by enabling verify_forge_codegen_vs_framework on local testing. please have a look on updated sanity logs [here](https://github.com/tenstorrent/tt-tvm/pull/80#issue-3049531393)  for reference.

nvukobratTT: Thanks @kamalrajkannan78 , let's leave them as xfail for now.

nvukobratTT (approved):

PR #1983, opened 2025-05-09
Improve lower data formats logic: Adding less cast ops and some bug fixes

### LINKED ISSUE #1982: Improve data formats override logic
Currently, logic for setting lower data format override adds some redundant casts. 
Also there are few bugs throughout the stack related to data formats. Namely, there are many points in code where tensors are cast manually (mainly in op eval).

dgolubovicTT:
### Ticket
Closes #1982 

### Problem description
Lower data format logic uses casts to ensure inputs to graph are in desired data format. However, some unnecessary casts were being added. Also, there were some bugs in logic for adding casts. Some data format bugs throughout the stack are also tackled in this PR. 

### What's changed
This PR adds:

    - [Fewer Cast Nodes] Insert a cast node on every constant node whose data format differs from the df_override.
    - [Configure Output Data Formats] Set the output data format of all nodes that need it to match df_override.
    
    Bug Fixes:
    - Improve to_torch_operands: if operands use different floating-point dtypes, cast all to the highest precision.
    - Conv2d.eval: avoid casting the output of the evaluation function to float32.
    - Eltwise Binary: remove unnecessary operand casting.
    
    Tests:
    - Add tests that reproduced the aforementioned bugs in forge/test/mlir/test_dataformats.py.
    - Enable lower data format support in YOLOv8 and YOLOv10, and remove related xfails.

This simplifies specifying lower data formats in tests.

Few side changes are also added:
- bugfix: while iterating through graph nodes and adding constants in loop, there was a segmentation fault hit. This was because we used range based loop `(for (Node* node : graph->nodes()))` where `graph->nodes()` returned a reference to vector of nodes. Inside loop we added new nodes to that vector. If that vector should be reallocated due to expanding, iterators in for loop wouldn't be valid anymore. Solution - iterate using vector size before adding any new nodes. Nodes are added to the back of the vector so this doesn't cause any problem.

### DIFF
--- a/forge/csrc/passes/dataformat.cpp
+++ b/forge/csrc/passes/dataformat.cpp
@@ -230,18 +230,20 @@ void lower_fallback_data_formats(graphlib::Graph *graph, DataFormat fp32_fallbac
     }
 }
 
-// Inserts a cast node after every input node in the graph,
+// Inserts a cast node after every constant input node in the graph,
 // placing it on the first user edge and re-routing all other edges.
 // The cast node will convert outputs to the given DataFormat.
-static void insert_cast_on_input_nodes(graphlib::Graph *graph, DataFormat df_override)
+static void insert_cast_on_const_input_nodes(graphlib::Graph *graph, DataFormat df_override)
 {
-    for (Node *node : graph->nodes())
+    for (Node *node : graph->nodes_by_type(graphlib::NodeType::kInput))
     {
-        if (node->node_type() != graphlib::NodeType::kInput)
-            continue;
+        InputNode *input_node = node->as<graphlib::InputNode>();
 
         std::vector<Edge> user_edges = graph->user_data_edges(node);
         if (user_edges.empty())
+            continue;
+
+        if (input_node->input_type() != graphlib::InputNodeType::Constant || node->output_df() == df_override)
             continue;
 
         // Generate cast node
@@ -256,10 +258,10 @@ static void insert_cast_on_input_nodes(graphlib::Graph *graph, DataFormat df_ove
             graph->get_subgraph_id_for_node(node->id()));
 
         cast_node->set_shape(node->shape());
-        cast_node->set_output_df(df_override);
 
         // First edge: insert cast node on it.
         graphlib::insert_node_on_edge(graph, user_edges[0], cast_node);
+        cast_node->set_output_df(df_override);
         TT_ASSERT(cast_node != nullptr, "Cast node should not be null");
 
         // Remaining edges: reconnect edges to cast node.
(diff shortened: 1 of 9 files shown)


vcanicTT (approved):

forge/csrc/passes/dataformat.cpp
```diff
@@ -286,7 +288,9 @@ void configure_output_data_formats(graphlib::Graph *graph, std::optional<DataFor
     for (Node *node : graph->nodes())
     {
         bool disallow_default_override = is_integer_data_format(node->output_df());
-        if (default_df_override and not disallow_default_override)
+        bool node_not_input = node->node_type() != graphlib::NodeType::kInput;
+        // Inputs come from host as they are. We shouldn't override their data format.
+        if (default_df_override and not disallow_default_override and node_not_input)
```
pilkicTT, line 293: please use symbols for logical operators, i.e. `&&`, `!`, etc.

forge/forge/verify/verify.py
```diff
@@ -251,7 +251,7 @@ def verify_golden(
     assert False  # Run ttnn golden
 
 
-def check_dtypes(fw_dtype: torch.dtype, co_dtype: torch.dtype):
+def check_dtypes(fw_dtype: torch.dtype, co_dtype: torch.dtype, expected_output_data_format: DataFormat = None):
```
pilkicTT, line 254: i would prefer not to add complexity to our verification methods

i can see that this feature is useful for the user who doesn't want to convert the original model to the lower precision data format, but in our test cases we actually want to compare apples to apples, so i don't see why we wouldn't cast the original model in our tests?

pilkicTT (changes requested):

forge/csrc/passes/dataformat.cpp
dgolubovicTT: ofc

forge/forge/verify/verify.py
nvukobratTT: Let's discuss this topic next week, with a focus on outlining how we want to handle different data format use cases. More precisely, how set of these features should be structured going forward.

One additional topic that came to me in favor of doing model casting as well is related to our op generation logic.  More precisely, if the model is left in FP32, and we use a flag to do conversion with additional verification updates, it'll make our life much harder to keep generated model ops correct. I.e. if the model is set to the desired data format, the current logic will just pick it up and generate ops in the valid data format. If that isn't the case, we'll have to add special handling when this flag is enabled during op generation. I assume this will propagate to similar tools as well. 

In sum, let's sit down and discuss these details next week.

nvukobratTT (approved):

pilkicTT (approved):

forge/csrc/passes/dataformat.cpp
```diff
@@ -230,20 +230,26 @@ void lower_fallback_data_formats(graphlib::Graph *graph, DataFormat fp32_fallbac
...
-// Inserts a cast node after every input node in the graph,
+// Inserts a cast node after every constant input node in the graph,
 // placing it on the first user edge and re-routing all other edges.
 // The cast node will convert outputs to the given DataFormat.
-static void insert_cast_on_input_nodes(graphlib::Graph *graph, DataFormat df_override)
+static void insert_cast_on_const_input_nodes(graphlib::Graph *graph, DataFormat df_override)
 {
-    for (Node *node : graph->nodes())
+    size_t original_size = graph->nodes().size();
+    for (size_t i = 0; i < original_size; ++i)
```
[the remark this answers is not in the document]
dgolubovicTT: Nice, haven't noticed `nodes_by_type `.

forge/csrc/passes/dataformat.cpp
```diff
@@ -286,7 +292,13 @@ void configure_output_data_formats(graphlib::Graph *graph, std::optional<DataFor
...
     {
         bool disallow_default_override = is_integer_data_format(node->output_df());
-        if (default_df_override and not disallow_default_override)
+        bool node_not_input = node->node_type() != graphlib::NodeType::kInput;
+        bool node_input_parameter = node->node_type() == graphlib::NodeType::kInput &&
+                                    node->as<graphlib::InputNode>()->input_type() == graphlib::InputNodeType::Parameter;
+        // We should override output data format for input parameters because TVM casts them to the default data format
+        // float32. Other types of input nodes like activations and constants are accurately represented in Forge graph
+        // (dataformat corresponding to what will come from the host).
+        if (default_df_override && !disallow_default_override && (node_not_input || node_input_parameter))
```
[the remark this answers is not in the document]
dgolubovicTT: Not all constant input will satisfy this constraint directly. TVM constants will have df float32 but will have cast after them. Other constants should have the same data format as averride...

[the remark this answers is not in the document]
dgolubovicTT: Implemented the check

PR #1993, opened 2025-05-12
Add support for image.resize1d op

### LINKED ISSUE #1997: [ops] image.resize1d - unsupported op

meenakshiramanathan1:
Fixes #1997


saiarthiraguram (approved):

nvukobratTT (commented):
Few general questions:
- Which model is affected?
- Isn't it possible to map resize1d to upsample/downsample 2d case? 

My concern is that we're doing decomposition into 20 or more ops, which will kill performance down the road. Maybe it makes more sense to push for proper support on Metal side. 

Let's get the details, and discuss further on this change.

PR #2000, opened 2025-05-13
Add save function for a flatbuffer file of compiled model

### LINKED ISSUE #1788: Compile to flatbuffer file
In order to be able to use some forge tools (e.g. ttrt, explorer) to analyze performance properly we need the compilation result to be in flatbuffer format.

sott0n:
### Ticket
Closes #1788 

### Problem description
There is currently no mechanism to save the compiled FlatBuffer, making it impossible to reuse the generated data with tools such as tt-explorer and ttrt.

### What's changed
Add a mechanism to save the compiled FlatBuffer to a file from CompiledModel.

### Checklist

### DIFF
--- a/forge/test/test_api.py
+++ b/forge/test/test_api.py
@@ -13,7 +13,6 @@ import forge
 from forge.config import CompilerConfig, MLIRConfig
 from forge.tensor import to_forge_tensors, to_pt_tensors
 from forge.verify.value_checkers import AutomaticValueChecker
-from forge._C.runtime import load_binary_from_file
 
 
 @pytest.mark.push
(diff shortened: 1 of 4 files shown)


nvukobratTT (commented):
Can you add some sanity for it? 

We'll need it to make sure that this feature isn't broken by some other change in the future. You can add it here for example: `forge/test/test_api.py`

sott0n:
@nvukobratTT Added a test.

nvukobratTT (approved):
Looks great! Thanks for pushing this one @sott0n 🙌

forge/test/test_api.py
```diff
@@ -94,6 +95,35 @@ def forward(self, x, y):
...
+    inputs = [torch.rand(shape), torch.rand(shape)]
+
+    compiled_model = forge.compile(model, sample_inputs=inputs)
+
+    file_path = "generated_add.ttnn"
+    compiled_model.save(file_path)
+
+    assert os.path.exists(file_path)
+    load_bin = load_binary_from_file(file_path)
+    assert load_bin.get_file_identifier() == "TTNN"
```
pilkicTT, line 118: I would prefer not to expose these functions just for the sake of testing, since these functions don't provide any value to the user (for the time being) - just to keep things clear...

pilkicTT (approved):

forge/test/test_api.py
sott0n: Thanks for your comment, I removed these functions.

PR #2026, opened 2025-05-14
Add pytest fixture to remove tmp_path directory after test execution

chandrasekaranpradeep:
1. Currently pytest built-in **tmp_path** fixture was added to each ONNX models tests to generate a unique temporary directory for saving ONNX files, but [tmp_path fixture](https://github.com/pytest-dev/pytest/blob/017dd1ccd60055bc6a7a41af11b74425da87a696/src/_pytest/tmpdir.py#L239C1-L253C46) leaves these directories intact by default, causing outdated artifacts to pile up. This PR introduces a new `forge_tmp_path` fixture that wraps `tmp_path` and automatically removes its directory (and all contents) immediately after each test completes—ensuring no stale ONNX files remain. 
2. Replaced `tmp_path` with  `forge_tmp_path` pytest fixture in the ONNX models tests.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -6,6 +6,7 @@ import gc
 import time
 import pytest
 import psutil
+import shutil
 from loguru import logger
 from datetime import datetime
 from forge.forge_property_utils import ForgePropertyHandler, ForgePropertyStore, ExecutionStage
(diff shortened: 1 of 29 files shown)


conftest.py
```diff
@@ -6,6 +6,8 @@
 import time
 import pytest
 import psutil
+import threading
```
nvukobratTT, line 9: Needed?

nvukobratTT (approved):

conftest.py
chandrasekaranpradeep: Not needed. Removed it

PR #2058, opened 2025-05-16
Removed forge_property_handler from cpp code.

acolicTT:
Removed forge_property_handler form cpp code.

ExecutionDepth (tests info) was used as a compiler substage execution information and it was implemented in cpp code. It is moved to python.

Since we don't need to separate generating TTIR and MLIR compiler pass, both are saved as a single FAILED_TTMLIR_COMPILATION execution depth.

Also, getting flatbuffer logic is simplified and it is moved to python code.

This is second part of forge property handler simplification.


acolicTT:
cc @chandrasekaranpradeep @ashokkumarkannan1 @kamalrajkannan78 @meenakshiramanathan1 @saiarthiraguram

conftest.py
```diff
@@ -69,8 +68,7 @@ def forge_property_recorder(request, record_property):
     # Create a handler that uses the property store; the handler is responsible for recording and managing property details.
     forge_property_handler = ForgePropertyHandler(forge_property_store)
 
-    # Set CI_FAILURE as default execution depth and FAILED_BEFORE_FORGE_COMPILATION_INITIATION as default execution stage
-    forge_property_handler.record_execution_depth(ExecutionDepth.CI_FAILURE)
+    # Set FAILED_BEFORE_FORGE_COMPILATION_INITIATION as default execution stage
```
pilkicTT, line 71: shouldn't this remain `CI_FAILURE`?

chandrasekaranpradeep:
> cc @chandrasekaranpradeep @ashokkumarkannan1 @kamalrajkannan78 @meenakshiramanathan1 @saiarthiraguram

Thanks for pushing this
It looks good to me

PR #2076, opened 2025-05-19
Moving data format override pass after decompose, and debug decompose 

### LINKED ISSUE #2075: Dataformat override pass is before Decompose
Since dataformat override pass is before decompose pass, constants added in decompose pass aren't casted to lower data format.
This is caught first in yolo_v9 but it is the case in many models.

To avoid doing data format override pass again after decompose, we should move data format pass after decompose. The change should be minimal since both are in post initial pass.

dgolubovicTT:
### Ticket
Closes #2075 

### Problem description
Since dataformat override pass is before decompose pass, constants added in decompose pass aren't casted to lower data format.
This is caught first in yolo_v9 but it is the case in many models.

### What's changed

1. To avoid doing data format override pass again after decompose, moved data format pass after decompose. The change is minimal since both passes are in post initial graph pass.
2. Bugfix : When adding new constant node in decomposing context, set its output data format based on pytorch testor dtype, not its consumer output dataformat.

### Checklist
- Added yolo_v9 test and enable bfloat16 in the test

### DIFF
--- a/forge/csrc/forge_passes.cpp
+++ b/forge/csrc/forge_passes.cpp
@@ -80,7 +80,6 @@ run_post_initial_graph_passes(
     std::shared_ptr<void> compiler_cfg = make_shared_py_object(compiler_cfg_object);
 
     passes::print_graph(graph, "INITIAL");
-    passes::apply_user_data_format_override(graph, compiler_cfg_object);
     passes::generate_initial_flops_estimate(graph);
     passes::decompose_nd_reshape_split(graph);
     passes::erase_unnecessary_4d_tm_sequence(graph);
@@ -90,6 +89,7 @@ run_post_initial_graph_passes(
 
     auto inserted_node_id_mapping = decompose_tt_forge_graph(graph, "get_f_forge_decompose", compiler_cfg);
     auto chip_id_assignments = passes::fracture(graph, fracture_groups);
+    passes::apply_user_data_format_override(graph, compiler_cfg_object);
     return std::make_tuple(inserted_node_id_mapping, chip_id_assignments);
 }
 
--- a/forge/csrc/passes/decomposing_context.cpp
+++ b/forge/csrc/passes/decomposing_context.cpp
@@ -161,15 +161,37 @@ void DecomposingContext::fuse(NodeContext operand, graphlib::PortId producer_out
     }
 }
 
-NodeContext DecomposingContext::tensor(std::shared_ptr<void> tensor, graphlib::Shape tensor_shape, DataFormat df)
+NodeContext DecomposingContext::tensor(std::shared_ptr<void> tensor, graphlib::Shape tensor_shape)
 {
+    TT_ASSERT(tensor, "Trying to create constant tensor node with null tensor");
+    // Import and call forge.tensor.pytorch_dtype_to_forge_dataformat
+    py::module_ tensor_module = py::module_::import("forge.tensor");
+
     auto new_node = graph->add_node(
         graphlib::create_node<graphlib::ConstantInputNode>(
             "dc.input_tensor." + this->node_->name() + "." + std::to_string(this->op_index), tensor, tensor_shape),
         subgraph_idx);
+
     new_node->set_shape(tensor_shape);
 
-    new_node->set_output_df((df != DataFormat::Invalid) ? df : this->node_->output_df());
+    // Try to extract torch tensor dtype and map to DataFormat
+    DataFormat output_df;
+
+    py::object py_tensor = borrow_shared_py_object(tensor);
+    py::object tensor_dtype = py_tensor.attr("dtype");
+
+    try
+    {
+        output_df = py::cast<DataFormat>(tensor_module.attr("pytorch_dtype_to_forge_dataformat")(tensor_dtype));
+    }
+    catch (py::error_already_set &e)
+    {
+        throw std::runtime_error(
+            "Encountered python error while dtype to DataFormat mapping: " + std::string(e.what()));
+    }
+
+    new_node->set_output_df(output_df);
+
     new_node->set_epoch_type(this->node_->get_epoch_type());
     this->op_index++;
 
--- a/forge/test/models/pytorch/vision/yolo/test_yolo_v9.py
+++ b/forge/test/models/pytorch/vision/yolo/test_yolo_v9.py
@@ -1,0 +1,52 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+
+
+import pytest
+import torch
+
+import forge
+from forge._C import DataFormat
+from forge.config import CompilerConfig
+from forge.forge_property_utils import Framework, Source, Task
+from forge.verify.verify import verify
+
+from test.models.pytorch.vision.yolo.model_utils.yolo_utils import (
+    YoloWrapper,
+    load_yolo_model_and_image,
+)
+
+
+@pytest.mark.nightly
+def test_yolov9(forge_property_recorder):
+    # Record Forge Property
+    module_name = forge_property_recorder.record_model_properties(
+        framework=Framework.PYTORCH,
+        model="Yolov9",
+        variant="default",
+        task=Task.OBJECT_DETECTION,
+        source=Source.GITHUB,
+    )
+
+    # Load  model and input
+    model, image_tensor = load_yolo_model_and_image(
+        "https://github.com/ultralytics/assets/releases/download/v8.3.0/yolov9c.pt"
+    )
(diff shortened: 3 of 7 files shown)


forge/csrc/forge_passes.cpp
```diff
@@ -90,6 +89,7 @@ run_post_initial_graph_passes(
 
     auto inserted_node_id_mapping = decompose_tt_forge_graph(graph, "get_f_forge_decompose", compiler_cfg);
     auto chip_id_assignments = passes::fracture(graph, fracture_groups);
+    passes::apply_user_data_format_override(graph, compiler_cfg_object);
```
nvukobratTT, line 92: Will moving after decomposition solve the issue? 

We also have decomposition after autograd and optimization passes (not on every op, but there are multiple cases). More precisely:
- decompose_post_optimize
- decompose_post_autograd

forge/csrc/passes/decomposing_context.cpp
```diff
@@ -161,15 +161,51 @@ void DecomposingContext::fuse(NodeContext operand, graphlib::PortId producer_out
...
+        }
+        else
+        {
+            // If tensor is null, set to default DataFormat
+            final_df = this->node_->output_df();
+        }
+    }
+
+    new_node->set_output_df(final_df);
+    std::cout << "********  final_df: " << final_df << std::endl;
```
nvukobratTT, line 206: Redundant, or trace log

forge/csrc/passes/decomposing_context.cpp
```diff
@@ -161,15 +161,51 @@ void DecomposingContext::fuse(NodeContext operand, graphlib::PortId producer_out
...
+        {
+            py::object py_tensor = borrow_shared_py_object(tensor);
+            py::object tensor_dtype = py_tensor.attr("dtype");
+
+            try
+            {
+                // Import and call forge.tensor.pytorch_dtype_to_forge_dataformat
+                py::module_ tensor_module = py::module_::import("forge.tensor");
+                final_df = py::cast<DataFormat>(tensor_module.attr("pytorch_dtype_to_forge_dataformat")(tensor_dtype));
+                std::cout << " tensor dtype: " << py::str(tensor_dtype) << std::endl;
```
nvukobratTT, line 190: Redundant, or trace log

forge/test/models/pytorch/vision/yolo/test_yolo_v9.py
```diff
@@ -0,0 +1,54 @@
...
+def test_yolov9(forge_property_recorder):
+    # Record Forge Property
+    module_name = forge_property_recorder.record_model_properties(
+        framework=Framework.PYTORCH,
+        model="Yolov9",
+        variant="default",
+        task=Task.OBJECT_DETECTION,
+        source=Source.GITHUB,
+    )
+    forge_property_recorder.record_group("red")
```
nvukobratTT, line 31: Let's add only compiler-focused models from the Red Team to the "red" group. As this one is not directly set for us, we should remove it from the red. 

@staylorTT as this model is recently added, can we expand it to the Forge as well? It'll work e2e and we're planing to push perf for it as well. If yes, @dgolubovicTT let's keep it as red :))

forge/csrc/passes/decomposing_context.cpp
dgolubovicTT: redundant

forge/csrc/passes/decomposing_context.cpp
dgolubovicTT: redundant

forge/csrc/forge_passes.cpp
dgolubovicTT: Main issue is solved in `decomposing_context.cpp` by not setting constant node output df to consumer output df. Which was bug. Now constant node df is set to dataformat that corresponds to dtype of torch tensor we pass from decompose in python. This change is enough, and whole model passes, but since dataformat override pass is before decompose, constant stays in f32. 

Since decompose is in the same pass as dataformat override pass - post initial graph pass, i moved dataformat override pass after decompose so that constants added in decompose can additionaly be casted.

To answer your question about decompose post optimize and post autograd. Since they all call same decompose function, constant nodes added in these passes will be in sync with runtime pytorch tensors. However, they will be float32 while the resto of the model will be in lower precision if dataformat override is set (due to dataformat pass being in post initial graph pass)

So main issue is not solved by moving dataformat pass after decompose, but solving bug in decompose

dgolubovicTT: The PR name is misleading

dgolubovicTT: Let's sync offline

nvukobratTT (approved):
Synced offline. 

We need to consider a few more cases before deciding the expected path through the compiler. This will be defined in the follow-up issue.

forge/csrc/forge_passes.cpp
pilkicTT: i am not sure i agree with this... with the decompose fix we are actually fixing a mismatch between actual tensor data format and the expected format in the graph, but not the "df override scenario" where we want everything to be bfp16...

so, for the first case i think the fix in decompose is ok.

for the "df override scenario", just moving the `apply_user_data_format_override()` doesn't seem right. we can do it as a quick workaround, but we should work on a more robust solution - possibly the one we talked about offline...

forge/csrc/passes/decomposing_context.cpp
```diff
@@ -161,15 +161,49 @@ void DecomposingContext::fuse(NodeContext operand, graphlib::PortId producer_out
...
+            }
+        }
+        else
+        {
+            // If tensor is null, set to default DataFormat
+            final_df = this->node_->output_df();
+        }
+    }
+
+    new_node->set_output_df(final_df);
```
pilkicTT, line 204: are there many cases where we specify data format, i.e. `df_fallback`? i feel this should be also fixed, since there can be a mismatch between the specified `df_fallback` and the actual data format of the tensor...

forge/csrc/passes/decomposing_context.cpp
```diff
@@ -161,15 +161,49 @@ void DecomposingContext::fuse(NodeContext operand, graphlib::PortId producer_out
...
+    {
+        if (tensor)
+        {
+            py::object py_tensor = borrow_shared_py_object(tensor);
+            py::object tensor_dtype = py_tensor.attr("dtype");
+
+            try
+            {
+                // Import and call forge.tensor.pytorch_dtype_to_forge_dataformat
+                py::module_ tensor_module = py::module_::import("forge.tensor");
```
pilkicTT, line 188: please make this a static var and move it outside of the loop

forge/csrc/passes/decomposing_context.cpp
```diff
@@ -161,15 +161,49 @@ void DecomposingContext::fuse(NodeContext operand, graphlib::PortId producer_out
     }
 }
 
-NodeContext DecomposingContext::tensor(std::shared_ptr<void> tensor, graphlib::Shape tensor_shape, DataFormat df)
+NodeContext DecomposingContext::tensor(
+    std::shared_ptr<void> tensor, graphlib::Shape tensor_shape, DataFormat df_fallback)
 {
+    // Create the new ConstantInputNode
```
pilkicTT, line 167: unneeded comment

pilkicTT (changes requested):

forge/csrc/passes/decomposing_context.cpp
dgolubovicTT: it's not inside the loop, but you are right for static

forge/csrc/passes/decomposing_context.cpp
dgolubovicTT: Actually, there is only one place where tensor is called with `df_fallback ` in stack. Maybe this parameter does more harm than good.I didn't challenge its existence, but now it seems we should remove it.

forge/csrc/passes/decomposing_context.cpp
```diff
@@ -161,15 +161,49 @@ void DecomposingContext::fuse(NodeContext operand, graphlib::PortId producer_out
...
+            catch (py::error_already_set &e)
+            {
+                throw std::runtime_error(
+                    "Encountered python error while dtype to DataFormat mapping: " + std::string(e.what()));
+            }
+        }
+        else
+        {
+            // If tensor is null, set to default DataFormat
+            final_df = this->node_->output_df();
```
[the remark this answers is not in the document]
dgolubovicTT: Adding assert is they way to go

forge/csrc/passes/decomposing_context.cpp
pilkicTT: yeah, i meant outside of the logic... :smile:

pilkicTT (approved):

PR #2086, opened 2025-05-20
Enable optimizer and consteval for segformer performance benchmark.

vcanicTT:
Enable optimizer and consteval for segformer performance benchmark.

### DIFF
--- a/forge/test/benchmark/benchmark/models/segformer.py
+++ b/forge/test/benchmark/benchmark/models/segformer.py
@@ -110,7 +110,7 @@ def test_segformer(
     compiler_config = CompilerConfig()
     # @TODO - For now, we are skipping enabling MLIR optimizations, because it is not working with the current version of the model.
     # Turn on MLIR optimizations.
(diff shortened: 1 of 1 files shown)


forge/test/benchmark/benchmark/models/segformer.py
```diff
@@ -110,7 +110,7 @@ def test_segformer(
     compiler_config = CompilerConfig()
     # @TODO - For now, we are skipping enabling MLIR optimizations, because it is not working with the current version of the model.
     # Turn on MLIR optimizations.
-    # compiler_config.mlir_config = MLIRConfig().set_enable_consteval(True).set_enable_optimizer(True)
+    compiler_config.mlir_config = MLIRConfig().set_enable_consteval(True).set_enable_optimizer(True)
```
odjuricicTT, line 113: Consteval is on by default in mlir as of recently so its not needed here.

odjuricicTT (approved):

forge/test/benchmark/benchmark/models/segformer.py
vcanicTT: Removed.

PR #2102, opened 2025-05-21
[ci] add blackhole p150 tests

gfengTT:
[ci] Add test mark fails_on_bh

These tests will be skipped when running on p150.

### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist

### DIFF
--- a/.github/workflows/on-nightly.yml
+++ b/.github/workflows/on-nightly.yml
@@ -91,6 +91,22 @@ jobs:
       runs-on: '[{"runs-on": "n150"}]'
       run_id: ${{ needs.build.outputs.run_id }}
 
+  test_full_model_passing_p150:
+    if: success() || failure()
+    needs:
+      - set-inputs
+      - docker-build
+      - build
+    uses: ./.github/workflows/test-sub.yml
+    secrets: inherit
+    with:
+      test_mark: 'nightly and not xfail'
+      test_group_cnt: ${{ needs.set-inputs.outputs.test_group_cnt }}
+      test_group_ids: ${{ needs.set-inputs.outputs.test_group_ids }}
+      docker-image: ${{ needs.docker-build.outputs.docker-image }}
+      runs-on: '[{"runs-on": "p150"}]'
+      run_id: ${{ needs.build.outputs.run_id }}
+
   perf-benchmark:
     if: success() || failure()
     needs:
@@ -119,6 +135,22 @@ jobs:
       run_id: ${{ needs.build.outputs.run_id }}
       runs-on: '[{"runs-on": "n150"}]'
 
+  test_full_model_xfailing_p150:
+    if: success() || failure()
+    needs:
+      - docker-build
+      - set-inputs
+      - build
+    uses: ./.github/workflows/test-sub.yml
+    secrets: inherit
+    with:
+      test_mark: 'nightly and xfail'
+      test_group_cnt: ${{ needs.set-inputs.outputs.test_group_cnt }}
+      test_group_ids: ${{ needs.set-inputs.outputs.test_group_ids }}
+      docker-image: ${{ needs.docker-build.outputs.docker-image }}
+      run_id: ${{ needs.build.outputs.run_id }}
+      runs-on: '[{"runs-on": "p150"}]'
+
   test_models_ops:
     if: ${{ needs.set-inputs.outputs.run_ops_sweeps }}
     needs:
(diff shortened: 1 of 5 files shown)


.github/workflows/test-sub.yml
```diff
@@ -156,13 +156,25 @@ jobs:
...
         run: |
           source env/activate
-          pytest --splits ${{ inputs.test_group_cnt }} \
-                 --group ${{ matrix.test_group_id }} \
-                 --splitting-algorithm least_duration \
-                 -m "${{ inputs.test_mark }}" \
-                 --junit-xml=${{ steps.strings.outputs.test_report_path }} \
-                 --log-memory-usage \
-                 2>&1 | tee pytest.log
+          # Skip tests marked as fails_on_bh when running on p150
```
nvukobratTT, line 159: I would still advise you to have a separate pipeline for a few reasons:
1. We still don't have enough BH cards to track these in our main workflows. This can be a blocker for our development
2. By skipping tests, we won't get results on our superset that we can use for further analysis. It's okay to have them failing until we stabilize it with a metal team
3. There are also plenty of model ops tests and ops sweeps that have a bunch of tests that would require updates with BH failing markere. It's much easier to have a separate failing pipeline and analyze results without skipping and/or making any of them

nsumrakTT: I agree with @nvukobratTT test are there to fail if stuff is not working/there is regression.
It is up to us to say what makes it pass and how we treat green checkmark.

gfengTT: I've pushed a new commit that addresses these concerns:
1. on-push and on-pr only run on n150s and n300s. We may add p150 when we have enough runners.
2. full model passing and xfailing are both running on nightly as a separate stage.
3. test_models_ops and test_sweeps are added to p150 on-nightly. They are allowed to fail.

nsumrakTT (approved):

.github/workflows/on-nightly.yml
```diff
@@ -119,6 +135,22 @@ jobs:
       run_id: ${{ needs.build.outputs.run_id }}
       runs-on: '[{"runs-on": "n150"}]'
 
+  test_full_model_xfailing_p150:
```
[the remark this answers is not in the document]
gfengTT: Good catch. They are not expected to fail or skip though. Just like test_full_model_passing and test_full_model_xfailing, each test should either pass or xfail.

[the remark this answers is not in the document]
gfengTT: I've added both test_full_model_passing_p150 and test_full_model_xfailing_p150 to allowed failures, as initial nightly tests may run into unexpected issues.

forge/test/mlir/test_loss.py
```diff
@@ -49,7 +49,7 @@ def test_l1_loss(forge_property_recorder, prediction_shape, reduction):
 @pytest.mark.parametrize(
     "prediction_shape",
     [
-        (3, 5),
+        pytest.param((3, 5), marks=pytest.mark.fails_on_bh),
```
nvukobratTT, line 52: Let's remove all occurrences of `pytest.mark.fails_on_bh ` as they are redundant now.

pytest.ini
```diff
@@ -16,6 +16,7 @@ markers =
     slow: marks tests as slow               # deprecated - slow tests, should not be run in push pipeline
     run_in_pp: marks tests as run_in_pp     # deprecated - tests that should run in push pipeline
     skip_model_analysis: marks tests as skip_model_analysis
+    fails_on_bh: tests failing on Blackhole (p150)
```
nvukobratTT, line 19: Also, this one is redundant now. Let's remove it from this PR as we have separate pipelines for it. 

The goal is to have everything functional on p150 as well, so I hope we will not need to introduce this flag at all.

.github/workflows/on-nightly.yml
```diff
@@ -91,6 +91,22 @@ jobs:
       runs-on: '[{"runs-on": "n150"}]'
       run_id: ${{ needs.build.outputs.run_id }}
 
+  test_full_model_passing_p150:
```
nvukobratTT, line 94: Also, let's test out `test_models_ops_p150` as well. This will gather additional details on op level (as proposed on previous comments). 

Regarding sweeps, we can skip them for now. Model ops + Models data should be good enough for initial insights.

nvukobratTT (changes requested):
@gfengTT thanks for pushing this change! 

I added a few more comments to clean up this PR, and make it ready to hit `main` :))

.github/workflows/on-nightly.yml
gfengTT: As things stand `test_models_ops` runs on both n150 and p150. Would you like to have two separate jobs, `test_models_ops` and `test_models_ops_p150`?

pytest.ini
gfengTT: Do we need some flags to keep track of which tests are passing and which are failing? We can keep this flag, only allow the tests with this flag to fail, and gradually remove these flags as we improve our compiler.

.github/workflows/on-nightly.yml
nvukobratTT: Let's have them separated. This will make our life easier stabilizing P150, until we can merge them into single one.

Also, model ops are part of the regular nightly, so keeping those running on both will make the issues here. Also, please make sure that we're not running anything else besides newly outlined jobs on P150.

nvukobratTT (commented):
Shared answers on your questions and suggestions @gfengTT. I'll be here for a bit more time, so let's try to push these changes up. Once this hits main, I'll take over on providing initial infrastructure and insights for tracking the p150 state :))

nvukobratTT (approved):

nvukobratTT:
Approving PR. It's good enough to collect initial data. We'll take over restructuring over the next week to make it more suitable for our current infrastructure setup. 

Thanks once again @gfengTT! I'm setting auto-merge to get this one in asap.

PR #2109, opened 2025-05-22
Disable pcc verification for classification.

vcanicTT:
Disable pcc verification for classification.

### DIFF
--- a/forge/test/benchmark/benchmark/models/resnet_hf.py
+++ b/forge/test/benchmark/benchmark/models/resnet_hf.py
@@ -35,6 +35,8 @@ TASK = [
     "classification",
 ]
 
+EVALUATION_SCORE_TARGET = 0.75
+
 # Batch size configurations
 BATCH_SIZE = [
     1,
@@ -120,23 +122,7 @@ def test_resnet_hf(training, batch_size, data_format, input_size, channel_size, 
 
     # Run for the first time to warm up the model, it will be done by verify function.
     # This is required to get accurate performance numbers.
-
-    # If the task is classification, we don't need verification with PCC, accuracy is stronger metric.
-    if task == "classification":
-        # Run the model
-        compiled_model(inputs[0])
-    elif task == "na":
-        verify(
-            [
-                inputs[0],
-            ],
-            framework_model,
-            compiled_model,
-            verify_cfg=VerifyConfig(value_checker=AutomaticValueChecker(pcc=0.95)),
-        )
-    else:
-        raise ValueError(f"Unsupported task: {task}.")
-
+    compiled_model(inputs[0])
     if task == "classification":
         predictions = []
         start = time.time()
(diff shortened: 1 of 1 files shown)


forge/test/benchmark/benchmark/models/resnet_hf.py
```diff
@@ -148,9 +156,14 @@ def test_resnet_hf(training, batch_size, data_format, input_size, channel_size,
     else:
         raise ValueError(f"Unsupported task: {task}.")
 
-    fw_out = framework_model(inputs[-1])[0]
-    co_out = co_out.to("cpu")
-    AutomaticValueChecker(pcc=0.95).check(fw_out=fw_out, co_out=co_out)
+    if task == "classification":
+        pass
```
odjuricicTT, line 160: Should we check that eval socre is above the 0.75 threshold here? And fail the test if this is not met.

forge/test/benchmark/benchmark/models/resnet_hf.py
```diff
@@ -120,14 +120,22 @@ def test_resnet_hf(training, batch_size, data_format, input_size, channel_size,
...
+        verify(
+            [
+                inputs[0],
+            ],
+            framework_model,
+            compiled_model,
+            verify_cfg=VerifyConfig(value_checker=AutomaticValueChecker(pcc=0.95)),
+        )
+    else:
+        raise ValueError(f"Unsupported task: {task}.")
```
odjuricicTT, line 138: Can you remove this part and just call `compiled_model(inputs[0])` to warm up the model in all cases?

You do check pcc or eval score later, so this would be fine.

forge/test/benchmark/benchmark/models/resnet_hf.py
```diff
@@ -148,9 +156,14 @@ def test_resnet_hf(training, batch_size, data_format, input_size, channel_size,
     else:
         raise ValueError(f"Unsupported task: {task}.")
```
odjuricicTT, line 158: Can we print the samples per second here, so we can see the resole even if the following eval score / pcc check fails?

odjuricicTT (changes requested):
A few minor changes

forge/test/benchmark/benchmark/models/resnet_hf.py
vcanicTT: Chnaged.

forge/test/benchmark/benchmark/models/resnet_hf.py
vcanicTT: Changed.

forge/test/benchmark/benchmark/models/resnet_hf.py
vcanicTT: Changed.

odjuricicTT (approved):
Looks good!

PR #2111, opened 2025-05-22
Simplified forge property recorder/handler usage.

acolicTT:
Forge property recorder/handler (FPH) was passed as parameter throughout all code base in order to collect various forge properties.

With this change, FPH is initialized as a context variable at the beginning of every test (with pytest fixture) and can be used with global recording functions that are introduced. Every function checks whether FPH is initialized, and only records properties if it is. This way, we have readable code with recording enabled only for tests.

This is the third part of forge property handler simplification.


chandrasekaranpradeep:
> cc @chandrasekaranpradeep @ashokkumarkannan1 @kamalrajkannan78 @meenakshiramanathan1 @saiarthiraguram
> 
> I also changed forge/forge/python_codegen.py, so we should test generation of a new models_ops tests. Is there any guideline how to do that?

Generating models ops tests will take around a day and then we need to add xfail and skip markers
Since you have replaced the methods with function in the forge/tests/models_ops directory path, we can trigger the Test Workflow with nightly_models_ops and check the report to verify is everything is properly recorded or not.

chandrasekaranpradeep:
@acolicTT  You need to import the forge property recording functions in the forge/test/models_ops path
`from forge.forge_property_utils import record_forge_op_name, record_op_model_names, record_forge_op_args, record_single_op_operands_info`

forge/forge/forge_property_utils.py
```diff
@@ -793,55 +606,282 @@ def store_property(self, record_property):
...
+        return
+
+    pcc, atol = determine_consistency_limits(framework_outputs=framework_outputs, compiled_outputs=compiled_outputs)
+    if pcc is not None:
+        fph.add("tags.pcc", pcc)
+    if atol is not None:
+        fph.add("tags.atol", atol)
+
+
+def record_error(request: FixtureRequest):
```
chandrasekaranpradeep, line 781: I think we don't need function right? because we are recording the refined_error_message and failure_category through record_error method in ForgePropertyHandler class in the record_forge_property pytest fixture present in the conftest.py

forge/forge/verify/verify.py
```diff
@@ -30,7 +30,13 @@
 from forge.compiled_graph_state import CompiledModel
 from forge.verify.compare import compare_tensor_to_golden, determine_consistency_limits
 from forge.verify.utils import convert_to_supported_pytorch_dtype
-from forge.forge_property_utils import ForgePropertyHandler, ExecutionStage
+from forge.forge_property_utils import (
+    ForgePropertyHandler,
```
chandrasekaranpradeep, line 34: Importing ForgePropertyHandler is not need right?

chandrasekaranpradeep (changes requested):

forge/forge/forge_property_utils.py
acolicTT: Good catch!

acolicTT:
> > cc @chandrasekaranpradeep @ashokkumarkannan1 @kamalrajkannan78 @meenakshiramanathan1 @saiarthiraguram
> > I also changed forge/forge/python_codegen.py, so we should test generation of a new models_ops tests. Is there any guideline how to do that?
> 
> Generating models ops tests will take around a day and then we need to add xfail and skip markers Since you have replaced the methods with function in the forge/tests/models_ops directory path, we can trigger the Test Workflow with nightly_models_ops and check the report to verify is everything is properly recorded or not.

Sounds good :)

chandrasekaranpradeep (approved):

acolicTT:
> it might be good to run nightly passing models, just to confirm everything works - since we are modifying all of the tests... if you haven't already done so...

I did last night, but it got canceled: https://github.com/tenstorrent/tt-forge-fe/actions/runs/15197504630
Anyway, ~65% passed before cancelation, so I think that we should be fine.

PR #2114, opened 2025-05-22
Always build wheel on nightly

jmcgrathTT:
Always rebuild the wheel in the nightly job so it can be picked up by `tt-forge` nightly release job.

https://github.com/tenstorrent/tt-forge/actions/runs/15179302234/job/42685332775#step:4:129

Test
https://github.com/tenstorrent/tt-forge-fe/actions/runs/15196733431/job/42742551052#step:2:8


nsumrakTT (changes requested):
The problem you've showed in the ticket is not related to this workflow but the processing one which encountered situation that nightly is in progress and tried to pickup release wheel from non-scheduled (manually triggered) run.
I think you should close this PR.

PR #2123, opened 2025-05-23
Enable conv sharding in resnet

odjuricicTT:
Resnet now at 240fps :)

🔥 @rpavlovicTT


nvukobratTT (approved):
As Windsurf would say, "🤙"!   😂

vcanicTT (approved):

PR #2130, opened 2025-05-25
Enable bfloat16 conversion for all remaining Conv models except those with regressions

kamalrajkannan78:


saiarthiraguram (approved):

vkovinicTT (commented):
- In the `monodle` log there is some `ConnectionResetError`, can you just confirm that tests are passing since you removed `skip` annotation from them? 

- Do `swin` model tests fail with `segmentation fault` since there is no log attached?

vkovinicTT:
> > * In the `monodle` log there is some `ConnectionResetError`, can you just confirm that tests are passing since you removed `skip` annotation from them?
> 
> existing skip reason "Floating point exception(core dumped)" didn't align with latest ConnectionResetError. so I removed that .should we add xfail in such cases?
> 
Sure add xfail reason or even better if you have some spare time (no more than 30mins) can you investigate why do we get that error?

PR #2137, opened 2025-05-26
Fix forge property recording and import record_model_properties function in models tests

chandrasekaranpradeep:
1. This PR addresses an issue where all instances of ForgePropertyHandler share a single default ForgePropertyStore due to the use of a mutable default argument in the constructor. This behavior leads to unexpected persistence of property values across different handler instances(i.e each test execution). The default argument `store: ForgePropertyStore = ForgePropertyStore()` is evaluated once when the class is defined, not each time __init__ is called. Every time we do ForgePropertyHandler() without passing a store, it reuses that one ForgePropertyStore instance created at definition time. Changes made to self.store in one handler (e.g. setting model properties) stay on that same object and thus appear in subsequent handlers which leads to duplicate model names.
2. Imported missing record_model_properties function in the models tests

### DIFF
--- a/forge/forge/forge_property_utils.py
+++ b/forge/forge/forge_property_utils.py
@@ -372,8 +372,8 @@ class ForgePropertyHandler:
         should be recorded.
     """
 
-    def __init__(self, store: ForgePropertyStore = ForgePropertyStore()):
-        self.store = store
+    def __init__(self, store: Optional[ForgePropertyStore] = None):
+        self.store = store if store is not None else ForgePropertyStore()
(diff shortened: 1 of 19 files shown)


nvukobratTT (approved):

forge/forge/forge_property_utils.py
```diff
@@ -372,8 +372,8 @@ class ForgePropertyHandler:
         should be recorded.
     """
 
-    def __init__(self, store: ForgePropertyStore = ForgePropertyStore()):
```
acolicTT, line 375: This is disturbing language design. Approved.

acolicTT (approved):

PR #2139, opened 2025-05-26
Disable memory layout analysis on resnet performance benchmark model.

vcanicTT:
Disable memory layout analysis on resnet performance benchmark model. When this parameter is enabled we get an error on resnet, until we fix the problem this parameter is disabled.


vcanicTT:
It was machine problem, after reboot on CI resnet works again. If we face this problem I will re-open the PR.

vcanicTT:
This is result after disabling the feature:
Screenshot 2025-05-28 at 17 45 53

PR #2150, opened 2025-05-27
Disable device perf measuring for all performance benchmark models.

vcanicTT:
We faced troubles with TTRT on all performance benchmark models. So, we disable device perf measuring for all performance benchmark models, until we solve the problem.


vcanicTT:
@odjuricicTT this flag is only on TTRT, and device perf, if e2e fails, the test we will fail, too.

odjuricicTT (approved):

Issue #2033, opened 2025-05-15
RuntimeError: Couldn't load custom C++ ops.


saiarthiraguram:
Facing RuntimeError: Couldn't load custom C++ ops. in the below models, 

- forge/test/models/pytorch/vision/rmbg/test_rmbg.py::test_rmbg[briaai/RMBG-2.0]
- forge/test/models/pytorch/vision/ssdlite320_mobilenetv3/test_ssdlite320_mobilenetv3.py::
test_ssdlite320_mobilenetv3[ssdlite320_mobilenet_v3_large]
- forge/test/models/pytorch/vision/ssd300_vgg16/test_ssd300_vgg16.py::test_ssd300_vgg16[ssd300_vgg16]
- forge/test/models/pytorch/vision/ssd300_vgg16/test_ssd300_vgg16.py::test_ssd300_vgg16[ssd300_vgg16]

skhatriTT:
@ashokkumarkannan1 does the linked commit fixed this issue?

PR #2159, opened 2025-05-29
Fix C++ ops runtime error in ssd mobilenetv3

ashokkumarkannan1:
Ticket
(https://github.com/tenstorrent/tt-forge-fe/issues/2033)

Problem description
The  RuntimeError: Couldn't load custom C++ ops. encountered during the execution of the framework model is due to an unsupported C++ operation. Specifically, the issue arises from the presence of the nms (Non-Maximum Suppression) operation in the model's forward pass, which is implemented in C++ and is currently not supported in this context.

What's changed
This PR includes the changes to make the postprocessing outside the model which resolves the issue.

Log:
[ssd_mobilenet.log](https://github.com/user-attachments/files/20500357/ssd_mobilenet.log)

### DIFF
--- a/forge/test/models/pytorch/vision/ssdlite320_mobilenetv3/model_utils/model_utils.py
+++ b/forge/test/models/pytorch/vision/ssdlite320_mobilenetv3/model_utils/model_utils.py
@@ -1,0 +1,139 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+
+
+import torch
+import torch.nn.functional as F
+from torch import Tensor
+from torchvision.models.detection import _utils as det_utils
+from torchvision.models.detection.anchor_utils import DefaultBoxGenerator
+from torchvision.models.detection.transform import GeneralizedRCNNTransform
+from torchvision.ops import boxes as box_ops
+
+
+def postprocess_detections(
+    head_outputs: dict[str, Tensor], image_anchors: list[Tensor], image_shapes: list[tuple[int, int]]
+) -> list[dict[str, Tensor]]:
+    bbox_regression = head_outputs["bbox_regression"]
+    pred_scores = F.softmax(head_outputs["cls_logits"], dim=-1)
+
+    num_classes = pred_scores.size(-1)
+    device = pred_scores.device
+
+    detections: list[dict[str, Tensor]] = []
+
+    for boxes, scores, anchors, image_shape in zip(bbox_regression, pred_scores, image_anchors, image_shapes):
+        box_coder = det_utils.BoxCoder(weights=(10.0, 10.0, 5.0, 5.0))
+        boxes = box_coder.decode_single(boxes, anchors)
+        boxes = box_ops.clip_boxes_to_image(boxes, image_shape)
+
+        image_boxes = []
+        image_scores = []
+        image_labels = []
+        for label in range(1, num_classes):
+            score = scores[:, label]
+
+            keep_idxs = score > 0.01
+            score = score[keep_idxs]
+            box = boxes[keep_idxs]
+
+            # keep only topk scoring predictions
+            num_topk = det_utils._topk_min(score, 400, 0)
+            score, idxs = score.topk(num_topk)
+            box = box[idxs]
+
+            image_boxes.append(box)
+            image_scores.append(score)
+            image_labels.append(torch.full_like(score, fill_value=label, dtype=torch.int64, device=device))
+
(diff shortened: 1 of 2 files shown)


ashokkumarkannan1:
> @ashokkumarkannan1 @saiarthiraguram is this still an issue with recent upgrade to PyTorch 2.7?


Hi @nvukobratTT , In latest main the above model is failing at the compile stage generate_initial_graph will fix it
Issue Log:
[ssd320_mobilenet.log](https://github.com/user-attachments/files/20689010/ssd320_mobilenet.log)

Since the NMS output solely depends on the number of bounding boxes, it is recommended to keep the post-processing outside the model.

ashokkumarkannan1:
> @ashokkumarkannan1 are we still getting C++ ops runtime errors after dependencies update?

Hi @nvukobratTT we are not getting C++ Ops issue in the current main we will be getting the ` AttributeError: 'dict' object has no attribute 'detach'` which is caused due to the postprocessing kept inside the model.

nvukobratTT:
> > @ashokkumarkannan1 are we still getting C++ ops runtime errors after dependencies update?
> 
> Hi @nvukobratTT we are not getting C++ Ops issue in the current main instead we will be getting the ` AttributeError: 'dict' object has no attribute 'detach'` which is caused due to the postprocessing kept inside the model.

Do you have more details on where specifically? We saw similar details on HF where many models return dict as default. However, when that one is disabled they return tesnsors as we expect :))

ashokkumarkannan1:
> that one is disabled they



> > > @ashokkumarkannan1 are we still getting C++ ops runtime errors after dependencies update?
> > 
> > 
> > Hi @nvukobratTT we are not getting C++ Ops issue in the current main instead we will be getting the ` AttributeError: 'dict' object has no attribute 'detach'` which is caused due to the postprocessing kept inside the model.
> 
> Do you have more details on where specifically? We saw similar details on HF where many models return dict as default. However, when that one is disabled they return tesnsors as we expect :))

Hi @nvukobratTT  this is the line in the model which returns the dict which is from postprocesing https://github.com/pytorch/vision/blob/6bbe0100fb7002f5daa66e9cc970377e367e5a33/torchvision/models/detection/ssd.py#L414

nvukobratTT:
> > that one is disabled they
> 
> > > > @ashokkumarkannan1 are we still getting C++ ops runtime errors after dependencies update?
> > > 
> > > 
> > > Hi @nvukobratTT we are not getting C++ Ops issue in the current main instead we will be getting the ` AttributeError: 'dict' object has no attribute 'detach'` which is caused due to the postprocessing kept inside the model.
> > 
> > 
> > Do you have more details on where specifically? We saw similar details on HF where many models return dict as default. However, when that one is disabled they return tesnsors as we expect :))
> 
> Hi @nvukobratTT this is the line in the model which returns the dict which is from postprocesing https://github.com/pytorch/vision/blob/6bbe0100fb7002f5daa66e9cc970377e367e5a33/torchvision/models/detection/ssd.py#L414 Yes we can do that in the huggingface models but we are here dealing with is pytorch vision models

I see, can we just import these functions, and keep the wrapper in a similar structure without copy-pasted code?

nvukobratTT:
> > > > that one is disabled they
> > > 
> > > 
> > > > > > @ashokkumarkannan1 are we still getting C++ ops runtime errors after dependencies update?
> > > > > 
> > > > > 
> > > > > Hi @nvukobratTT we are not getting C++ Ops issue in the current main instead we will be getting the ` AttributeError: 'dict' object has no attribute 'detach'` which is caused due to the postprocessing kept inside the model.
> > > > 
> > > > 
> > > > Do you have more details on where specifically? We saw similar details on HF where many models return dict as default. However, when that one is disabled they return tesnsors as we expect :))
> > > 
> > > 
> > > Hi @nvukobratTT this is the line in the model which returns the dict which is from postprocesing https://github.com/pytorch/vision/blob/6bbe0100fb7002f5daa66e9cc970377e367e5a33/torchvision/models/detection/ssd.py#L414 Yes we can do that in the huggingface models but we are here dealing with is pytorch vision models
> > 
> > 
> > I see, can we just import these functions, and keep the wrapper in a similar structure without copy-pasted code?
> 
> Hi @nvukobratTT, The postprocess_detection function is part of the SSD class and relies on parameters from the class via self, so it can't be imported or used directly. Additionally, the Postprocessor class in model_utils.py is customized to fit the model wrapper and other specific functionalities.

I see, I'm not in favor of structuring it this way. Let's put this one on hold until we transfer most of the models to the tt-forge-models, and then figure out per case basis how to structure them. Potentially, we'll extract model core architecture to tt-forge-models, and run it that way. However, let's discuss this one later.

PR #2166, opened 2025-05-30
Refactor model analysis ops tests generation workflow

chandrasekaranpradeep:
Currently, the model analysis workflow will extract the unique ops configuration across all the models present in the `forge/test/models` path and generate models ops tests by utilising the `scripts/model_analysis/generate_models_ops_test.py` script. The `scripts/model_analysis/generate_models_ops_test.py `script will first collect all the models tests which doesn't contains skip_models_analysis marker and run the collected tests one by one through subprocess run method to extract and export the unique ops configuration. After running all the models tests, will use the exported unique ops configuration to generate models ops tests. The model analysis workflow will take around a day to do this process and it also need large disk space. Since it need large disk space, it can't run on shared runners or organisation runners which produces no space left on device issues. 

To resolve this, will split the model analysis extraction and generation process into different jobs.Mainly we are going to use shared runners to run the models tests for extracting and exporting the uniques ops configuration and save it as artefacts. Finally one jobs will collect all the unique ops configuration by downloading the artefacts and combine the unique ops config and generate models ops tests.

New Model analysis Workflow:
1) For running the models tests to extract and export unique ops config across runners, created `extract-unique-ops-configuration` job which will utilise the test-sub workflow to run those model analysis tests and save the extracted unique ops configuration as artefacts
2) Created `generate-models-ops-tests` job which will download all the unique ops config artefacts from the workflow across attempts and then combine and generate unique ops tests and then create PR for generate models ops tests.

Created ExitTest exception and raised the exception after extracting and exporting model unique ops config which help to exit the tests instead of exiting the pytest entirely and  pytest_runtest_makereport will mark the tests as passed instead of making it as failure which helps to distinguish the issues before extracting the unique ops configuration.

In current ci runners some of the models test will kill pytest process due to out of memory issues so we have skipped the models tests in CI. The existing skip markers will also skip failures after forge module generation so I have added new marker for out of memory models which helps to run those tests locally and generate ops tests.

### DIFF
--- a/.github/workflows/model-analysis.yml
+++ b/.github/workflows/model-analysis.yml
@@ -2,12 +2,35 @@ name: Model Analysis
 
 on:
   workflow_dispatch:
-  push:
-    branches: [ "pchandrasekaran/models_ops_workflow" ]
+    inputs:
+      test_group_cnt:
+        description: 'Test group count'
+        required: false
+        default: "10"
+        type: choice
+        options:
+          - "2"
+          - "4"
+          - "6"
+          - "8"
+          - "10"
+          - "12"
+      runs-on:
+        description: 'Runs on'
+        required: false
+        default: runner
+        type: choice
+        options:
+          - runner
+          - n150
+          - n300
+          - p150
 
 permissions:
   packages: write
   checks: write
+
+run-name: "Model Analysis Ops Test Generation (${{inputs.runs-on}}-${{inputs.test_group_cnt}})"
 
 jobs:
 
@@ -15,8 +38,26 @@ jobs:
     uses: ./.github/workflows/build-image.yml
     secrets: inherit
 
-  build:
+  set-inputs:
+    runs-on: ubuntu-latest
     needs: docker-build
+    if: always()
+    outputs:
+      test_group_cnt: ${{ steps.set-inputs.outputs.test_group_cnt }}
+      test_group_ids: ${{ steps.set-inputs.outputs.test_group_ids }}
+      runs-on: ${{ steps.set-inputs.outputs.runs-on }}
+    steps:
+      - name: Inputs Management
+        id: set-inputs
+        run: |
+          echo "test_group_cnt=${{ inputs.test_group_cnt }}" >> $GITHUB_OUTPUT
+          echo "test_group_ids=[$(seq -s ',' 1 ${{ inputs.test_group_cnt }})]" >> $GITHUB_OUTPUT
+          echo "runs-on=[{\"runs-on\": \"${{ inputs.runs-on }}\"}]" >> $GITHUB_OUTPUT
+
+  build:
+    needs:
+      - docker-build
+      - set-inputs
     uses: ./.github/workflows/build.yml
     secrets: inherit
     with:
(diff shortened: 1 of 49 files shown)


.github/workflows/model-analysis.yml
```diff
@@ -2,6 +2,28 @@ name: Model Analysis
...
+        required: false
+        default: "8"
+        type: choice
+        options:
+          - "2"
+          - "4"
+          - "6"
+          - "8"
+          - "10"
+          - "12"
```
nsumrakTT, line 17: We have 10 n150, and much less of n300 and p150. I think you should remove 12 as an option if you plan using forge-fe runners

.github/workflows/model-analysis.yml
```diff
@@ -13,10 +35,58 @@ jobs:
...
+    needs:
+      - docker-build
+      - set-inputs
+    uses: ./.github/workflows/build.yml
+    secrets: inherit
+    with:
+      docker-image: ${{ needs.docker-build.outputs.docker-image }}
+      build: 'Release'
+
+  extract-unique-ops-configuration:
```
nsumrakTT, line 64: Do you know how long this will take with 10 runners?

nsumrakTT (changes requested):
I think it would be much better if you create test-model-analysis-sub.yml instead of changing test-sub.yml as difference is substantial and maintenance would be difficult with so many ifs.

.github/workflows/model-analysis.yml
vmilosevic: We can manage how many machines we want to use for analysis by adding a new `analysis` label to some of the runners, and ask for in in a workflow `runs-on: analysis`. 
For example in that way work will be split in 12 batches but will execute only on 2 runners.

.github/workflows/model-analysis.yml
chandrasekaranpradeep: For 10 runners, it takes around 2 hours 5 mins but due to out of memory issues from phi4 some tests are not ran in 3 jobs, so it make takes approximately 2 hours 30 mins - https://github.com/tenstorrent/tt-forge-fe/actions/runs/15374631757

chandrasekaranpradeep:
> I think it would be much better if you create test-model-analysis-sub.yml instead of changing test-sub.yml as difference is substantial and maintenance would be difficult with so many ifs.

Created test-model-analysis-sub.yml

.github/workflows/model-analysis.yml
chandrasekaranpradeep: The model analysis ops tests generation is independent of device since it will generate models ops tests after TVM compilation. So we will run the tests on any runners(n150/p150/n300) so I have changed the default option to runner for runs-on input so it will pick runner which has in-service and runner tag. Now we are able to use more runners.

chandrasekaranpradeep: > The model analysis ops tests generation is independent of device since it will extract and export unique ops config after TVM compilation and stop the test execution after that. So we can run the tests on any runners(n150/p150/n300) so I have changed the default option to runner for runs-on input so it will pick runner which has in-service and runner tag. Now we are able to use more runners.

.github/workflows/model-analysis.yml
```diff
@@ -2,6 +2,29 @@ name: Model Analysis
...
+      runs-on:
+        description: 'Runs on'
+        required: false
+        default: runner
+        type: choice
+        options:
+          - runner
+          - n150
+          - n300
+          - p150
```
nsumrakTT, line 28: You can add:
`run-name: "Model Analysis (${{inputs.runs-on}}-${{inputs.test_group_cnt}})`
That way the name on the Actions will reflect run options.

chandrasekaranpradeep: Sure

nvukobratTT (approved):

PR #2177, opened 2025-06-02
Save inputs for test workflow

vbrkicTT:
### Problem description
Once a test workflow is executed it's hard to determine which input parameters were used

### What's changed
Store input parameters in test step `save-inputs` summary.
Store input parameters in inputs.json artifact.

### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,6 +58,29 @@ permissions:
 
 jobs:
 
+  save-inputs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Show input parameters
+        run: |
+          echo "## Input Parameters" >> $GITHUB_STEP_SUMMARY
+          echo "- Branch: ${{ github.ref_name }}" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -58,6 +58,29 @@ permissions:
```
nsumrakTT, line 58: Add this and display name of all Test runs will have parameters clearly visible:
`run-name: "Test (Rebuild: ${{ inputs.rebuild }} - Preset: ${{ inputs.preset }} - Mark: ${{ inputs.test_mark }} - ${{ inputs.runs-on }} - ${{ inputs.test_group_cnt }} - Ops: ${{ inputs.operators }} - Filters: ${{ inputs.filters }})"`
Screenshot 2025-06-02 at 13 42 15

nsumrakTT (approved):

.github/workflows/test.yml
nsumrakTT: With this change we also get inputs into superset and it is searchable
Screenshot 2025-06-02 at 14 07 12

Issue #1959, opened 2025-05-08
Issues in TVM decomposition of upsampleLinear1d


saiarthiraguram:
During decomposition of upsample_linear 1d in forge test, we face the below issues.
```
TVMError: In function relay.op._make.squeeze(0: RelayExpr, 1: Array<IntImm>) -> RelayExpr:
error while converting argument 0: InternalError: Check failed: (!checked_type.defined()) is false:
Expected RelayExpr, but got Array

```
Created sanity decomposition for upsamplelinear1d:

> Using Relay ops: shape_of, strided_slice, cast, arange, take, floor, ceil, expand_dims, etc.
> 
> Sanity passed for various input shapes and target sizes.

Sanity Test:

```
import pytest
import torch
import torch.nn.functional as F
import tvm
from tvm import relay
import numpy as np

def run_custom_upsample_linear1d(input_np, target_size):
    """Run manual decomposition for 1D linear upsampling."""
    x = relay.var("x", shape=input_np.shape, dtype="float32")
    size = relay.const(target_size, dtype="int32")

    input_shape = relay.shape_of(x)
    input_w = relay.strided_slice(input_shape, begin=[2], end=[3])  # (1,)
    input_w_f = relay.cast(input_w, "float32")  # (1,)
    output_w_f = relay.cast(size, "float32")

    scale = relay.divide(input_w_f, output_w_f)

    zero_const = relay.const(0, "int32")
    out_indices = relay.arange(zero_const, size, dtype="int32")
    indices_f = relay.cast(out_indices, "float32")

    half_const = relay.const(0.5, "float32")
    mapped_pos = relay.subtract(
        relay.multiply(relay.add(indices_f, half_const), scale),
        half_const
    )

    floor_idx_f = relay.floor(mapped_pos)
    ceil_idx_f = relay.ceil(mapped_pos)

    floor_idx = relay.cast(floor_idx_f, "int32")
    ceil_idx = relay.cast(ceil_idx_f, "int32")

    one_const = relay.const(1, "int32")
    input_w_minus_1 = relay.subtract(input_w, one_const)

    floor_idx = relay.maximum(floor_idx, zero_const)
    floor_idx = relay.minimum(floor_idx, input_w_minus_1)
    ceil_idx = relay.maximum(ceil_idx, zero_const)
    ceil_idx = relay.minimum(ceil_idx, input_w_minus_1)

    floor_idx = relay.cast(floor_idx, "int64")
[...]

PR #2158, opened 2025-05-29
Add support for image.resize1d op

meenakshiramanathan1:
This PR adds support for image.resize1d op till mlir stage.

Logs:
[image_resize1d.log](https://github.com/user-attachments/files/20498010/image_resize1d.log)

PR #2212, opened 2025-06-05
Unblock SAM Model by adding support for upsample_linear1d 

kamalrajkannan78:
### Summary 

This PR tracks [this changes](https://github.com/tenstorrent/tt-tvm/pull/88) and verifies it with sanity & fixes https://github.com/tenstorrent/tt-forge-fe/issues/1959

### Logs

[jun5_sanity_bf.log](https://github.com/user-attachments/files/20610798/jun5_sanity_bf.log)
[jun5_sanity_af.log](https://github.com/user-attachments/files/20610797/jun5_sanity_af.log)

### DIFF
--- a/forge/test/mlir/operators/nn/test_nn.py
+++ b/forge/test/mlir/operators/nn/test_nn.py
@@ -840,3 +840,32 @@ def test_resize1d(data, output_size, align_corners, forge_tmp_path):
     compiled_model = forge.compile(framework_model, sample_inputs=[x])
 
     verify([x], framework_model, compiled_model)
+
+
+@pytest.mark.parametrize(
+    "shape, output_size",
+    [
+        ((1, 64, 27), 54),
+        ((1, 3, 12), 48),
+        ((1, 1, 8), 24),
+        ((2, 36, 5), 50),
+        ((4, 4, 7), 56),
+    ],
+)
+@pytest.mark.xfail(
+    reason="RuntimeError: Found Unsupported operations while lowering from TTForge to TTIR in forward graph - resize1d"
+)
+def test_upsample_linear1d(shape, output_size):
+    class upsample_linear1d(nn.Module):
+        def __init__(self):
+            super().__init__()
+
+        def forward(self, x):
+            return nn.functional.interpolate(x, size=output_size, mode="linear")
+
+    inputs = [torch.rand(shape)]
+
+    framework_model = upsample_linear1d()
+    compiled_model = forge.compile(framework_model, sample_inputs=inputs)
+
+    verify(inputs, framework_model, compiled_model)


saiarthiraguram (approved):

dgolubovicTT:
In which model we need this upsample linear1D?

Also, note that upsample2d with bilinear mode doesn't work because we don't shard it: https://github.com/tenstorrent/tt-mlir/issues/3396

dgolubovicTT:
If this upsample1D is needed for some Red team model, we should bump up the priority of this: https://github.com/tenstorrent/tt-mlir/issues/3396

kamalrajkannan78:
> In mlir we have upsample2d that lowers to [ttnn::upsample](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/pool/upsample/upsample.cpp). Upsample 1D is special case of 2D. We should do decomposition of upsample1D and use 2D.
> 
> However, before all of that you should test if upsample works on ttnn for shape (1, len).
> 
> Also, upsample 2D has only mode "bilinear" not linear - upsample2d with bilinear mode is the 2D generalization of 1D linear interpolation.

- In [SAM model](https://github.com/tenstorrent/tt-forge-fe/blob/main/forge/test/models/pytorch/vision/sam/test_sam.py) (red team Model) , interpolation is happening [here](https://github.com/huggingface/transformers/blob/5d7739f15a6e50de416977fe2cc9cb516d67edda/src/transformers/models/sam/modeling_sam.py#L765C27-L765C41). there are only 2 cases present during model's forward pass

| Case | Input Size      | Output Size | Mode   |
|------|-----------------|-------------|--------|
| 1    | [1, 64, 27]   | 27        | linear |
| 2    | [1, 64, 127]  | 127       | linear |

- for more details refer this log [jun10_sam_info.log](https://github.com/user-attachments/files/20672450/jun10_sam_info.log).you can see upsampling is not happening here !!
- As functionality of upsample_linear1d and image.resize1d is same & [forge frontend support for resize1d](https://github.com/tenstorrent/tt-forge-fe/pull/2158) is added recently , i just mapped it to resize1d and verified it with some valid upsampling cases like `input size - ([1, 64, 27]) out_size - (54) mode - linear ` instead of the one present in model because we may need it in future. but as it now involves [some additional](https://github.com/tenstorrent/tt-forge-fe/pull/2212#pullrequestreview-2910012189) efforts, can we go with below decomposition ?

to unblock this model we just need  to

```
def upsample_linear1d(self, inputs, input_types):
    data = inputs[0]
    out_size = inputs[1]

    data_shape = infer_shape(data)

    # Check if the last dimension of data matches out_size
    if data_shape[-1] == out_size:
        return data  # Return input directly if sizes match
```

instead of https://github.com/tenstorrent/tt-tvm/pull/88/files#:~:text=def%20upsample_linear1d(,%2C%20out_size)

what do you think @dgolubovicTT ?

kamalrajkannan78:
closing this PR as support was added here - https://github.com/tenstorrent/tt-forge-fe/pull/2840

PR #2231, opened 2025-06-09
Fix torch update regressions in dpr context & question encoder

kamalrajkannan78:
the first 2 forward calls of DPRContextEncoder doesn't have position_ids argument. so position_ids argument was added as argument via monkey patch

### Summary

This PR addresses regressions introduced by recent PyTorch updates in the DPR context and question encoders.

### Fix

To ensure consistent static tracing and prevent dynamic control flow issues during TorchScript or Forge compilation, this PR manually sets position_ids in test inputs. This avoids conditional logic on None values within the model’s forward methods, which can cause graph breaks.

With PyTorch 2.7.0, dynamic control flow tracing via TorchDynamo became the default behavior in torch.compile. As a result, optional inputs like position_ids—when not explicitly passed—can trigger dynamic shape errors during tracing.
Monkey Patch

The original forward methods of DPRContextEncoder,DPRQuestionEncoder and DPREncoder didn't position_ids as an argument. This PR adds  position_ids as an argument via monkey-patching 

### Logs

[jun9_dpr_before_fix_all.log](https://github.com/user-attachments/files/20656415/jun9_dpr_before_fix_all.log)
[jun9_dpr_context_enc_after_fix.log](https://github.com/user-attachments/files/20656416/jun9_dpr_context_enc_af.log)
[jun9_dpr_question_enc_after_fix.log](https://github.com/user-attachments/files/20656417/jun9_dpr_question_af.log)


kamalrajkannan78:
> I spoke with @nvukobratTT and we agreed to skip monkey-patching the models. We want them to work directly from Hugging Face as is, except for preprocessing and postprocessing which we can modify.

@vkovinicTT could you please suggest a way to patch this error ?

kamalrajkannan78:
> Hmmmm I don't have enough context to solve this right now, so I think it's okay to leave them with the `xfail` mark for now until we decide how to handle the tracing problems, since they're also occurring in other models. Do you agree @nvukobratTT?

I explained the context here https://github.com/tenstorrent/tt-forge-fe/pull/2231#issue-3130502110. do you need additional context @vkovinicTT ?

PR #2276, opened 2025-06-13
[wheel] modify project name to avoid conflicts

pilkicTT:
To avoid conflicts with existing project on official `pypi` index, rename the project from `forge` to `tt_forge_fe`.

Package name stays the same, so there are no functional changes; i.e. `import forge` still works.


nvukobratTT (approved):

nsumrakTT (approved):
Thank you!

PR #2343, opened 2025-06-18
Removed first part of legacy buda ops lowering context.

acolicTT:
This change removes: LoweringContext, ForgeBlocks, ForgeName, ForgeQueueLayout, set tile dimensions pass, ForgeOpAttrQueueDramLocs, ForgeKernelBroadcastInputs, DramLoc.


forge/forge/op/eval/forge/abs.py
```diff
@@ -44,7 +44,7 @@ def backward(self, ac, operand, inputs, output, grad):
         stretched = ac.op("multiply", (subtract, ac.constant(2.0)))
         return ac.op("multiply", (stretched, grad))
 
-    def lower(self, lc, tensors, outputs):
```
nvukobratTT, line 47: We could entirely remove all `def lower` functions as well. However, please let me know if you think it would be better to do this now or during the op transition to the new structure.

nvukobratTT (approved):
Minor comment, approving the change.

forge/forge/op/eval/forge/abs.py
acolicTT: It is better now. Removed :)

forge/csrc/lower_to_forge/common.cpp
```diff
@@ -8,17 +8,6 @@
...
-    return str.find(substr) != std::string::npos;
-}
-
-std::ostream &operator<<(std::ostream &os, DramLoc const &dram_loc)
-{
-    os << "[" << dram_loc.channel << ", 0x" << std::hex << dram_loc.address << std::dec << "]";
-    return os;
-}
-
 inline void to_netlist(std::ostream &os, int i) { os << i; }
```
dgolubovicTT, line 11: Should we get rid of all occurrences of to_netlist in future?

dgolubovicTT (approved):
Thanks for this one!

forge/csrc/lower_to_forge/common.cpp
acolicTT: Yes, yes. I will pick it up in one of the next PRs.

PR #2360, opened 2025-06-20
Run on PR pipeline on p150.

nvukobratTT:
Note: Not ready to be enabled yet due to perf concerns on BH.


nvukobratTT:
Note: We shouldn't push this change until we figure out the performance issues across tests when running on BH. 

cc @nsumrakTT @vkovinicTT

Issue #2326, opened 2025-06-17
RuntimeError Generated MLIR module failed verification


nvukobratTT:
## Summary
Affected models:
```py
forge/test/models/onnx/text/bert/test_bert.py::test_bert_sentence_embedding_generation_onnx[emrecan/bert-base-turkish-cased-mean-nli-stsb-tr]
forge/test/models/onnx/vision/swin/test_swin.py::test_swin_v2_tiny_masked_onnx[microsoft/swinv2-tiny-patch4-window8-256]
forge/test/models/pytorch/vision/swin/test_swin.py::test_swin_torchvision[swin_v2_b]
forge/test/models/pytorch/vision/swin/test_swin.py::test_swin_torchvision[swin_v2_s]
forge/test/models/pytorch/vision/swin/test_swin.py::test_swin_torchvision[swin_v2_t]
forge/test/models/pytorch/vision/yolo/test_yolo_world.py::test_yolo_world_inference
```

PR #2367, opened 2025-06-20
remove convert_broadcast_ops_to_tms pass

mstojkovicTT:
### Ticket
fixes: #2326

### Problem description
We used to have `input_tms` as a concept where, instead of having a broadcast op, we had an `input_tm` that indicated one or more inputs should change dimensions. When lowering to MLIR, that field was simply ignored and left for the rest of the compiler to handle (implicit broadcasting, usually possible with elementwise binary ops). That worked fine until implicit broadcasting wasn't possible later on (for example, in unary ops).

### What's changed
With the removal of the `convert_broadcast_ops_to_tms` pass, we no longer create those `input_tms` and instead leave the broadcast op as is. Also added mapping for the broadcast op.

### DIFF
--- a/forge/csrc/forge_passes.cpp
+++ b/forge/csrc/forge_passes.cpp
@@ -206,10 +206,6 @@ graphlib::Graph *run_pre_lowering_passes(graphlib::Graph *graph, const std::opti
         fuse_gelu(graph);
     }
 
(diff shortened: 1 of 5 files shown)


nvukobratTT (commented):
Can you describe the impact and reason for pushing this change as part of PR?

PR #2373, opened 2025-06-23
Refactor error messages:Remove duplicate and inconsistent messages

meenakshiramanathan1:
This PR cleans up and consolidates error messages by grouping them based on current limitations. Duplicate messages  have been generalized. The approach focuses on showing only the current blocking limitation, allowing future updates as issues are resolved.

### DIFF
--- a/forge/test/models/pytorch/text/qwen/test_qwen_v2.py
+++ b/forge/test/models/pytorch/text/qwen/test_qwen_v2.py
@@ -177,7 +177,4 @@ def test_qwen2_conditional_generation(variant):
         priority=ModelPriority.P1,
     )
 
-    if variant in ["Qwen/Qwen2.5-VL-72B-Instruct", "Qwen/QVQ-72B-Preview"]:
-        raise RuntimeError("Requires transformers>=4.49.0 and multi-chip support")
-    else:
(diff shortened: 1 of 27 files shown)


saiarthiraguram (approved):

forge/test/models/pytorch/text/qwen/test_qwen_v2.py
```diff
@@ -177,7 +177,4 @@ def test_qwen2_conditional_generation(variant):
         priority=ModelPriority.P1,
     )
 
-    if variant in ["Qwen/Qwen2.5-VL-72B-Instruct", "Qwen/QVQ-72B-Preview"]:
-        raise RuntimeError("Requires transformers>=4.49.0 and multi-chip support")
-    else:
-        raise RuntimeError("Requires transformers>=4.49.0")
+    raise RuntimeError("Requires upgrade of `transformers` version.")
```
vkovinicTT, line 180: Can we also put `xfail` mark here (and in other files as well)?

vkovinicTT (commented):
I've synced with Vukobrat offline and he explained it to me why we are not putting `xfail` reason above the function definition, so no need to address these comments. 

Just one question: can we set the rest of `RuntimeError`s as `xfail` as well?

Sorry for the confusion :))

PR #2377, opened 2025-06-23
[Tech Debt] Remove optimization flag from test models

vkovinicTT:
### What's changed
-  Disabled optimization pass from the models that are working without it. Our goal is to deprecate and eventually completely remove optimization pass from the `forge-fe` since it will be moved to `tt-mlir`.
- Modified the `explicate_unsqueeze` function to add dimensions at the beginning of tensors rather than at the end. This change was necessary because the optimization pass previously canceled out these unsqueezes, masking the dimension placement issue.

### DIFF
--- a/forge/test/models/onnx/vision/mobilenetv2/test_mobilenetv2.py
+++ b/forge/test/models/onnx/vision/mobilenetv2/test_mobilenetv2.py
@@ -57,7 +57,7 @@ def test_mobilenetv2_onnx(variant, forge_tmp_path):
 
     pcc = 0.99
     if variant == "mobilenetv2_050":
-        pcc = 0.96
+        pcc = 0.95
 
     fw_out, co_out = verify(
         inputs,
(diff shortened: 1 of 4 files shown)


forge/csrc/passes/explicate_unsqueeze.cpp
```diff
@@ -52,8 +52,8 @@ void explicate_unsqueeze(graphlib::Graph *graph)
                     auto rank = current_node->shape().size();
                     std::string name = to_be_unsqueeze->name() + "_" + eltwise->name() + "_unsqueeze_" +
                                        std::to_string(rank) + "_operand_0";
-                    auto attr = std::vector<graphlib::OpType::Attr>{0, ((int)rank)};
```
nvukobratTT, line 55: Let's add a sanity test to check this behavior. 

This way, we'll be able to confirm that this kind of issue won't appear in future changes (at least related to this logic).

nvukobratTT (approved):

PR #2389, opened 2025-06-25
Fix ttcore perf regression

odjuricicTT:
### Problem description
In tt-mlir tt dialect was renamed to ttcore. The argument type created when lowering to mlir was not renamed casing args to not have proper input/constant/param types which leads to const-eval and fusing not working in tt-mlir.

### What's changed
Change argument_type named attr key from `tt.argument_type` to `ttcore.argument_type`. Why is this a hardcoded string in the first place?

Resnet is now back from 12fps to 320fps 🚀


nvukobratTT:
Nice one!

PR #2393, opened 2025-06-25
Update group and priority for vovnet models

chandrasekaranpradeep:
The PR will update the vovnet onnx and pytorch models group and priority properties. It will set Group to Red and Priority to P1 for the ese_vovnet19b_dw.ra_in1k variant and for other variants will set Group to Generality and Priority to P2(i.e default)

### DIFF
--- a/forge/test/models/onnx/vision/vovnet/test_vovnet.py
+++ b/forge/test/models/onnx/vision/vovnet/test_vovnet.py
@@ -15,7 +15,15 @@ from test.models.pytorch.vision.vovnet.model_utils.model_utils import (
     preprocess_timm_model,
 )
 from test.models.pytorch.vision.vovnet.model_utils.src_vovnet_stigma import vovnet39, vovnet57
-from forge.forge_property_utils import Framework, Source, Task, ModelArch, record_model_properties
+from forge.forge_property_utils import (
+    Framework,
+    Source,
+    Task,
+    ModelArch,
+    record_model_properties,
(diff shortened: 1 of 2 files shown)


nvukobratTT (approved):

forge/test/models/onnx/vision/vovnet/test_vovnet.py
```diff
@@ -144,13 +152,22 @@ def generate_model_vovnet_imgcls_timm_pytorch(variant):
 @pytest.mark.parametrize("variant", ["ese_vovnet19b_dw.ra_in1k"])
 def test_vovnet_timm_onnx(variant, forge_tmp_path):
 
+    if variant == "ese_vovnet19b_dw.ra_in1k":
```
mstojkovicTT, line 155: Just a quick question, i see that the variant changed (`vovnet27s` -> `ese_vovnet19b_dw`), as well as changing from `onnx` to `pytorch` tracking in general. Is this a single case where we want to track `onnx` model instead of `pytorch`, or are we chaging how we track models? 
cc: @nvukobratTT

mstojkovicTT (approved):

forge/test/models/onnx/vision/vovnet/test_vovnet.py
nvukobratTT: Single case for now as it's available. However, we should see to push this one to be available through PyTorch as well.

Issue #2074, opened 2025-05-19
Replace default system descriptor with device-specific descriptor


odjuricicTT:
To my knowledge ffe created a "default" system descriptor and passes that to ttmlir instead of creating one from the existing device.

odjuricicTT:
@nvukobratTT @pilkicTT Is this still the case? If so this could be a problem for BH? If not please just close the issue.

PR #2403, opened 2025-06-26
Replace default system descriptor with device-specific descriptor

mstojkovicTT:
### Ticket
fixes: #2074

### Problem description
Instead of checking which architecture is being used, we default to `wormhole` architecture, which is problematic as we are planning to introduce `blackhole`.

### What's changed
Added call to `TTDevice` to get information about architecture that is being used, and passing it later on to get the right `config`.

### DIFF
--- a/forge/csrc/passes/lower_to_mlir.cpp
+++ b/forge/csrc/passes/lower_to_mlir.cpp
@@ -26,6 +26,7 @@
 #include "mlir/IR/Value.h"
 #include "mlir/IR/ValueRange.h"
 #include "passes/extract_unique_op_configuration.hpp"
+#include "runtime/tt_device.hpp"
 #include "utils/logger.hpp"
 
 // MLIR headers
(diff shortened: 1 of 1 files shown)


forge/csrc/passes/lower_to_mlir.cpp
```diff
@@ -170,8 +171,35 @@ class MLIRGenerator
     mlir::ModuleOp emit_mlir(tt::ForgeGraphModule &module)
     {
         graphModule_ = mlir::ModuleOp::create(get_module_location(module), module.name());
+
+        // Get the system and extract architecture information
```
acolicTT, line 175: Minor advice: don't comment every line, because it doesn't add any value. You have to do double job, and reader must read what you meant to do (comment) and what you actually did.
If something is not obvious and comment would clarify things, then it makes perfect sence :)

forge/csrc/passes/lower_to_mlir.cpp
```diff
@@ -170,8 +171,35 @@ class MLIRGenerator
...
+        {
+            ARCH tt_arch = system.devices[0]->arch;
+
+            // Map from ARCH to mlir::tt::Arch
+            switch (tt_arch)
+            {
+                case ARCH::WORMHOLE_B0: arch = mlir::tt::Arch::WormholeB0; break;
+                case ARCH::BLACKHOLE: arch = mlir::tt::Arch::Blackhole; break;
+                default:
+                    log_info("Unsupported architecture: {}, using default WormholeB0", to_string_arch(tt_arch));
```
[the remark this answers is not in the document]
acolicTT: Also, when you separate something to a function you can assert on invariant. You don't need to have deep if/else statements that are hard to read. You can simply check everything at the beginning as Nikola said, and user can see what is important and what is not.

acolicTT (approved):

forge/csrc/passes/lower_to_mlir.cpp
mstojkovicTT: Makes sense, non of the logic above is that complex that it needs comments to describe them :))

dgolubovicTT (approved):

PR #2421, opened 2025-06-30
Fix transformers update failures

chandrasekaranpradeep:
The PR will update transformers package version from 4.47.0 to 4.52.4 version and fix models failing with transformers update and also removed monkey patching for all the models expect bloom.

1. In llama decode tests, converted legacy past key values(i.e Tuple of Tuple of past key/values tensor) to dynamic cache before passing to the actual model inside the wrapper forward and converted the output past key values from dynamic cache to legacy past key values. In the previous transformer version(i.e 4.47.0) the conversion from legacy to dynamic and vice versa is taken care inside the LlamaModel nn module but in the latest transformer version they have removed the conversion and only expecting Cache type.

2.  The monkey patching is written to avoid tracing issues in the casual mask calculation. In the Llama decode tests, we have faced similar tracing issues. However for this case, we have calculated casual mask from attention mask using [_prepare_4d_causal_attention_mask](https://github.com/huggingface/transformers/blob/51f94ea06d19a6308c61bbb4dc97c40aabd12bad/src/transformers/modeling_attn_mask_utils.py#L302) function which was used by previous transformers version text generation models and then will pass the casual mask as input to the original model forward function since we have already calculated casual mask, it won't calculate casual mask again which helps to resolve the tracing issues. Created TextModelWrapper nn module which will take input_ids and attention mask as forward function inputs and then calculate casual mask from attention mask using [_prepare_4d_causal_attention_mask](https://github.com/huggingface/transformers/blob/51f94ea06d19a6308c61bbb4dc97c40aabd12bad/src/transformers/modeling_attn_mask_utils.py#L302)  and then pass attention mask and input_embeds as input to original model forward function. Replaced monkey patching present in all the text generation models(i.e LLama3, GPTNeo, Phi1, Phi2, Phi3, Phi4, NanoGPT, Gemma2, codegen, qwen) with TextModelWrapper and now some of the models like Llama, GPTNeo, Codegen are passing. Additionally I have updated all the text generation model to pass attention mask input which help to calculate attention only on non padding tokens.

3. Fixed some of the tokenizer issues(like`TypeError: PreTrainedTokenizerFast._batch_encode_plus() got an unexpected keyword argument 'pad_to_max_length'`) in all the text generation model.

4. In the ONNX text based models, we are using optimum package for generating onnx files for text generation models but not using framework models so removed redundant framework model created using transformer package codes.

5. In phi3 model, while applying rotary position embedding to query  states, we are creating two tensors  q_rot and q_pass by slicing the query states in the last dimension(i.e head dimension) which leads to the Index op error(i.e  `AssertionError: start < operandA.shape[dim]`) while running the generated ForgeModule. Below is the operations flow which leads to above index op assertion error
```

1. Slice Operation:
=======================================================


a)  For q_rot tensor creation,

query (1, 32, 256, 96) # (batch_size, num_of_heads, seq_len, head_dim)
             | 
Slice(start=0, end=96, dim=-1)
             |
q_rot (1, 32, 256, 96)


b) For q_pass tensor creation, 

query (1, 32, 256, 96) # (batch_size, num_of_heads, seq_len, head_dim)
             |
Slice(start=96, end=96, dim=-1)
             |
q_pass (1, 32, 256, 0) # empty tensor


2. Concat Operation:
=======================================================


q_rot (1, 32, 256, 96)    q_pass (1, 32, 256, 0) # empty tensor
                |________________|
                              |
                          Concat(dim=-1)
                             |
                   q_rot (1, 32, 256, 96) 

```

After slicing the query tensor to create q_pass, it will be empty tensor and then perform concat operation between q_rot and q_pass which is return q_rot 

Created DFPattern callback to remove the unwanted strideslice and concat which helps to resolve the Index op error and created sanity test for this pattern callback.


forge/test/models/pytorch/text/llama/test_llama3.py
```diff
@@ -29,134 +22,74 @@
...
+            ),
+            pytest.mark.out_of_memory,
+        ],
+    ),
+    pytest.param(
+        "meta-llama/Llama-3.1-8B-Instruct",
+        marks=[
+            pytest.mark.xfail,
+            pytest.mark.out_of_memory,
+        ],
```
vkovinicTT, line 61: Shouldn't we skip this version as well since it's 8B model?

forge/test/models/pytorch/text/llama/test_llama3.py
```diff
@@ -197,47 +130,35 @@ def test_llama3_causal_lm(variant):
...
         "meta-llama/Llama-3.2-3B-Instruct",
-        "meta-llama/Llama-3.2-1B-Instruct",
+        "meta-llama/Meta-Llama-3.1-70B",
         "meta-llama/Meta-Llama-3.1-70B-Instruct",
         "meta-llama/Llama-3.3-70B-Instruct",
-        "meta-llama/Meta-Llama-3.1-8B-Instruct",
     ]:
         pytest.xfail(reason="Requires multi-chip support")
-    elif variant == "meta-llama/Llama-3.1-8B-Instruct":
-        pytest.xfail(reason="Segmentation Fault")
```
vkovinicTT, line 209: Did you verify this?

forge/test/models/pytorch/text/llama/test_llama3.py
```diff
@@ -197,47 +130,35 @@ def test_llama3_causal_lm(variant):
...
-            truncation=True,
-        )
-    else:
-        inputs = tokenizer(
-            input_prompt,
-            return_tensors="pt",
-            max_length=256,
-            pad_to_max_length=True,
-            truncation=True,
-        )
```
vkovinicTT, line 235: Is there a strong reason why we changed this?

forge/test/models/pytorch/text/llama/test_llama3.py
```diff
@@ -369,4 +281,4 @@ def test_llama3_sequence_classification(variant):
     # post processing
     predicted_value = co_out[0].argmax(-1).item()
 
-    print(f"Prediction : {framework_model.config.id2label[predicted_value]}")
+    print(f"Prediction : {model.config.id2label[predicted_value]}")
```
vkovinicTT, line 284: Why don't we use `framework_model` here?

forge/test/models/pytorch/text/phi1/test_phi1.py
```diff
@@ -21,15 +21,15 @@
...
 
-variants = ["microsoft/phi-1_5"]
+variants = ["microsoft/phi-1", "microsoft/phi-1_5"]
 
 
 @pytest.mark.nightly
 @pytest.mark.parametrize("variant", variants)
-@pytest.mark.xfail
-def test_phi_1_5_causal_lm_pytorch(variant):
+def test_phi1_causal_lm_pytorch(variant):
```
vkovinicTT, line 32: In the `record_model_properties`, should we change `model=ModelArch.PHI1_5` -> `model=ModelArch.PHI1`?

forge/test/models/pytorch/text/phi2/test_phi2.py
```diff
@@ -37,14 +33,12 @@
...
-        marks=[
-            pytest.mark.skip(reason="Insufficient host DRAM to run this model (requires a bit more than 28 GB"),
-            pytest.mark.out_of_memory,
-        ],
+        marks=[pytest.mark.skip(reason="Insufficient host DRAM to run this model (requires a bit more than 31 GB")],
     ),
 ]
 
 
+@pytest.mark.out_of_memory
```
vkovinicTT, line 41: Do we really need this here? Btw since when have we introduced this mark?

forge/test/models/pytorch/text/phi2/test_phi2.py
```diff
@@ -156,22 +138,23 @@ def test_phi2_token_classification(variant):
     compiled_model = forge.compile(framework_model, sample_inputs=inputs, module_name=module_name)
 
     # Model Verification
-    verify(inputs, framework_model, compiled_model)
+    verify(inputs, framework_model, compiled_model, VerifyConfig(value_checker=AutomaticValueChecker(pcc=0.98)))
```
vkovinicTT, line 141: Correct me if I am wrong, but how do you know that we have to reduce `pcc` since we are skipping both models due to the out of memory problems?

forge/test/models/pytorch/text/phi4/test_phi4.py
```diff
@@ -46,13 +47,22 @@ def test_phi_4_causal_lm_pytorch(variant):
...
     framework_model.eval()
 
     # input_prompt
     input_prompt = "Africa is an emerging economy because"
-    inputs = tokenizer(input_prompt, return_tensors="pt")
+
+    inputs = tokenizer(
+        input_prompt,
+        return_tensors="pt",
+        max_length=256,
```
vkovinicTT, line 61: What do you think about reducing `max_length` to 128 everywhere?

forge/test/test_model_parts.py
```diff
@@ -110,3 +110,36 @@ def forward(self, attn):
 
     compiled_model = forge.compile(onnx_model, inputs)
     verify(inputs, model, compiled_model)
+
+
+@pytest.mark.push
```
vkovinicTT, line 115: Can you move this file into `forge/test/models` directory?

forge/test/models/pytorch/text/llama/test_llama3.py
chandrasekaranpradeep: We want to record model properties for this variant(i.e meta-llama/Llama-3.1-8B-Instruct) so we have added [xfail marker in pytest param](https://github.com/tenstorrent/tt-forge-fe/blob/fe0250711a28a48e990a5ff27a4744903cc3c0a3/forge/test/models/pytorch/text/llama/test_llama3.py#L63) and inside the test function we have added [pytest.xfail](https://github.com/tenstorrent/tt-forge-fe/blob/fe0250711a28a48e990a5ff27a4744903cc3c0a3/forge/test/models/pytorch/text/llama/test_llama3.py#L132) so that it won't execute.

forge/test/models/pytorch/text/llama/test_llama3.py
chandrasekaranpradeep: The model variant is not failing with segmentation fault, we are getting runtime error not able to allocate memory while running the compiled model and it consumes > 30 GB however it is red model so we need to record properties so I have added this variant in the [if condition](https://github.com/tenstorrent/tt-forge-fe/blob/fe0250711a28a48e990a5ff27a4744903cc3c0a3/forge/test/models/pytorch/text/llama/test_llama3.py#L132)

forge/test/models/pytorch/text/llama/test_llama3.py
chandrasekaranpradeep: While porting [tt-torch models to tt-forge-fe](https://github.com/tenstorrent/tt-forge-fe/commit/ff19e264324a33ea966ec61460db7351105d52b8#diff-1a7d9481d64f313a8b08bcd92b99edd602d546b7fa83ee1b1a2236299795f76b), the have retained the tokenizer.encode_plus method  with max_length = 32 for the meta-llama/Llama-3.2-3B variant but there is no specific reason for it so I have changed it to use common max_length for the all the model variants.

forge/test/models/onnx/text/llama/test_llama3.py
```diff
@@ -162,29 +58,19 @@ def test_llama3_causal_lm_onnx(variant, forge_tmp_path):
...
-            pad_to_max_length=True,
-            truncation=True,
-        )
+    inputs = tokenizer(
+        input_prompt,
+        return_tensors="pt",
+        max_length=128,
+        padding="max_length",
+        truncation=True,
+    )
```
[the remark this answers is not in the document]
chandrasekaranpradeep: Addressed here - https://github.com/tenstorrent/tt-forge-fe/pull/2421#discussion_r2182159910. In [this commit](af831285f8f1d68d5246f56a113a87c5189a47bb631047d6215f8cd60955c1da), they have copied the code from pytorch to onnx

forge/test/models/pytorch/text/phi4/test_phi4.py
chandrasekaranpradeep: Reduced the max_length

chandrasekaranpradeep:
> Did you run all the tests that you have changed, since this is massive change?

I have triggered the nightly model [passing](https://github.com/tenstorrent/tt-forge-fe/actions/runs/15997003560) and [xfailing](https://github.com/tenstorrent/tt-forge-fe/actions/runs/16003335060) pipeline to check the tests  and also addressed the failure.

forge/forge/tvm_calls/relay/op/forge_passes.py
```diff
@@ -4990,6 +4990,30 @@ def callback(self, pre, post, node_map):
...
+        super().__init__(rewrite_once=False, require_type=True)
+        self.slice_act = wildcard()
+        self.concat_act = wildcard()
+        self.slice = is_op("strided_slice")(self.slice_act)
+        self.pattern = is_op("concatenate")(
+            is_tuple([self.concat_act, self.slice]) | is_tuple([self.slice, self.concat_act])
+        )
+
+    def callback(self, pre, post, node_map):
+        act = node_map[self.concat_act][0]
```
vkovinicTT, line 5004: If we are pattern matching both `concat(slice, tensor)` and `concat(tensor, slice)` we can't garantee that:
```
act = node_map[self.concat_act][0]
```
So we need to take tensor that doesn't have 0 on `concat_axis`.

forge/forge/tvm_calls/relay/op/forge_passes.py
```diff
@@ -4990,6 +4990,30 @@ def callback(self, pre, post, node_map):
...
+        slice_axes = list(node_map[self.slice][0].attrs.axes)
+        concat_axis = int(node_map[self.pattern][0].attrs.axis)
+        slice_out_shape = list(node_map[self.slice][0].checked_type.shape)
+        concat_out_shape = list(node_map[self.pattern][0].checked_type.shape)
+        slice_axes = [axes - len(slice_out_shape) if axes >= 0 else axes for axes in slice_axes]
+        if concat_axis >= 0:
+            concat_axis -= len(concat_out_shape)
+        if concat_axis in slice_axes and slice_out_shape[concat_axis] == 0:
+            return act
```
vkovinicTT, line 5013: I think that we can simplify this by just checking:
```
slice_out_shape = list(node_map[self.slice][0].checked_type.shape)
if slice_out_shape[concat_axis] == 0:
            return act
```
since this will be the case for all the concats that concatenate along dim which is 0, regardless of whether we have slices along that dim or not. Correct me if I am wrong.

vkovinicTT (changes requested):

forge/test/models/pytorch/text/phi2/test_phi2.py
vkovinicTT: Yep, thanks :)

forge/forge/tvm_calls/relay/op/forge_passes.py
chandrasekaranpradeep: Understood. I have updated the pattern callback to return the act which doesn't have 0 on concat_axis

forge/forge/tvm_calls/relay/op/forge_passes.py
chandrasekaranpradeep: Updated the pass.

vkovinicTT (approved):

chandrasekaranpradeep:
@vkovinicTT  @nvukobratTT  Triggered the nightly models [passing](https://github.com/tenstorrent/tt-forge-fe/actions/runs/16071470266/job/45357890366) and [xfailing](https://github.com/tenstorrent/tt-forge-fe/actions/runs/16053669920) pipeline in this PR and all the test are passed and for crashed tests I have tested locally

nvukobratTT (approved):

PR #2435, opened 2025-07-02
Switching ops functions to separate namespace.

acolicTT:
With current implementation are defining all ops functions within Op class and differentiate them by prefix (e.g. abs_eval). This change introduces namespace for every op and all functions that op implements should be implemented in op
namespace (e.g. abs::eval).


vkovinicTT (approved):

pilkicTT (approved):
`-Wgnu-zero-variadic-macro-arguments`

PR #2445, opened 2025-07-03
fix build file

einnesTT:
### Ticket
New TT-Forge docs have issue where it's referencing the wrong file name for the build of the html docs

### Problem description
The wrong file is referenced, so the build page appears blank. 

### What's changed
Remove the build page, we already have the new build page under a different title now.
This change will remove the blank page issue, and allow people to focus on the new build page.

### Checklist


docs/src/SUMMARY.md
```diff
@@ -7,8 +7,7 @@
     - [Getting Started with Building from Source](./getting_started_build_from_source.md)
 - [Architecture Overview](./architecture_overview.md)
 
-# Project setup
-- [Building](./build.md)
+# Project Setup
```
pilkicTT, line 10: We should maybe rename this section to something like: `Developing TT-Forge-Fe`

pilkicTT (commented):
Please change the `README.md` to point to the correct location as well...

PR #2449, opened 2025-07-04
[watchdog] show timout exceeded message

pilkicTT:
When running the tests with pytest capture enabled, the watchdog timeout message is not being showed. This is because the watchdog will abort the process, so pytest doesn't print out the captured logs.

If capture enabled, suspend the capture allowing us to print out the message directly. Useful for CI, since i suspect there are watchdog timeouts occuring in recent nightlies...


nsumrakTT (approved):
Thank you! I was trying to find a way to see this print (unsuccessfully).

Issue #2378, opened 2025-06-23
Migrating operations from python to cpp.


acolicTT:
This document contains an overview of migrating forge operations from python to c++ code.

Our long term goal is to migrate most of the compiler to c++ over time. In the ideal case, after the all transitions, only the input layer necessary for working with the rest of the world will remain in python, while on the other hand, all the compiler implementations will be in c++.

This change is only one part of that transition in which we change the functionality of operations (eval, shape, backward, decompose, etc.). This way, we will reduce the number of jumps between the python interpreter and the compiled cpp code.

Current implementations of ops are in /forge/csrc/graph_lib/node_types.hpp (OpType, current cpp op definition), forge/forge/op/interface.py (interface definition for ops), forge/forge/op/eval folder (various calculations).

Some time ago, an initiative was started to transfer all ops currently implemented in python to a new interface: https://yyz-gitlab.local.tenstorrent.com/tenstorrent/pybuda/-/issues/1675
That attempt failed, but something was still done, so the current situation is that some ops are implemented with a new interface, while the majority are based on the old one, but they all have in common that they are written in python.

With the proposed change, we will support the existing functionalities of both, while at the same time we will be able to migrate ops one by one to c++ code.

The new ops are implemented in the forge/csrc/ops/ folder, and are statically linked.
Each op contains an operation type and attributes. Each operation is of the same size, which allows us to have only one Op class. Also, with a simple runtime dispatch, we can delegate functionality based on type of operation.
Since the operation type is dense enum, we practically have free dispatching in the runtime, since the compiler creates a lookup table for the switch in that case. With this, we avoided polymorphism, which would complicate the manipulation of operations in the code itself. With the polymorphism we would have to allocate each op on the heap separately from the op node, we would have to use unique_ptr in order to put ops in a container etc.

It should be noted that the main goal of this implementation is not performance. If it becomes important for us in the future, the first thing to do is to remove the map that holds the attributes and design a new mechanism (I vote for vector pairs). Another thing that should be done is to create attribute key enum and map named attributes based on that enum, instead on string as we do now.

As for the automatic dispatching that we get with polymorphism, we should have the same result with a simple switch/case, as I already said. Also, since all parameters are just passed as received, we should also get TCO (tail call optimization) for free. LLVM also uses switch/case for its ops, and they are generating our binaries, so we should have no problems there
[...]

PR #2455, opened 2025-07-04
Replaced string comparison to op type comparison for ops.

acolicTT:
In order to determine op type, we were comparing strings. Since op type enum is introduced, this change replaces string comparisons with op type comparisons.

This change is part of ops migration from python to cpp code.


pilkicTT (commented):
can you please split this into two PRs? one for changing the fields, the other for string -> op type comparison...

pilkicTT (approved):

PR #2457, opened 2025-07-05
[ops] fix malformed index op

pilkicTT:
With commit 20033e3, a regression was introduced in `decompose_nd_reshape_split`. This was observed in the latest nightly.

Failing nightly test:
`test_mamba[state-spaces/mamba-790m-hf]`

The modified `index` op didn't have its named attributes set, causing a failure during MLIR lowering.

This change adds the missing named attributes and updates the mamba test duration, since this was also causing issues in nightly where the watchdog would terminate the test due to the timeout.


vkovinicTT (approved):
Thanks for catching this :))

PR #2491, opened 2025-07-10
Add torch.where operator sweeps test

vobojevicTT:


forge/test/operators/pytorch/eltwise_nary/test_where.py
```diff
@@ -0,0 +1,202 @@
...
+            random_seed=sum(shape) + 3,
+        )
+        self.register_buffer("constant1", self.c1)
+        self.register_buffer("constant2", self.c2)
+        self.register_buffer("constant3", self.c3)
+
+    def forward(self, x, y, z):
+        condition = self.c1 > self.treshold
+        where_result = self.operator(condition, self.c2, self.c3)
+        return where_result + x + y + z  # consuming the inputs to match the number of operands
```
dgolubovicTT, line 85: Why can't we have only 1 input not x,y,z?

vobojevicTT: We use the same `verify` function for all 3 models (`from-another-op, from-host, const-eval-pass`) and pass `input_shapes` as a list to the `verify` function. That's why we also use inputs x, y, and z in the const-eval-pass model here, as we do in the other 2 models.

PR #2584, opened 2025-07-16
Enable param groups in optimizers

### LINKED ISSUE #1693: Parameter groups
We want to be able to set different hyperparameters (e.g. learning rate, optimizer alpha, beta) for different parameters/parameter groups (proven to improve results in practice in various experiments).

pglusacTT:
### Ticket
Closes #1693

### Problem description
Optimizers do not have parameter groups that can have different hyperparameters which is a common use case for training ML models.

### What's changed
SGD and Adam were updated to support param groups.
User can set default hyperparameters and override them by passing them to the optimizer like in the example below:
```python
    tt_params = list(framework_model.named_parameters())
    linear1_tt = [name for name, p in tt_params if "linear_relu_stack.0" in name]
    ...
    # Create parameter groups
    tt_param_groups = [
        {'params': linear1_tt, 'lr': 1e-5},
        {'params': linear2_tt, 'lr': 1e-4},
        {'params': linear3_tt, 'lr': 1e-3}
    ]

    # Create optimizers with parameter groups
    tt_optimizer = forge.optimizers.Adam(
        learning_rate=0.005,
        bias_correction=True,
        parameters=tt_param_groups
    )
```


pglusacTT:
Note: The solution uses names to match nodes with param groups which differs from torch but I think that for ffe this is the best way to do it. Feel free to give your suggestions.

PR #2595, opened 2025-07-17
Implemented template for all remaining operations in cpp.

acolicTT:
Since we have a lot of merge conflicts while working on different ops, are wasting a lot of time waiting on PR actions to be completed after conflicts resolution.

To resolve that, all op files are created in this change and all dispatching is done too. Also, all is_tm, is_eltwise etc. checks are implemented at once.


vkovinicTT (approved):

pilkicTT (approved):
after some discussion offline, we've decided to go with this change to speed us up...

PR #2346, opened 2025-06-19
Upgrades jax to 0.6.1

saiarthiraguram:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-forge-fe/issues/2327)

### Problem description
To address below issue 
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
tt-forge-fe 0.1.0.dev20250617 requires jax==0.4.30, but you have jax 0.6.1 which is incompatible.

```

PR #2600, opened 2025-07-18
Revert "Upgrades jax to 0.6.1 (#2346)"

vcanicTT:
This reverts commit 300608b1c5763963aaa90f8e58b2f1b486700f0b.


acolicTT (approved):

acolicTT:
Can you add a comment why are we reverting this? Or do some investigation?
Also, if we are doing revert, why are we updating pyarrow in the same change?

PR #2617, opened 2025-07-21
Add enable_fusing_conv2d_with_multiply_pattern flag

jserbedzijaTT:
### Problem description
To enable fusing by default in `tt-mlir`, we need to disable the `Conv2dWithMultiply` pattern by default. In this PR, I’m enabling it specifically for ResNet due to the performance regression when it’s disabled.

### Checklist

### DIFF
--- a/forge/test/benchmark/benchmark/models/resnet_hf.py
+++ b/forge/test/benchmark/benchmark/models/resnet_hf.py
@@ -114,7 +114,11 @@ def test_resnet_hf(training, batch_size, data_format, input_size, channel_size, 
 
     # Turn on MLIR optimizations.
     compiler_cfg.mlir_config = (
-        MLIRConfig().set_enable_optimizer(True).set_enable_fusing(True).set_enable_memory_layout_analysis(True)
+        MLIRConfig()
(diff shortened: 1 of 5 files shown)


vcanicTT (approved):

forge/test/benchmark/benchmark/models/resnet_hf.py
```diff
@@ -114,7 +114,11 @@ def test_resnet_hf(training, batch_size, data_format, input_size, channel_size,
 
     # Turn on MLIR optimizations.
     compiler_cfg.mlir_config = (
-        MLIRConfig().set_enable_optimizer(True).set_enable_fusing(True).set_enable_memory_layout_analysis(True)
+        MLIRConfig()
+        .set_enable_optimizer(True)
+        .set_enable_fusing(True)
```
odjuricicTT, line 119: Is this on by default now?

odjuricicTT (approved):

jserbedzijaTT:
Will merge this before the 24-hour mark because it is causing a perf regression for ResNet.

PR #2619, opened 2025-07-21
[OP] Tensor and other tensor creation support

mstojkovicTT:
### Problem description
Right now, there is no easy way to create tensor in cpp when writing ops.

### What's changed
Added support for create_constant that supports `at::Tensor` from torch


mstojkovicTT:
Still in progress...

pilkicTT (commented):
can we rewire everything to use the new method? and remove the old one?

acolicTT (approved):
Looks good. You can just add a little wrapper to avoid always passing the same parameters for this function in every call. Look at autograd::create_op for example :)

PR #2635, opened 2025-07-23
Fix verify() sharing mutable default VerifyConfig across tests

### LINKED ISSUE #2634: Emitc is running on non RED models
Emitc should be running only for RED models (Added in [PR](https://github.com/tenstorrent/tt-forge-fe/commit/291682249c2f4454989733a7deab3d85c4c7a835))

However, it seems like it is running also for some non RED models due to some leakage in CI. This can be seen on [EmitC superset](https://superset.tenstorrent.com/superset/dashboard/210/?native_filters_key=mUEPRkANOds). Also nightly tests are failing because some operations in non RED models are not supported in emitc.

dgolubovicTT:
### Ticket
Closes #2634

### Problem description
verify_cfg was shared across tests that didn't pass it inside compile. This was because verify_cfg was default argument of verify() function, and python evaluates default arguments once when function is defined. If the argument is mutable and changed in one test, the change will carry over to later tests...

### What's changed
  Replaced default argument `verify_cfg=VerifyConfig()` with `None`
  to avoid shared state across test calls.
  Prevents unintended carry-over of modified config values.

### DIFF
--- a/forge/forge/verify/verify.py
+++ b/forge/forge/verify/verify.py
@@ -259,7 +259,7 @@ def verify_backward(
     compiled_output: torch.Tensor,
     framework_model: torch.nn.Module,
     compiled_model: CompiledModel,
-    verify_cfg: VerifyConfig = VerifyConfig(),
+    verify_cfg: VerifyConfig = None,
 ):
     """
     Performs verification of a compiled model by comparing its outputs against a reference framework model.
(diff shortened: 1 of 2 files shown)


acolicTT (approved):

svuckovicTT:
Such a cool (and annoying) bug! Internet suggests using a linter to detect mutable default args. Maybe it would be cool to add this check to `pre-commit`?

dgolubovicTT:
> Such a cool (and annoying) bug! Internet suggests using a linter to detect mutable default args. Maybe it would be cool to add this check to `pre-commit`?

Good idea. flake8 has support for that. I added this to follow up [issue](https://github.com/tenstorrent/tt-forge-fe/issues/2636)

PR #2637, opened 2025-07-23
Implemented dropout op in cpp.

acolicTT:
This change is part of ops migration from python to cpp.


forge/forge/op/eltwise_unary.py
```diff
@@ -521,36 +521,6 @@ def LogicalNot(name: str, operandA: Tensor) -> Tensor:
     return op("logical_not", name, operandA).get_tensor()
 
 
-def Dropout(name: str, operandA: Tensor, p: float = 0.5, training: bool = True, seed: int = 0) -> Tensor:
```
pilkicTT, line 524: we should leave this, no?

forge/forge/op/eval/forge/eltwise_unary.py
```diff
@@ -24,22 +24,13 @@ def eval(type, attr, ops):
...
         or (type == "cumsum" and len(attr) == 2)
-        or (type == "dropout" and len(attr) == 3)
         or (type == "pow" and len(attr) == 1)
     ), "Eltwise unary should have no attributes, execpt for clip, relu, cumsum, dropout and pow"
 
     t_ops = to_torch_operands(*ops)
 
     original_types = [o.dtype for o in t_ops]
 
-    if type == "dropout":
```
pilkicTT, line 35: shouldn't we remove the backward and shape implementations as well?

PR #2672, opened 2025-07-28
Enable cache tensor to be input rather than constant.

### LINKED ISSUE #2671: Enable cache tensor to be input rather than constant
We need to use StaticCache object to manage Llama kv cache. On the other hand, TVM requires us to have only tensors as inputs to modules (so having StaticCache object as input won't work). Solution is to create StaticCache within the module, but overwrite k and v tensors with ones provided on input. Therefore, we end up with StaticCache created within model, but compiler sees its k,v tensors as inputs.

dgolubovicTT:
### Ticket
Closes #2671

### Problem description
We need to use StaticCache object to manage Llama kv cache. On the other hand, TVM requires us to have only tensors as inputs to modules (so having StaticCache object as input won't work). Solution is to create StaticCache within the module, but overwrite k and v tensors with ones provided on input. Therefore, we end up with StaticCache created within model, but compiler sees its k,v tensors as inputs.

### What's changed
- Make cache input by overwriting constant K,V tensors with input cache tensors in forward of module. This trick enables us to still have only tensors as inputs (required by TVM and torch.jit.trace) while using StaticCache abstraction to handle K,V cache update. Previously, we only created StaticCache object within function forward, which resulted in K,V cache tensors being seen as constants from compiler.
- Add small test which tests only StaticCache update on device without all other attention parts.

Caught a bug while doing this - input tensors were changed due to model having in-place operations.
Solution:
- Clone input tensors in verify before running framework model.

### DIFF
--- a/forge/forge/verify/verify.py
+++ b/forge/forge/verify/verify.py
@@ -14,11 +14,17 @@ from loguru import logger
 from forge.forgeglobal import align_up_tile
 
 import torch
+import tensorflow as tf
+import paddle
+import jax
+import keras
 from forge.tensor import to_pt_tensors
 
 from ..tensor import (
     FrameworkTensor,
     Tensor,
+    TensorFromPytorch,
+    TensorFromTrace,
     pytorch_dtype_to_forge_dataformat,
     forge_dataformat_to_pytorch_dtype,
 )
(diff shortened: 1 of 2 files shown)


forge/forge/verify/verify.py
```diff
@@ -401,8 +448,9 @@ def verify(
         raise TypeError(
             f"Compiled model must be of type {verify_cfg.compiled_model_types}, but got {type(compiled_model)}"
         )
-
-    fw_out = framework_model(*inputs)
+    # perform forward on copy of inputs in case you have inplace operations
+    fw_inputs = clone_framework_tensors(inputs)
```
nvukobratTT, line 452: As we're cloning the tensor, let's delete it once we use it. This will speed up memory cleanup a bit.

nvukobratTT (approved):
Minor comment, approving

forge/forge/verify/verify.py
dgolubovicTT: It won't help too much unless we are running tight on memory.  Python will cleanup memory as soon as `fw_inputs `goes out of scope. As you said it can seep up cleanup a bit, so will do.

PR #2678, opened 2025-07-29
Implemented buffer op in cpp.

acolicTT:
This change is part of ops migration from python to cpp code.


pilkicTT (commented):
i think we don't have a need for this op anymore... are there any uses of it in the codebase?

acolicTT:
Closed PR, since buffer is not used anymore. Instead, it will be deleted in next iterations.

PR #2639, opened 2025-07-23
Fix yolov5 pytorch model regression

chandrasekaranpradeep:
Fixes #2631

### Summary:

The PR will resolve `AssertionError: only single axis transpose supported at this time, decompose in tvm` thrown in Yolov5 models. The issues raised due to elimination of sanity run in [extract_framework_model_outputs](https://github.com/tenstorrent/tt-forge-fe/blob/718596db22c4e6468a0fc68b87651143cbd4092c/forge/forge/tvm_calls/forge_utils.py#L28) function present in `forge/forge/tvm_calls/forge_utils.py` because [in this commit](https://github.com/tenstorrent/tt-forge-fe/commit/4fb30b801650a2190c523a45173fcb9411106ed6#diff-f815ebe537268e224d2f90dc51a797f081a4aa3a3e2b4186a0a597356a7be619L28), we have modified to extract framework outputs only when `verify_tvm_compile` is set to `True`. Since we have skipped the sanity run, the generated TVM relay IR is different which leads to multiple axes transpose in `ReplaceYolov5Perf` pattern callback. To resolve this issue, I modified the tests to execute sanity run before forge compilation. The `ReplaceYolov5Perf` pattern callback is implemented to remove index and concatenate ops that are unsupported by HW due to its dimension with masking in PyBuda. But concat and index ops are supported in Forge, so commented the pattern. With all these changes all the yolov5 tests(i.e includes regression and xfail) are passed.

### Without Sanity Run (Relay IR before and after ReplaceYolov5Perf Callback):
[Relay_IR_before_ReplaceYolov5Perf_callback_without_sanity_run.log](https://github.com/user-attachments/files/21411394/Relay_IR_before_ReplaceYolov5Perf_callback_without_sanity_run.log)
[Relay_IR_after_ReplaceYolov5Perf_callback_without_sanity_run.log](https://github.com/user-attachments/files/21411391/Relay_IR_after_ReplaceYolov5Perf_callback_without_sanity_run.log)

### With Sanity Run (Relay IR before and after ReplaceYolov5Perf Callback):
[Relay_IR_before_ReplaceYolov5Perf_callback_with_sanity_run.log](https://github.com/user-attachments/files/21411393/Relay_IR_before_ReplaceYolov5Perf_callback_with_sanity_run.log)
[...]

PR #2694, opened 2025-07-30
Fix ConvertEmulatedDtypes and remove sanity run before forge compilation in yolov5

chandrasekaranpradeep:
The yolov5 model was regressed with `AssertionError: only single axis transpose supported at this time, decompose in tvm` and fixed it in this [PR](https://github.com/tenstorrent/tt-forge-fe/pull/2639) by performing sanity before forge compilation  and also noticed generated relay IR difference between with and without sanity run and found out constant folding is happening when we do compilation with sanity and also if the sanity run is skipped, framework model output remains in float32 rather than  bfloat16. After further analysis found out, the yolov5 model has Detect nn module which has grid and anchor_grid tensor initialised in the init function and at first sanity, grid and anchor_grid tensor are computed using make_grid when the grid and input height and width are not same(i.e [if self.dynamic or self.grid[i].shape[2:4] != x[i].shape[2:4]:](https://github.com/ultralytics/yolov5/blob/2540fd4c1c2d9186126a71b3eb681d3a0a11861e/models/yolo.py#L101))  and at the second inference it will not compute grid and anchor_grid tensor since the grid is already computed and matches with input height and width which leads to different graph IR generation so it is not exactly constant folding. The framework model output remains in float32 instaed of bfloat16 this problem was also raised in grid and anchor_grid present in the Detect module. If don't perform sanity run, the model is in bfloat16 and in the ConvertEmulatedDtypes class converts params, buffers, inputs and tensor peresent in the init function from bfloat16 to float32 for TVM Relay IR generation but the grid and anchor_grid are list of tensor in bfloat16 dtype present in the init function which is not converted from bfloat16 to float16 and and at the same time we did run sanity so at the jit trace, the grid and anchor are computed using make_grid since the grid and input height and width are not same which converts the grid and anchor_grid from bfloat16 to float32 because the grid and anchor_grid dtypes are determined by anchors buffer which is in float32 and after TVM graph generation all the params, buffers, inputs and tensors present in the init function are converted back to bfloat16 from float32 but the grid and anchors are not converted and it will be in float32 which leads model output remains in float32 instaed of bfloat16 in forge verification. To resolve this issues, added support for converting the list/tuple of tensor present in the init function from bfloat16 to float32 before TVM graph generated and vice versa in the ConvertEmulatedDtypes class. Now all the yolov5 model tests runs wihout sanity run before compilation

### DIFF
--- a/forge/forge/tvm_calls/forge_compile.py
+++ b/forge/forge/tvm_calls/forge_compile.py
@@ -470,6 +470,15 @@ class ConvertEmulatedDtypes:
                         key = f"{mod_name}.{attr}" if mod_name else attr
                         self.attr_dfs[key] = (mod, attr, val.dtype)  # store module, attr name, and dtype
                         setattr(mod, attr, val.to(self.fallback))  # overwrite with fallback dtype
+                    elif isinstance(val, (list, tuple)) and all(isinstance(v, torch.Tensor) for v in val):
+                        original_dtypes = [v.dtype for v in val]
+                        if any(dt in self.emulated_dfs for dt in original_dtypes):
+                            key = f"{mod_name}.{attr}" if mod_name else attr
+                            self.attr_dfs[key] = (mod, attr, original_dtypes)
+                            converted_val = [v.to(self.fallback) if v.dtype in self.emulated_dfs else v for v in val]
+                            if isinstance(val, tuple):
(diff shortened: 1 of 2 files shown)


saiarthiraguram:
Can you add the log for this change @chandrasekaranpradeep

PR #2703, opened 2025-07-31
Implemented cumulative sum op in cpp.

acolicTT:
This change is part of migrating ops to cpp.
For details, checkout: https://github.com/tenstorrent/tt-forge-fe/issues/2378


forge/csrc/ops/op_cumulative_sum.cpp
```diff
@@ -25,16 +24,27 @@ using namespace graphlib;
...
 {
     TT_DBG_ASSERT(op.type() == OpType::CumulativeSum, "Wrong op type.");
-    return op.base_shape(old_op_type, in_shapes);
+    TT_ASSERT(in_shapes.size() == 1, "Cumulative sum should have one input.");
+    TT_ASSERT(op.attrs().size() == 1, "Cumulative sum should have one attribute.");
+
+    return {Shape::create(in_shapes[0]), {}};
 }
 
+/**
```
nvukobratTT, line 43: Do we need these block comments for bwd (in this detail) if we're not implementing it? 

In sum, this might be misleading, as there are multiple ways to calculate gradients for the cumsum op. Some of them might even be available today. Therefore, I'm just thinking is having detailed comments like this one here without actual implementation more beneficial or misleading for future changes.

forge/csrc/test/ops/test_eltwise.cpp
```diff
@@ -133,6 +134,7 @@ std::vector<tt::ops::Op> get_unary_eltwise_ops()
     return {
         tt::ops::OpType::Abs,
         // tt::ops::OpType::Clip, // Tested separately.
```
nvukobratTT, line 136: Do we need commented-out lines here?

nvukobratTT (approved):
Minor comments, approving.

forge/csrc/test/ops/test_eltwise.cpp
acolicTT: It is just for easier orientation, since these ops are not tested as a part of eltwise unary operations, because they have attributes.

forge/csrc/ops/op_cumulative_sum.cpp
acolicTT: Makes sense. I will remove it.

PR #2723, opened 2025-08-04
[Debt] Removed unused code from eltwise binary

vkovinicTT:
### Problem description
Code from the decomposition of the `eltwise_binary` ops was unused, hence I deleted it.


forge/forge/op/eval/forge/eltwise_binary.py
```diff
@@ -0,0 +1,126 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
```
acolicTT, line 1: Don't retrieve deleted files.

forge/forge/op/eval/forge/resize.py
```diff
@@ -0,0 +1,330 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
```
acolicTT, line 1: Same here.

PR #2789, opened 2025-08-11
Update K,V tensors on device and hold them during decode loop.

### LINKED ISSUE #2788: Enable K,V cache tensors staying on device in decode loop.
We want to keep K,V cache tensors on device while decoding. Therefore, we don't have to push them in each forward call and they are updated on device.

dgolubovicTT:
### Ticket
Closes #2788
### Problem description
We want to keep K,V cache tensors on device while decoding. Therefore, we don't have to push them in each forward call and they are updated on device.

### What's changed
- Enable Ctensor as input to compiled model. Ctensor stays alive on device during decode loop. Pass K,V cache tensors as Ctensors and they will stay on device during decode loop. They will get updated in each pass.
- Pass K,V tensors to wrapper and then overwrite StaticCache values with ones from input (inside wrapper forward ).
- Add test for static cache update on device: test_decode_cache_on_device


forge/test/mlir/llama/tests/test_llama_decode.py
```diff
@@ -416,10 +486,180 @@ def test_llama_prefill_on_cpu_decode_on_tt_cache(model_path, run_on_tt_device, n
...
+    current_seq_len = past_seq_len + seq_len  # Total actual sequence length
+
+    # Create attention mask for current sequence length
+    decode_attention_mask = torch.zeros(batch_size, current_seq_len, dtype=dtype)
+
+    # Expand to 4D: [batch_size, 1, 1, seq_len]
+    decode_attention_mask = decode_attention_mask.unsqueeze(1).unsqueeze(2)
+
+    # Pad attention mask to max_seq_len
+    pad_len = max_seq_len - current_seq_len
```
nvukobratTT, line 570: This isn't handled when static cache is set up?

forge/test/mlir/llama/tests/test_llama_decode.py
```diff
@@ -416,10 +486,180 @@ def test_llama_prefill_on_cpu_decode_on_tt_cache(model_path, run_on_tt_device, n
...
+    current_seq_len = current_seq_len + 1
+
+    start = time.perf_counter()
+
+    for decode_step in range(max_new_tokens):
+        logits = compiled_model(*decode_inputs)
+
+        # Get next token
+        logits_tensor = logits if not isinstance(logits, list) else logits[0]
+        next_token_logits = logits_tensor[:, -1, :]
```
nvukobratTT, line 635: Ideally, all of these should be done on the device as well, instead of the CPU. For now, I don't think we should push it further that this :))

nvukobratTT (approved):
Minor comments, approving

forge/test/mlir/llama/tests/test_llama_decode.py
dgolubovicTT: Can you clarify a bit more? If you refer to attention mask, we still need to provide it to te model forward along with cache and cache_position...

PR #2800, opened 2025-08-12
Fix resize2d

chandrasekaranpradeep:


forge/csrc/ops/op_resize_2d.cpp
```diff
@@ -155,19 +144,25 @@ void decompose_initial(
     TT_DBG_ASSERT(op.type() == OpType::Resize2d, "Wrong op type.");
     TT_ASSERT(inputs.size() == 1, "Resize2d expects 1 input");
 
-    auto sizes = op.attr_as<std::vector<int>>("sizes");
+    auto sizes = op.attr_as<std::vector<int32_t>>("sizes");
```
mstojkovicTT, line 147: I dont see a reason to change this to `int32_t` when in all other ops its `int`

forge/csrc/ops/op_resize_2d.cpp
```diff
@@ -155,19 +144,25 @@ void decompose_initial(
...
-    std::string resize_method = op_common::get_resize_method(method);
+    bool align_corners = op.attr_as<bool>("align_corners");
 
     NodeContext result = inputs[0];
     Shape input_shape = result.shape;
 
     // Determine whether to use upsample or downsample
-    bool upsample = channel_last ? sizes[0] >= static_cast<int>(input_shape[input_shape.size() - 3])
-                                 : sizes[0] >= static_cast<int>(input_shape[input_shape.size() - 2]);
+    bool upsample_h = channel_last ? sizes[0] >= static_cast<int>(input_shape[input_shape.size() - 3])
```
mstojkovicTT, line 157: Same thing as above

forge/csrc/ops/op_resize_2d.cpp
```diff
@@ -155,19 +144,25 @@ void decompose_initial(
...
+                                   : sizes[0] >= static_cast<int>(input_shape[input_shape.size() - 2]);
+    bool upsample_w = channel_last ? sizes[1] >= static_cast<int>(input_shape[input_shape.size() - 2])
+                                   : sizes[1] >= static_cast<int>(input_shape[input_shape.size() - 1]);
+
+    if ((upsample_h and (!upsample_w)) or ((!upsample_h) and upsample_w))
```
mstojkovicTT, line 162: use `&&` and `||` instead of `or` and `and`

forge/csrc/ops/op_resize_2d.cpp
```diff
@@ -119,14 +77,45 @@ std::tuple<Shape, std::vector<DimBroadcast>> shape(
...
+
     if (channel_last)
     {
+        if (upsample_h and upsample_w)
+            TT_ASSERT(
+                (size_h % static_cast<int>(input_shape[input_shape.size() - 3]) == 0) and
+                    (size_w % static_cast<int>(input_shape[input_shape.size() - 2]) == 0),
+                "Only support upsample with integer scale factor");
+        else
+            TT_ASSERT(
```
mstojkovicTT, line 99: This is not correct, because `%` operation isnt commutative, there is a chance that either `upsample_h` or `upsample_w` are not `true`. Also we are currently not supporting `downsample` at all, so you should just focus on case `resize2d` will be decomposed to `upsample2d`

forge/csrc/ops/op_resize_2d.cpp
```diff
@@ -119,14 +77,45 @@ std::tuple<Shape, std::vector<DimBroadcast>> shape(
...
     }
     else
     {
+        if (upsample_h and upsample_w)
+            TT_ASSERT(
+                (size_h % static_cast<int>(input_shape[input_shape.size() - 2]) == 0) and
+                    (size_w % static_cast<int>(input_shape[input_shape.size() - 1]) == 0),
+                "Only support upsample with integer scale factor");
+        else
+            TT_ASSERT(
```
mstojkovicTT, line 115: Same comment for `downsample`

forge/csrc/ops/op_upsample_2d.cpp
```diff
@@ -27,34 +27,53 @@ at::Tensor eval(const graphlib::OpType &old_op_type, const Op &op, const std::ve
...
-    {
-        activations = activations.permute({0, 3, 1, 2});
-    }
+        activation = activation.permute({0, 3, 1, 2});
 
     torch::nn::functional::InterpolateFuncOptions options = torch::nn::functional::InterpolateFuncOptions();
-    options =
-        options.scale_factor(std::vector<double>{static_cast<double>(scale_factor), static_cast<double>(scale_factor)});
+
+    Attr attr = op.get_attr("scale_factor");
```
mstojkovicTT, line 39: I am not a fan of `std::visit` approach because it is simply too complicated for logic it provides. you can try something like `std::holds_alternative` or `std::get_if` to determine what `std::variant` is holding.

forge/csrc/ops/op_upsample_2d.cpp
```diff
@@ -66,24 +85,52 @@ std::tuple<Shape, std::vector<DimBroadcast>> shape(
...
 
     const auto &input_shape = in_shapes[0];
-    TT_ASSERT(input_shape.size() >= 4, "Upsample2d input must have at least 4 dimensions");
+    TT_ASSERT(input_shape.size() == 4, "Upsample2d input must have at least 4 dimensions");
+
+    Attr attr = op.get_attr("scale_factor");
+
+    std::vector<uint32_t> scale_factor_vec;
+
+    std::visit(
```
mstojkovicTT, line 94: same comment for `std::visit` here

forge/csrc/test/ops/test_resize.cpp
```diff
@@ -63,18 +65,6 @@ bool valid_resize2d_inputs(
     return false;  // Mixed up/down scaling not supported
 }
 
-bool valid_upsample2d_inputs(const std::vector<graphlib::Shape>& shapes, bool channel_last)
```
mstojkovicTT, line 66: Can you double check that we support shapes where `H` and `W` are different?

forge/csrc/test/ops/test_resize.cpp
```diff
@@ -155,11 +145,11 @@ std::vector<tt::ops::Op> get_upsample2d_ops()
 {
     std::vector<tt::ops::Op> ops;
 
-    std::vector<int> scale_factors = {2, 3, 4};
+    std::vector<std::vector<int>> scale_factors = {{2, 2}, {2, 3}, {8, 4}, {3, 9}};
```
mstojkovicTT, line 148: Also here, check if different scale factor for `H` and `W` is supported.

mstojkovicTT (changes requested):
Comment in general: please split your changes into logically grouped commits. Avoid submitting a single commit that includes all changes, especially when modifying over 500 lines of code.

forge/csrc/ops/op_resize_2d.cpp
```diff
@@ -119,14 +77,45 @@ std::tuple<Shape, std::vector<DimBroadcast>> shape(
 
     std::vector<uint32_t> output_shape = input_shape;
 
+    // Determine whether to use upsample or downsample
+    bool upsample_h = channel_last ? sizes[0] >= static_cast<int>(input_shape[input_shape.size() - 3])
+                                   : sizes[0] >= static_cast<int>(input_shape[input_shape.size() - 2]);
+    bool upsample_w = channel_last ? sizes[1] >= static_cast<int>(input_shape[input_shape.size() - 2])
+                                   : sizes[1] >= static_cast<int>(input_shape[input_shape.size() - 1]);
+
+    if ((upsample_h and (!upsample_w)) or ((!upsample_h) and upsample_w))
```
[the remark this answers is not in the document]
chandrasekaranpradeep: Updated to use `&& `and `||` instead `and` and `or`

forge/csrc/ops/op_resize_2d.cpp
```diff
@@ -119,14 +77,45 @@ std::tuple<Shape, std::vector<DimBroadcast>> shape(
 
     std::vector<uint32_t> output_shape = input_shape;
 
+    // Determine whether to use upsample or downsample
+    bool upsample_h = channel_last ? sizes[0] >= static_cast<int>(input_shape[input_shape.size() - 3])
```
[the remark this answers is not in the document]
chandrasekaranpradeep: Updated to use size_h and size_w

forge/csrc/ops/op_resize_2d.cpp
chandrasekaranpradeep: Updated

forge/csrc/ops/op_upsample_2d.cpp
chandrasekaranpradeep: Updated to use `std::get_if `instead of `std::visit` for scalar factor

forge/csrc/ops/op_upsample_2d.cpp
chandrasekaranpradeep: Updated to use `std::get_if`

forge/csrc/test/ops/test_resize.cpp
chandrasekaranpradeep: Upsample2d supports different H and W and also added the gtest and pytest

forge/csrc/test/ops/test_resize.cpp
chandrasekaranpradeep: Yeah it support different scalar factor. Added gtest and pytest

mstojkovicTT (approved):
Looks good :))

Issue #1357, opened 2025-03-03
Modify model analysis to support backward pass analysis


ndrakulicTT:
Idea is to enable model analysis to generate single ops test with training enabled in compile and in the verify function.
 
Blocked by #1356

PR #2805, opened 2025-08-12
Add test generation for backwards pass

mmilosevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-forge-fe/issues/1357

### Problem description
TT-Forge-FE currently has infrastructure for extracting ops from all models running in tests. These tests are only run for forward (inference). We want to add support for testing both forward and backward passes in training mode.

### What's changed
- Addition of test creation for training on single ops.
- Changes to verify_backwards to support tagging of stages.

### Checklist


forge/forge/forge_property_utils.py
```diff
@@ -278,7 +278,10 @@ class ExecutionStage(Enum):
     FAILED_FORGE_MLIR_COMPILATION = auto()
     FAILED_TTNN_BINARY_EXECUTION = auto()
```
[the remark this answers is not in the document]
mmilosevicTT: I can, I was just wandering if it is safe to do this. If we change naming, we need to update all dashboards that use these errors as filters, no?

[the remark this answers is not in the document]
chandrasekaranpradeep: Thats sounds good !
In sum at the end of the forge compilation we are going to record the FAILED_TTNN_BINARY_EXECUTION and at the start of the verify function, we are going to record FAILED_TTNN_BINARY_FORWARD_EXECUTION and at the start of the verify_backward function, we are going to record FAILED_TTNN_BINARY_BACKWARD_EXECUTION

Am I right?

In additions, I have checked the Superset Model analysis and FFE models dashboard, we are not using the Execution Stage as filter we will mostly use BringUp status so updating the names is not issues. The ExecutionStage is recorded to give more granular overview of the compilation stages.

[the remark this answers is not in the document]
chandrasekaranpradeep: @nvukobratTT  Let me know your thoughts here too

forge/forge/python_codegen.py
```diff
@@ -909,49 +913,55 @@ def write_pytest_function(
         """
         self.wl("")
         self.wl("")
-        if use_ids_function:
```
mmilosevicTT, line 912: @chandrasekaranpradeep @nvukobratTT currently I generate two tests in a single file, should we maybe generate 2 separate files, for example one being test_add and other test_add_backward. In this way we can have separation between tests and no pollution of the code with this many ifs.

forge/forge/verify/verify.py
```diff
@@ -365,17 +370,23 @@ def verify_backward(
         co_gradients[name] = grad.to_torch().clone() if name.startswith("grad_acc_") else grad.to_torch()
 
     # Run backward on framework model
-    framework_model.zero_grad()
+    # This test is currently used only for torch.nn.Module and model_ops tests which use ForgeModule's (which do not require zero_grad)
```
nvukobratTT, line 373: Let's expand the explination with a reason why we're resetting gradients during verification as well.

forge/forge/verify/verify.py
```diff
@@ -365,17 +370,23 @@ def verify_backward(
...
         co_grad = co_gradients[name]
 
         if name.startswith("grad_acc_"):
             name = name.replace("grad_acc_", "")
             name = name.replace("_grad_accumulator", "")
-            fw_grad = framework_model.get_parameter(name).grad
+            if isinstance(framework_model, torch.nn.Module):
+                fw_grad = framework_model.get_parameter(name).grad
+            else:
+                fw_grad = framework_model.get_parameter(name).value().grad
```
nvukobratTT, line 389: If we run this with other models, it'll cause conflicts. Let's add a specific check for Forge Modules and error out in other cases.

forge/forge/forge_property_utils.py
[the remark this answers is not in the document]
nvukobratTT: Why do we need to update the names and add backward-specific stages? E.g., `FAILED_FORWARD_VERIFICATION ` and `FAILED_BACKWARD_VERIFICATION ` represent the same information: "Failed during verification". Therefore, for this and other cases, why do we need to separate these cases?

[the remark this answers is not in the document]
nvukobratTT: Have in mind that we can modify existing SQL queries to differentiate between inference and training tests based on names/parameters. We don't need to add more granular logic for parameter recording in order to bypass this.

forge/forge/python_codegen.py
nvukobratTT: How about just generating a single file and a single test with 2 test cases? One for inference, and one for training (e.g., pytest parameter with train flag). Splitting further will generate x2 load of already existing code, with most of the code duplicated between the two.

nvukobratTT (commented):
Left a few general comments. Let's discuss them and do a re-review afterwards :))

forge/forge/forge_property_utils.py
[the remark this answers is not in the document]
chandrasekaranpradeep: Let say for training tests if the tests is failing in verification either in verify function(i.e forward pass) or verify_backward function(i.e backward pass), by recording FAILED_VERIFICATION we will be not sure whether the issues in forward/backward program but by recording this execution stage which help to let us know where is the exact issues.

forge/forge/verify/verify.py
```diff
@@ -344,17 +344,22 @@ def verify_backward(
     if not isinstance(compiled_output, torch.Tensor):
         raise TypeError(f"Compiled output tensor must be of type {torch.Tensor}, but got {type(compiled_output)}")
 
-    if not isinstance(framework_model, torch.nn.Module):
-        raise TypeError(f"Framework model must be of type {torch.nn.Module}, but got {type(framework_model)}")
+    if not isinstance(framework_model, verify_cfg.framework_model_types):
```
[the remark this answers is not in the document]
mmilosevicTT: Yeah, op tests are using ForgeModels, so we need to pass them through. We can switch this to only support ForgeModels and torch.nn.Module not to make untested paths go through.

forge/forge/verify/verify.py
mmilosevicTT: Done.

forge/forge/forge_property_utils.py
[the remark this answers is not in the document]
nvukobratTT: That is not entirely correct. Currently, `verify_backward` only does verification of gradients. Therefore, even if we just record `FAILED_VERIFICATION`, we can differentiate the two by the mode in which the model is run. E.g., inference verifies outputs, training verifies gradients. They are not mixed in the current implementation.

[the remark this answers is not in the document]
nvukobratTT: Therefore, I don't see a value in having these as duplicates

forge/forge/python_codegen.py
```diff
@@ -961,7 +961,8 @@ def write_pytest_function(
             )
         else:
             self.wl('@pytest.mark.parametrize("forge_module_and_shapes_dtypes", forge_modules_and_shapes_dtypes_list)')
-        self.wl("def test_module(forge_module_and_shapes_dtypes):")
+        self.wl('@pytest.mark.parametrize("enable_training", [False, True], ids=["inference", "training"])')
```
nvukobratTT, line 964: Let's come up with a better name. `enable_training` seems misleading. Maybe to go with an option: "verify_training", "training_mode", etc. Something that is a bit more specific.

forge/forge/forge_property_utils.py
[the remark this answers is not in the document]
mmilosevicTT: I am not entirely sure I understand your point. We need to run both forward and backward in order to verify training. Forward layers might have different behaviours in training and inference mode. Because of this we will first run `verify` than `verify_backwards` for the training test. I agree that we would know which mode we failed in, but it might be useful for training to know which part failed, forward or backward pass. 

I might be missing something on the logging side tho. Is there something specific we can do to get this info in the case of training?

forge/forge/python_codegen.py
mmilosevicTT: Maybe just "test_mode"? It seems more standard way of calling.

forge/forge/verify/verify.py
```diff
@@ -518,7 +533,23 @@ def verify(
...
+        verify_backward(
+            inputs,
+            grad,
+            fw_out[0],
+            co_out[0],
+            framework_model,
+            compiled_model,
+            verify_cfg=verify_cfg,
+        )
+    else:
```
chandrasekaranpradeep, line 549: Can we record the ExecutionStage.PASSED at the end of the verify function by removing this else condition and it also help to avoid recording PASSED at the end of the verify_backward function?

forge/forge/verify/verify.py
```diff
@@ -518,7 +533,23 @@ def verify(
         if verify_cfg.verify_values:
             verify_cfg.value_checker.check(fw, co)
 
-    record_execution(ExecutionStage.PASSED)
+    # This will only work if model is compiled in training mode
+    if with_backward:
+        record_execution(ExecutionStage.FAILED_TTNN_BINARY_EXECUTION)
```
chandrasekaranpradeep, line 538: May be we record FAILED_TTNN_BINARY_EXECUTION at the start of the verify_backward instead of recording it here and before running compiled model backward program?

chandrasekaranpradeep (changes requested):

forge/forge/verify/verify.py
mmilosevicTT: I just wanted to make sure we do not enter verify_backward with the previous phase FAILED_VERIFICATION as technically inputs preparation is more of a FAILED_TTNN_BINARY_EXECUTION problem. I can remove it.

chandrasekaranpradeep: Its better to remove it because we are recording it twice before entering verify_backward and before running backward program.

chandrasekaranpradeep (approved):
Changes looks good to me  :)
Thanks for pushing this @mmilosevicTT

nvukobratTT (approved):

PR #2830, opened 2025-08-18
Add support for topk op

meenakshiramanathan1:
### Ticket

- Fixes https://github.com/tenstorrent/tt-forge-fe/issues/2836

### Problem description

- Topk op cannot be directly supported in forge since it's returns values and inidices as outputs while forge only supports ops with single outputs.

### What's changed
This PR adds  support for the TopK operation (returning both values and indices) across the stack and also updates the graph builder and tensor utilities to better handle ops with multiple outputs. Core changes are mentioned below.

- compile.py: Ensures a single IR op node per source op name and reuses it across multiple consumers/outputs and while creating tensors to build the graph, caches created op nodes by tensor.src_op.name and reuses them instead of creating duplicates and keeps output naming unique and stable so both TopK outputs can attach to the same underlying op node.

- tvm_to_python.py: It Maps topk to forge and when consuming a TopK node’s indices (producer port 1), rewrites the input name to <topk_node_name>_indices and when a TopK node is a graph output, adds a secondary output name <nid>_indices so both values and indices are returned.

- python_codegen.py: Modifies forgewriter to emit proper LHS unpacking for TopK and generates values, values_indices.

- topk.py: Adds a Python-friendly layer and tuple-like wrapper for TopK. TensorPair wraps (values, indices) and behaves like a 2‑tuple, forwarding attributes to values and supporting set_src_layer for both. It computes golden outputs with torch.topk when input has values, otherwise creates correctly-shaped zero tensors and it creates Forge tensors from trace and sets their values and data formats to returns a TensorPair.

- tensor.py: Extends conversion utilities to accept tuple-like wrappers (e.g., TensorPair) when converting to Forge tensors and detects iterable/sequence-like wrappers and flattens them into the corresponding list of Tensors.

The generated graph now unpacks two values from topk.
Screenshot 2025-08-19 at 10 50 54 AM
Screenshot 2025-08-19 at 1 55 08 PM

- Topk op sanity passes tvm verification successfully and further fails in run_mlir_compiler_stage with unsupported op topk.

### Logs
[topk_pcc_correction.log](https://github.com/user-attachments/files/21848659/topk_pcc_correction.log)


nvukobratTT:
Inactive, closing this PR.

PR #2837, opened 2025-08-19
Fix tensor layouts in runtime

pglusacTT:
### Ticket
N/A

### Problem description
We were incorrectly indexing tensor layouts, which sometimes causes our ttnn tensors to have wrong dtypes.
This caused workloads to fail in Debug build, while they worked in Release.


pmarkovicTT (approved):

pilkicTT (approved):
thanks for the fix!

PR #2843, opened 2025-08-19
Deleted pixel shuffle op.

acolicTT:
Deleted pixel shuffle op, because it is not used anymore.


acolicTT:
Closing this PR, since we will migrate this op to cpp.

PR #2876, opened 2025-08-22
Add new tags for more granular training mode logging

mmilosevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-forge-fe/issues/2865

### Problem description
We want to have a way of logging tags in a more granular way for training mode.

### What's changed
Addition of run_mode and is_forward tags to our logs.

### Checklist


forge/forge/compile.py
```diff
@@ -881,7 +911,9 @@ def run_post_autograd_pass(context: CompileContext) -> CompileDepth:
         context.losses,
         context.outputs,
     )
-    record_execution(ExecutionStage.FAILED_FORGE_POST_AUTOGRAD_DECOMP)
+    record_execution(
+        ExecutionStage.FAILED_FORGE_POST_AUTOGRAD_DECOMP, ExecutionRunMode.from_training_param(context.training)
```
ndrakulicTT, line 915: Do we need to have ExecutionPass here?

forge/forge/forge_property_utils.py
```diff
@@ -556,7 +574,7 @@ class ForgePropertyHandler:
 
     def __init__(self, store: Optional[ForgePropertyStore] = None):
         self.store = store if store is not None else ForgePropertyStore()
-        self.record_execution(ExecutionStage.FAILED_BEFORE_FORGE_COMPILATION_INITIATION)
+        self.record_execution(ExecutionStage.FAILED_BEFORE_FORGE_COMPILATION_INITIATION, None, None)
```
ndrakulicTT, line 577: parametars are optional so maybe we dont need None, None

forge/forge/forge_property_utils.py
```diff
@@ -676,15 +694,44 @@ def record_execution_stage(self, execution_stage: ExecutionStage):
...
+        Args:
+            execution_pass (ExecutionPass): The execution pass value.
+        """
+        self.add("tags.is_forward", True if execution_pass == ExecutionPass.FORWARD else False)
+
+    def record_execution(
+        self,
+        execution_stage: ExecutionStage,
+        execution_run_mode: Optional[ExecutionRunMode],
+        execution_pass: Optional[ExecutionPass],
```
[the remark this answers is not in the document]
mmilosevicTT: We pass None in 1 place, as this is required. IMO, it is not a good practice to put default value in all the nested functions. This is because you actually want to be aware of which function you are calling i.e. with what parameters. Putting defaults only in the highest level function is fine, but all the nested ones should stay with passed parameters. In some languages this is even blocked in the compiler.

forge/forge/compile.py
mmilosevicTT: I am not sure. This is run in both inference and training mode. So I was sceptical about logging backwards with inference mode.

forge/forge/compile.py
```diff
@@ -635,7 +641,11 @@ def generate_initial_graph(context: CompileContext) -> CompileDepth:
 
             modules_.append(module)
 
-    record_execution(ExecutionStage.FAILED_FORGE_INITIAL_GRAPH_PASS)
+    record_execution(
```
chandrasekaranpradeep, line 644: @mmilosevicTT Could you clarify the reason for recording ExecutionRunMode inside every Forge graph pass (e.g., generate_initial_graph, run_post_initial_graph_pass, etc.) and TVM function? Since the training member in the CompileContext dataclass is set to either True or False and doesn’t change afterward, wouldn’t it be sufficient to record ExecutionRunMode only once in the compile_main function?

forge/forge/compile.py
```diff
@@ -681,7 +691,11 @@ def run_post_initial_graph_pass(context: CompileContext) -> CompileDepth:
     graph_name = context.graph_name
     graph, intermediate_tensors = context.graph, context.intermediate_tensors
 
-    record_execution(ExecutionStage.FAILED_FORGE_POST_INITIAL_GRAPH_PASS)
+    record_execution(
```
chandrasekaranpradeep, line 694: Could you please share some insight into why we are recording the ExecutionPass? What is its purpose, and how does it differ from ExecutionRunMode?

chandrasekaranpradeep (changes requested):

forge/forge/compile.py
mmilosevicTT: So in training mode, we have two parts of execution: forward and backward pass. We want to have more granular way of knowing in which pass we failed. Since I am decoupling the tags from the other comment, I would say we might not even need to record ExecutionPass at all in compilation. If we fail with anything in compilation, it is self-explanatory which pass it was. Will update this.

chandrasekaranpradeep (approved):

forge/forge/compile.py
chandrasekaranpradeep: Synced offline

PR #2885, opened 2025-08-25
Handle nearest interpolation cases

chandrasekaranpradeep:


saiarthiraguram (approved):

mstojkovicTT (commented):
I have a question, have you tried to cover paths where there is no need to do float scale factor logic?

chandrasekaranpradeep:
> I have a question, have you tried to cover paths where there is no need to do float scale factor logic?

Yes, I’ve also covered the integer scale factor cases. For integer scaling, the resize2d op is decomposed into a ttnn.upsample op. For floating scale factors, downsampling, or mixed sampling cases, the resize2d op is decomposed into repeat_interleave or conv2d, since ttnn.upsample does not support floating scale factors or downsampling.

mstojkovicTT (approved):

nvukobratTT (approved):

PR #2923, opened 2025-09-02
Support test-expression input in test workflow

chandrasekaranpradeep:
### Summary
Add a new workflow input `test-expression` that lets developers run a subset of pytest tests using a substring expression. This exposes pytest’s -k matching as an easy-to-use workflow parameter to quickly target tests by name.

### Motivation
We're enabling training support in the model-analysis pipeline, and in those scenarios we sometimes need to run either inference or training test subsets. test-expression makes it possible to select the appropriate subset

### DIFF
--- a/.github/workflows/test-sub.yml
+++ b/.github/workflows/test-sub.yml
@@ -35,6 +35,10 @@ on:
         description: 'Filters for tests (comma separated)'
         required: false
         type: string
+      test-expression:
+        description: 'Run tests which match the given substring expression'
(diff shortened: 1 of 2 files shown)


nsumrakTT (approved):

mmilosevicTT:
Question, so by default, are we going to run all tests? We need to be careful not to break existing pipelines with newly added training tests in the future.

nvukobratTT (approved):

PR #2954, opened 2025-09-18
Update rev of unimport

gfengTT:
A non-existent rev was causing pre-commit to fail.

### Problem description
See https://github.com/tenstorrent/tt-forge-fe/actions/runs/17833610925/job/50704865494)
error: pathspec '1.2.1' did not match any file(s) known to git


gfengTT:
Already fixed

PR #3098, opened 2025-12-12
Add scheduled trigger to on-nightly workflow

chandrasekaranpradeep:
## Summary

Add a scheduled trigger to automatically run the nightly workflow daily at 00:00 UTC. The workflow will run automatically on a daily schedule to check onnx and paddlepaddle model failures.


nsumrakTT (changes requested):
Why do we need this?

PR #3103, opened 2025-12-18
[nightly] use nightly marker

pilkicTT:
To cover all of the wanted tests in the nightly run - use the `nightly` marker.

Additionally:

- filter out all tests which are not ONNX, paddlepaddle or tensorflow.
- for nightly only build release
- run everything on n150
- skip one flaky test

### DIFF
--- a/.github/workflows/on-nightly.yml
+++ b/.github/workflows/on-nightly.yml
@@ -24,6 +24,7 @@ jobs:
     secrets: inherit
     with:
       docker-image: ${{ needs.docker-build.outputs.docker-image }}
+      build: 'Release'
 
   test_forge_model:
     needs:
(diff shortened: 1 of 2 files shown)


chandrasekaranpradeep:
@pilkicTT  @nsumrakTT 
1. There are some model parts tests present in [forge/test/models/test_model_parts.py](https://github.com/tenstorrent/tt-forge-fe/blob/main/forge/test/models/test_model_parts.py) where the tests are marked with the `nightly marker.` If, in the nightly workflow, we are only going to test models, Can I mark these tests with the `push` marker and include them in the `ON-PR` workflow, since most of these tests are ONNX framework tests.
2. Do we need to build in both `release` and `debug` modes in the ON-NIGHTLY workflow? The ON-PR workflow builds in both modes, but here we are testing the models on the main branch. Please let me know your thoughts on building only in `release` mode.

PR #3122, opened 2025-12-31
Enable Verbose Output for Model Tests in ON-PR and ON-NIGHTLY Workflows

chandrasekaranpradeep:
## Summary

This PR adds a `enable-verbose-output` input parameter to the `test-sub.yml` workflow and enables it for model testing jobs in both ON-PR and ON-NIGHTLY workflows. This will provide detailed verbose output (`-svv` flag) during test execution to help diagnose flaky test failures and crashes.

## Changes Made

1. **Added `enable-verbose-output` input to `test-sub.yml` workflow**
   - New boolean input parameter (default: `false`)
   - When enabled, adds `-svv` flag to pytest command for verbose output

2. **Enabled verbose output in ON-PR and ON-NIGHTLY workflow**
   - Added `enable-verbose-output: true` to the  `test-models`  and`test_forge_model` job

## Reason for Adding This Feature

We are experiencing crashed cases in both ON-PR and ON-NIGHTLY workflow pipelines when testing ONNX and PaddlePaddle models. Currently, we face the following challenges:

1. **Unclear failure stage**: We cannot determine at which stage of the Forge compilation the crash is occurring
2. **Root cause identification**: We cannot determine if the same failure causes crashes across different test runs

This will help us track and resolve the flaky test issues more effectively.

### DIFF
--- a/.github/workflows/test-sub.yml
+++ b/.github/workflows/test-sub.yml
@@ -241,6 +241,7 @@ jobs:
           path: collected_test_list
 
       - name: Run Test
+        id: run-test
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
           HF_HOME: /mnt/dockercache/huggingface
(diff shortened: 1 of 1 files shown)


.github/workflows/test-sub.yml
```diff
@@ -251,7 +256,8 @@ jobs:
           python .github/workflows/test_runner.py  \
                --junit-xml=${{ steps.strings.outputs.test_report_path }} \
                --log-memory-usage \
-               $(if [ "${{inputs.continue-on-crash}}" = "true" ]; then echo "--continue-after-crash"; fi)
+               $(if [ "${{inputs.continue-on-crash}}" = "true" ]; then echo "--continue-after-crash"; fi) \
+               $(if [ "${{inputs.enable-verbose-output}}" = "true" ]; then echo "-svv"; fi)
```
nvukobratTT: Consider storing the log in a file and retaining it as a workflow artifact. In sum, GitHub has limitations on the size of output it can display. 

Note: I'm not sure if this limitation is still present, but please double-check.

chandrasekaranpradeep: Thanks for catching this :) 
Thats right it has limitation, so will store it in separate file and retain it as artifacts

nsumrakTT (approved):

.github/workflows/test-sub.yml
```diff
@@ -263,10 +278,10 @@ jobs:
 
       - name: Upload Test Log
         uses: actions/upload-artifact@v4
-        if: success() || failure()
+        if: always()
```
nsumrakTT: please use `success() || failure()` instead of always() because always() doesn't handle canceling of the job.

.github/workflows/test-sub.yml
chandrasekaranpradeep: @nvukobratTT  Updated the workflow to capture and persist test runner logs as workflow artifacts. Additionally, added comprehensive logging for TVM compilation stages (frontend conversion, Relay passes, Forge passes, and partition) and Forge module generation to improve visibility into the compilation pipeline and aid in debugging flaky test failures.
In the below log file, you can find the difference between before and after of log info updation,
[mnist_after_log_info_updation.log](https://github.com/user-attachments/files/24393650/mnist_after_log_info_updation.log)
[mnist_before_log_info_updation.log](https://github.com/user-attachments/files/24393653/mnist_before_log_info_updation.log)

.github/workflows/test-sub.yml
```diff
@@ -248,11 +253,21 @@ jobs:
         shell: bash
         run: |
           source env/activate
-          python .github/workflows/test_runner.py  \
-               --junit-xml=${{ steps.strings.outputs.test_report_path }} \
-               --log-memory-usage \
-               $(if [ "${{inputs.continue-on-crash}}" = "true" ]; then echo "--continue-after-crash"; fi)
-          exit_code=$?
+          if [ "${{inputs.enable-verbose-output}}" = "true" ]; then
```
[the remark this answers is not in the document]
chandrasekaranpradeep: Updated

PR #3129, opened 2026-01-08
Add support for external tt-mlir builds via TTMLIR_SOURCE_DIR

anuragsingh-tt:
Enable tt-forge-fe to consume a pre-built external tt-mlir instead of requiring the tt-mlir submodule.
- Allow overriding tt-mlir location via TTMLIR_SOURCE_DIR (CMake, env, or cache).
- Update include/lib paths to derive from the resolved tt-mlir root.
- Add CMake logic to validate and auto-install required tt-mlir components when using an external build.
- Detect external tt-mlir in Python runtime to correctly set TT_METAL_RUNTIME_ROOT and source/wheel mode.
- Resolve tt-mlir and tt-metal commit SHAs from external sources in setup.py.

Default behavior (building tt-mlir from the submodule) remains unchanged.

### DIFF
--- a/docs/src/getting_started_build_from_source.md
+++ b/docs/src/getting_started_build_from_source.md
@@ -195,6 +195,12 @@ cmake --build env/build
 >
 > **NOTE:** Special care should be taken to ensure that the already built TT-MLIR environment (toolchain) version is compatible with the one TT-Forge-FE is using.
 
+> **Using a Pre-built TT-MLIR:** If you are developing on TT-MLIR and want to use your own build instead of the submodule, you can specify the path using the `TTMLIR_SOURCE_DIR` option when configuring the main build (step 8):
+> ```bash
+> cmake -G Ninja -B build -DTTMLIR_SOURCE_DIR=/path/to/tt-mlir -DCMAKE_CXX_COMPILER=clang++-17 -DCMAKE_C_COMPILER=clang-17
+> ```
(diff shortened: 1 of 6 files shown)


chandrasekaranpradeep (changes requested):
1. In `mlir_compiler.cpp` ,  the hardcoded [third_party/tt-mlir](https://github.com/tenstorrent/tt-forge-fe/blob/68adc1743b6128536effe11b0f311d0f104ae048/forge/csrc/passes/mlir_compiler.cpp#L176) path breaks external tt-mlir builds. May be we can use `TTMLIR_SOURCE_DIR` environment variable (set in `__init__.py`) to construct `standalone_dir` dynamically with fallback to default submodule path.
 
2. In `setup.py`, the [CMake cache reading logic ](https://github.com/tenstorrent/tt-forge-fe/blob/aee344dca6e20cf3b54055e26c82ab77e915de3e/setup.py#L259) is duplicated in `get_tt_mlir_commit_hash()` and `get_tt_metal_commit_hash()`. Can we create a unified helper function like `get_ttmlir_source_dir()` to eliminate duplication?
 
3. Can you add the steps for using pre-built tt-mlir in the forge docs?

chandrasekaranpradeep (approved):
@anuragsingh-tt  Thanks for addressing the comments and I have tried building the tt-forge-fe using pre-built tt-mlir and validated by running model.

nvukobratTT (approved):

pilkicTT (changes requested):
I don't think this is a good idea, it is introducing lot of complexity for unclear gains. Let's discuss this...

nvukobratTT:
Closing stale PRs

PR #3135, opened 2026-01-12
Fix Flaky YOLOv5 ONNX Test with 3-Tier Fallback Strategy

chandrasekaranpradeep:
## Summary

This PR addresses flakiness in the YOLOv5 ONNX model test (`test_yolov5_320x320`) by implementing a robust 3-tier fallback strategy for model loading. The test was previously failing intermittently in on-PR pipelines while downloading YOLOv5 PyTorch weight files from ultralytics, but would pass on retry. The new implementation prioritizes loading pre-converted ONNX models from S3, significantly improving test reliability.

### Failure
`
YOLOv5 loading error: Failed to download the model after multiple retries.
`

## Problem Statement

The YOLOv5 ONNX test was experiencing flakiness with the following symptoms:
- **Intermittent failures** in on-PR pipeline runs
- **Passing on retry** - tests would succeed when manually retriggered
- **Root cause**: Download failures of YOLOv5 `.pt` files from GitHub releases after multiple retry attempts

The original implementation relied solely on downloading PyTorch weights and loading the model and converting them to ONNX format, which was prone to network-related failures.

## Solution: 3-Tier Fallback Strategy

Implemented a robust fallback mechanism that attempts model loading in the following order:

### Tier 1: Direct ONNX Load from S3 (Primary)
- **Path**: `test_files/onnx/yolov5/yolov5{size}_320x320.onnx`
- **Method**: Directly loads pre-converted ONNX models from S3 bucket using `get_file()`

### Tier 2: PyTorch Weights from S3 + Conversion (Secondary)
- **Path**: `test_files/pytorch/yolov5/yolov5{size}.pt`
- **Method**: Downloads PyTorch weights from S3, loads model, and converts to ONNX

### Tier 3: Original Method (Fallback)
- **Method**: Falls back to original implementation using `torch.hub.load()` from GitHub

### Implementation Details

**Key Functions:**
- `_load_onnx_from_s3()`: Attempts direct ONNX load from S3
- `_load_pytorch_from_s3_and_convert()`: Loads PyTorch weights from S3 and converts to ONNX
- `_load_via_torch_hub()`: Original fallback method using torch.hub
- `load_model_and_inputs()`: Orchestrates the 3-tier fallback strategy

### Problem

While testing the YOLOv5 model with the 3-tier fallback strategy, Tier 1 (ONNX load from S3) and Tier 2 (PyTorch weights from S3) were failing on non-shared runners with the following error:

```
--2026-01-13 08:29:36--  http://large-file-cache.large-file-cache.svc.cluster.local/test_files/onnx/yolov5/yolov5n_320x320.onnx
Resolving large-file-cache.large-file-cache.svc.cluster.local (large-file-cache.large-file-cache.svc.cluster.local)... failed: Temporary failure in name resolution.
wget: unable to resolve host address 'large-file-cache.large-file-cache.svc.cluster.local'
Failed to load ONNX from S3 (test_files/onnx/yolov5/yolov5n_320x320.onnx): wget failed with exit code 1024 when downloading http://large-file-cache.large-file-cache.svc.cluster.local/test_files/onnx/yolov5/yolov5n_320x320.onnx
Failed to load PyTorch weights from S3 and convert (test_files/pytorch/yolov5/yolov5n.pt): wget failed with exit code 1024 when downloading http://large-file-cache.large-file-cache.svc.cluster.local/test_files/pytorch/yolov5/yolov5n.pt
Falling back to torch.hub.load method
```

**Root Cause:**
In the current Test Sub workflow, the `IRD_LF_CACHE` environment variable was set for all runners. When running the YOLOv5 model test on non-shared runners, the above-mentioned issues were raised, but on shared runners, the weights/ONNX files can be downloaded from the S3 bucket. While checking the tt-xla CI, found that they have set `IRD_LF_CACHE` for shared runners, which allows direct downloads via wget from the cache server, and `DOCKER_CACHE_ROOT` was set for non-shared runners, which points to a directory (`/mnt/dockercache`) that is automatically synced with the S3 bucket every hour - [reference](https://github.com/tenstorrent/tt-xla/blob/8bea3421b0b6d56c69e2fe3f787091cd08287374/.github/workflows/call-test.yml#L82). This is because the `get_file()` function in `third_party/tt_forge_models/tools/utils.py` accepts either `IRD_LF_CACHE` or `DOCKER_CACHE_ROOT` environment variables.

### Solution

Added similar environment variable configuration in `.github/workflows/test-sub.yml` (lines 183-191) to set the appropriate cache variable based on runner type:

- **Shared runners**: Set `IRD_LF_CACHE` from GitHub variables to enable direct downloads via wget from the cache server
- **Non-shared runners**: Set `DOCKER_CACHE_ROOT=/mnt/dockercache` to use the locally synced S3 cache directory

This fix ensures that `get_file()` can properly locate and load model files from S3 cache on both shared and non-shared runners. When `DOCKER_CACHE_ROOT` is set, `get_file()` looks for files in the specified directory (which is automatically synced with S3 every hour) instead of attempting to download via wget, allowing Tier 1 and Tier 2 of the fallback strategy to work correctly across all runner types.

So updated the similar environment variable configuration in other workflows and removed unwanted setting of DOCKER_CACHE_ROOT and IRD_LF_CACHE in on-pr, model-analysis and on-nightly workflows

### DIFF
--- a/.github/workflows/on-nightly.yml
+++ b/.github/workflows/on-nightly.yml
@@ -8,11 +8,6 @@ permissions:
   packages: write
   checks: write
 
-# Set cache root to avoid IRD_LF_CACHE error when loading CenterNet ONNX model in CI
-env:
-  DOCKER_CACHE_ROOT: /mnt/dockercache
-  IRD_LF_CACHE: ${{ vars.IRD_LF_CACHE }}
-
 jobs:
   docker-build:
     uses: ./.github/workflows/build-image.yml
--- a/.github/workflows/on-pr.yml
+++ b/.github/workflows/on-pr.yml
@@ -20,10 +20,6 @@ permissions:
   id-token: write
   actions: write
 
-# Set cache root to avoid IRD_LF_CACHE error when loading CenterNet ONNX model in CI
-env:
-  DOCKER_CACHE_ROOT: /mnt/dockercache
-  IRD_LF_CACHE: ${{ vars.IRD_LF_CACHE }}
 
 jobs:
   spdx:
--- a/.github/workflows/test-model-analysis-sub.yml
+++ b/.github/workflows/test-model-analysis-sub.yml
@@ -83,7 +83,6 @@ jobs:
     env:
       HF_TOKEN: ${{ secrets.HF_TOKEN }}
       HF_HOME: /mnt/dockercache/huggingface
-      IRD_LF_CACHE: ${{ vars.IRD_LF_CACHE }}
       FORGE_MODELS_CACHE: /mnt/dockercache/forge_models_cache
       HF_HUB_DISABLE_PROGRESS_BARS: 1
       FORGE_DISABLE_REPORTIFY_DUMP: 1
@@ -191,6 +190,16 @@ jobs:
             pip install "${forge_wheels[@]}" --force-reinstall || true
           else
             echo "No tt_forge_fe wheel found; continuing."
+          fi
+
+      - name: Set caching env variables
+        shell: bash
+        run: |
+          shared_runners=${{ inputs.sh-runner }}
+          if [ "$shared_runners" = "true" ]; then
+            echo "IRD_LF_CACHE=${{ vars.IRD_LF_CACHE }}" >> $GITHUB_ENV
+          else
+            echo "DOCKER_CACHE_ROOT=/mnt/dockercache" >> $GITHUB_ENV
           fi
 
       - name: Extract and Export Unique Ops Configuration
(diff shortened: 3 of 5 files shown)


chandrasekaranpradeep:
> This is fine for this flaky case. However, this will not be that good to scale for other ONNX models.
> 
> In sum, let's see whether we will be able to hose those on HF, which is expected to be much more stable. If that proves to be a good approach, we can apply it to most of our workloads, while keeping this 3-tier approach only when necessary.
Sure @nvukobratTT

PR #3158, opened 2026-01-25
removed reference to "mac_requirements.txt"

iharabukhouski:
### Ticket
na

### Problem description
"mac_requirements.txt" has been removed as part of this [pr](https://github.com/tenstorrent/tt-forge-onnx/pull/1544/files)

also, according to [this](https://github.com/tenstorrent/tt-forge-onnx/pull/1544#issuecomment-2884029861) comment macos is not considered officially supported

### What's changed
only removed reference to "mac_requirements.txt" in "create_venv.sh"

### Checklist
~- [ ] New/Existing tests provide coverage for changes~ - na

### DIFF
--- a/env/create_venv.sh
+++ b/env/create_venv.sh
@@ -22,19 +22,10 @@ if [[ -z "$CURRENT_SOURCE_DIR" ]]; then
     exit 1
 fi
 
-# Torch requires a specific version of wheel to be installed
-# which depends on the platform
-if [[ "$OSTYPE" == "darwin"* ]]; then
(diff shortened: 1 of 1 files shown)


iharabukhouski:
@nvukobratTT @chandrasekaranpradeep seems like some regression tests failed for this pr. it doesn't seem like this pr could have caused them. also i have noticed that some other prs also have similar regression test fails even when the pr itself seems to be fairly innocuous. can you please tell me if you know if the regression checks are known to be flaky sometimes?

PR #3258, opened 2026-04-10
Update ubuntu version and initial UV migration

### LINKED ISSUE #3171: Use ubuntu 24.04 as base docker image in tt-forge-onnx
Part of: https://github.com/tenstorrent/tt-xla/issues/3133

Update tt-forge-onnx Docker configuration to use Ubuntu 24.04 as the base image.

chandrasekaranpradeep:
Fixes #3171

## Summary

This PR migrates the entire repository from Ubuntu 22.04 to Ubuntu 24.04, replaces
`pip` with `uv` as the Python package manager throughout the codebase, optimizes the
Docker build pipeline, and fixes several CI reliability issues uncovered during the
migration.

### 1. Ubuntu 22.04 → 24.04 Migration

All Docker images and CI runners now target Ubuntu 24.04 LTS.

- Base, CI, IRD, and release Docker images updated from `ubuntu:22.04` to `ubuntu:24.04`
- All GHCR image tags updated: `*-ubuntu-22-04` → `*-ubuntu-24-04`
- CI runners updated: `tt-ubuntu-2204-large-stable` → `tt-ubuntu-2404-large-stable`
- Build dependency updated: `libstdc++-12-dev` → `libstdc++-14-dev` (Ubuntu 24.04 default toolchain)
- Documentation updated from Ubuntu 22.04 to Ubuntu 24.04 instructions

### 2. `pip` → `uv` Migration

`uv` replaces `pip` as the Python package manager everywhere in the codebase.

**Issues encountered:**
- **`uv: command not found` (exit 127):** The original approach tried to install `uv` by
  downloading `astral.sh/uv/install.sh` via `curl`. This silently failed with a `403`
  because `astral.sh` is blocked by the corporate proxy in our CI environment.
  The failure was invisible because `set -e` alone does not catch failures inside a
  `curl | sh` pipeline — only the last command's exit code (`sh`) matters, and `sh`
  exits `0` on empty input.
- **Fix — `set -eo pipefail`:** Added to `docker_install.sh` so pipeline failures are
  surfaced immediately instead of being swallowed.
- **Fix — `COPY --from=ghcr.io/astral-sh/uv:latest`:** Instead of downloading via curl,
  the `uv` binary is now pulled directly from the official GHCR image at build time.
  Docker image pulls use the Docker daemon's registry transport, which is not subject to
  the same HTTP proxy restrictions that blocked `astral.sh`.
- **`--break-system-packages` on Ubuntu 24.04 (PEP 668):** Ubuntu 24.04 marks the system
  Python at `/usr` as "externally managed" by Debian and refuses direct installs. Adding
  `--break-system-packages` to `uv pip install` in `Dockerfile.base` was required to
  install `cmake` and `pytest` into the system Python inside the container.
- **`UV_INDEX_STRATEGY=unsafe-best-match`:** Set in `env/activate` to allow `uv` to
  correctly resolve packages when multiple indexes are in use (PyPI + PyTorch CPU
  find-links). Without this, `uv` fails to pick the best match across indexes.

**Scope of changes:**
- `env/create_venv.sh`: replaced `pip install` with `uv pip install`, removed
  `pip install --upgrade pip` bootstrap step
- `forge/forge/CMakeLists.txt`, `CMakeLists.txt`: `pip install` → `uv pip install`
- `forge/forge/tvm_calls/forge_compile.py`: `pip show tvm` → `uv pip show tvm`
- `forge/test/conftest.py`: `pip freeze`, `pip install` → `uv pip freeze`, `uv pip install`
- `forge/test/utils.py`: subprocess-based `pip install` → `uv pip install`
- `scripts/model_analysis/utils.py`: `pip install pre-commit` → `uv pip install pre-commit`
- `pyproject.toml`: removed `pip` from `build-system.requires`
- All GitHub Actions workflows: `pip install` → `uv pip install`

### 3. GDB: From-Source Compilation → `apt install`

**Before:** GDB 14.2 was compiled from source in `Dockerfile.ird` — a 5–10 minute
compile step downloading a tarball from a mirror, running `./configure`, `make -j`,
`make install`.

**After:** `apt-get install gdb` — Ubuntu 24.04 ships GDB 15.x natively, which is
newer than what was being compiled. Build time for the IRD image drops by ~10 minutes.

### 4. Docker Build System Optimizations

- **`docker_install.sh` (new shared script):** Common system package installation
  extracted into a single script, mirroring the pattern used in `tt-xla`. Changes to
  this script now automatically bust the Docker image cache via `get-docker-tag.sh`.
- **`--no-install-recommends` + `apt-get clean`:** Applied to all `apt-get install`
  calls across all Dockerfiles to reduce layer size.
- **Layer consolidation:** Multiple `RUN` blocks merged across `Dockerfile.ird`,
  `Dockerfile.ci`, and `Dockerfile.release-ubuntu` to reduce the number of image layers.
- **`ADD` → `COPY` in `Dockerfile.ci`:** `COPY` is preferred for local paths; `ADD`
  has implicit tar-extraction and URL-fetch semantics that are not needed here.
- **`ln -sf` instead of `ln -s` for Clang symlinks:** Makes symlink creation idempotent
  — safe to re-run on layer cache misses.
- **`FileCheck-17` symlink added:** Required for LLVM tool testing.
- **`DOCKER_BUILDKIT=1`:** Enabled in `build-docker-images.sh` for parallel multi-stage
  builds and better layer caching.
- **Toolchain directory `mkdir` consolidated:** `Dockerfile.ird` merged two separate
  `mkdir`/`chmod` blocks into one `RUN` layer.

### 5. CI Reliability Fixes

- **`set pipefail` → `set -o pipefail` in `build-image.yml`:** `set pipefail` (without
  `-o`) is a no-op — it creates a shell variable, not a shell option. The pipe
  `build-docker-images.sh | tee docker.log` would silently pass even if the build
  script crashed, because `tee` exits `0`. Fixed to `set -o pipefail`.
- **pre-commit workflow:** Migrated from `actions/setup-python` + pip cache to
  `astral-sh/setup-uv` + `uvx pre-commit run --all-files`, consistent with the `uv`
  migration.
- **`Dockerfile.release-ubuntu`:** Also fixed to use `COPY --from=ghcr.io/astral-sh/uv:latest`
  for `uv` installation, since `astral.sh` is blocked in this CI environment.

### 6. Documentation Updates

- `getting_started_build_from_source.md`: Updated OS version to Ubuntu 24.04, added
  `uv` installation section, fixed `update-alternatives` command syntax, changed
  `ln -s` to `ln -sf`, added `FileCheck-17` symlink step, updated `libstdc++` version.
- `getting_started.md`: Updated wheel install commands from `pip` to `uv`.
- `tools.md`: Updated pre-commit installation from `pip install` to `uv tool install`.
- `docs/src/SUMMARY.md`: Removed the `# CI / Docker` section (internal tooling, not
  user-facing documentation).


vmilosevic (approved):

nsumrakTT (approved):

.github/workflows/build.yml
```diff
@@ -260,10 +260,10 @@ jobs:
 
       - name: Prepare code coverage report
         run: |
-          lcov --directory build --capture --output-file coverage.info --gcov-tool ${{ steps.strings.outputs.work-dir }}/.github/workflows/gcov_for_clang.sh
-          lcov --extract coverage.info '**/tt-forge-onnx/forge/csrc/*' --output-file coverage.info
+          lcov --directory build --capture --output-file coverage.info --gcov-tool ${{ steps.strings.outputs.work-dir }}/.github/workflows/gcov_for_clang.sh --ignore-errors inconsistent
+          lcov --extract coverage.info '**/tt-forge-onnx/forge/csrc/*' --output-file coverage.info --ignore-errors inconsistent
```
nvukobratTT, line 264: Why are we forced to introduce `--ignore-errors inconsistent`

env/activate
```diff
@@ -6,6 +6,10 @@
 # For tt-mlir we have similar env variables but since we build toolchain and binaries of tt-mlir
 # using install.sh from tt-mlir project, we only define them in tt-forge project top level cmake file.
 
+# Allow uv to select the best package version across multiple indexes (PyPI + PyTorch CPU find-links).
+# Without this, uv may fail to resolve packages correctly when extra indexes are in use.
+export UV_INDEX_STRATEGY=unsafe-best-match
```
nvukobratTT, line 11: Is this fine? @vmilosevic are we using this strategy for XLA as well?

nvukobratTT (approved):
Nice one! Leaving two small comments

.github/workflows/build.yml
chandrasekaranpradeep: The `--ignore-errors inconsistent` flag is required because Ubuntu 24.04 ships **lcov 2.x**, which treats coverage data inconsistencies as hard errors by default — whereas Ubuntu 22.04's lcov 1.14 silently ignored them. These inconsistencies are not a bug in our code, they are an expected artefact of Clang's gcov-compatible output format, where template instantiations and inline functions compiled across multiple translation units produce overlapping coverage records with differing hit counts. `--ignore-errors inconsistent` simply restores the lenient behaviour lcov 1.x had by default

env/activate
chandrasekaranpradeep: tt-xla use the same strategy - https://github.com/tenstorrent/tt-xla/blob/c62a0d95b59350a3ed628caa5384d2804d73ee24/venv/activate#L15

PR #3278, opened 2026-04-22
Replace scikit-image with pillow

nsumrakTT:
### Problem description
With ubuntu 24.04 and python 3.12 scikit-image requirement needs build tools to build scikit while pip install

### What's changed
Remove scikit-image with pillow

### Checklist

### DIFF
--- a/forge/test/models/onnx/vision/ssd300_resnet50/model_utils/image_utils.py
+++ b/forge/test/models/onnx/vision/ssd300_resnet50/model_utils/image_utils.py
@@ -2,13 +2,14 @@
 
 # SPDX-License-Identifier: Apache-2.0
 import numpy as np
-import skimage
+from PIL import Image
 
 
 def load_image(image_path):
     """Code from Loading_Pretrained_Models.ipynb - a Caffe2 tutorial"""
     mean, std = 128, 128
-    img = skimage.img_as_float(skimage.io.imread(image_path))
+    img = Image.open(image_path)
+    img = np.array(img).astype(np.float32) / 255.0  # Convert to float [0, 1]
     if len(img.shape) == 2:
         img = np.array([img, img, img]).swapaxes(0, 2)
     return img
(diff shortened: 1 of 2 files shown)


forge/test/models/onnx/vision/ssd300_resnet50/model_utils/image_utils.py
```diff
@@ -2,13 +2,14 @@
...
 import numpy as np
-import skimage
+from PIL import Image
 
 
 def load_image(image_path):
     """Code from Loading_Pretrained_Models.ipynb - a Caffe2 tutorial"""
     mean, std = 128, 128
-    img = skimage.img_as_float(skimage.io.imread(image_path))
+    img = Image.open(image_path)
```
chandrasekaranpradeep: `Image.open()` preserves the source colour mode. If the image is a palette PNG (`mode='P'`), RGBA (`mode='RGBA'`), or greyscale JPEG (`mode='L'`), `np.array(img)` will produce a shape other than `(H, W, 3)` — which then silently breaks `Image.fromarray((img * 255).astype(np.uint8))` in `rescale` (expects 3 channels) and everything downstream.

Suggested fix:
```python
 - img = Image.open(image_path)
 + img = Image.open(image_path).convert("RGB")
```
With this, the `if len(img.shape) == 2` grayscale guard below also becomes unreachable and can be removed.

forge/test/models/onnx/vision/ssd300_resnet50/model_utils/image_utils.py
```diff
@@ -17,17 +18,23 @@ def load_image(image_path):
...
+    # Convert to PIL Image for resizing
+    img_pil = Image.fromarray((img * 255).astype(np.uint8))
+
     if aspect > 1:
         # landscape orientation - wide image
         res = int(aspect * input_height)
-        imgScaled = skimage.transform.resize(img, (input_width, res))
-    if aspect < 1:
+        # PIL resize takes (width, height), skimage takes (height, width)
+        imgScaled = img_pil.resize((res, input_width), Image.BILINEAR)
```
chandrasekaranpradeep: `Image.BILINEAR` is deprecated since Pillow 9.1 and emits a `DeprecationWarning` on Pillow ≥ 10.0.0 

We can use `Image.Resampling.BILINEAR`.

Suggested fix (applies to all three resize calls):
 ```python
- imgScaled = img_pil.resize((res, input_width), Image.BILINEAR)
+ imgScaled = img_pil.resize((res, input_width), Image.Resampling.BILINEAR)
 ```

chandrasekaranpradeep (approved):

forge/test/models/onnx/vision/ssd300_resnet50/model_utils/image_utils.py
nsumrakTT: Good catch, thank you @chandrasekaranpradeep !

PR #3384, opened 2026-07-22
tvm_to_python: import CompilerConfig used by the compiler_cfg default path

uditjainstjis:
### Problem

`tvm_to_python.py` calls `CompilerConfig()` at two places without importing it:

- `generate_forge_module()` — `forge/forge/tvm_to_python.py:1820`
- `compile_tvm_to_python()` — `forge/forge/tvm_to_python.py:1918`

Both sit on the `compiler_cfg=None` default path that each function's own signature documents (`:1811`, `:1913`). The module imports `forge.verify.config._get_global_verify_config`, but never `forge.config.CompilerConfig`, so either function called with the documented default raises:

```
NameError: name 'CompilerConfig' is not defined
```

`pyflakes` reports this as two `F821 undefined name 'CompilerConfig'` on the current tip.

### Fix

Add the missing import, matching how `compile.py`, `forge_property_utils.py` and `tvm_unique_op_generation.py` already import the same symbol:

```python
from forge.config import CompilerConfig
```

One line, one file. `CompilerConfig` is a dataclass whose fields are all defaulted, so `CompilerConfig()` is valid, and there is no import cycle — `forge.config` does not import `forge.tvm_to_python`.

### Scope, stated honestly

This path is currently latent: every internal caller passes a `compiler_cfg` explicitly, so nothing in-tree hits it today. The value is that the documented default no longer raises, and `pyflakes` goes from two `F821`s to zero on this file.

### Verification

No device needed. `pyflakes forge/forge/tvm_to_python.py` — two `F821` before, zero after. `black --line-length=120 --check` clean.

### DIFF
--- a/forge/forge/tvm_to_python.py
+++ b/forge/forge/tvm_to_python.py
@@ -12,6 +12,7 @@ import numpy as np
 import pytest
 
 # import forge._C.pattern_matcher as pypattern_matcher
+from forge.config import CompilerConfig
 from forge.module import OnnxModule, ForgeModule, TFLiteModule
 from forge.verify.config import _get_global_verify_config
 import forge


chandrasekaranpradeep:
@uditjainstjis Thanks for your contribution! It looks good to me. I have triggered the CI pipeline once it passes we are good to merge

chandrasekaranpradeep:
@nvukobratTT CI pipeline passed, can you please take a look?

nvukobratTT (approved):

